Issue #245, opened 2015-11-24
infinite loop on Windows 10 in pa_win_wdmks.c because of RealTek HW


PortAudio-admin:
*Issue created by @philburk*

David Freese reported this bug to the mail list.

----------- BEGIN QUOTE
I am the author of fldigi, a digital modem program that uses the portaudio library.  fldigi runs on all versions of Windows beginning with XP, OS X (i386 and ppc), Linux, Free-BSD and Open-BSD.  The combined user list exceeds 10,000.

A segment of the Windows user community began experiencing application lock up problems commencing with the introduction of Windows 10.  The problem extended to other versions of Windows as Microsoft OS installed patches to their user base.  The common factor in all of the failures was discovered to be Realtek audio h/w installed on the system mother board.  In some (but not all) instances the problem could be averted by installing a new Realtek driver.

To assist in tracking the problem I modified the pa_devs.c program and distributed it for Linux, OS-X and Windows testers.  The Windows build uses mxe and is cross compiled on a Linux platform.  The portaudio library was built with the "--enable-debug-output" flag enabled with suitable callback in the modified pa_devs.c program.

The pa library would never exit a while(1){...} block in the function ULONG GetConnectedPin(...) found in the file pa_win_wdmks.c.  This only occurred with the Realtek h/w.  A copy of the modified pa_devs.c and a part of the debug output is attached

unmodified debug output:  locked.txt

after mod to pa_win_wdmks.c the debug output: pa_debug.txt

the result file from p_devs.c: result.txt

The modified pa library completely resolved the Windows user problem.  I'm not sure it was the best code solution, simply an expedient way to avoid the problem of never leaving the while(1) loop.

Thank you all for an excellent multiple OS library.  I know how difficult it can be to write for multiple platforms.

David Freese
--------------------- END QUOTE

From the debug log:

FindStartConnectionFrom: Checking 2 connections...GetConnectedPin: count=1, forward=0, muxPosition=0
GetNumberOfConnectionsTo: Checking 2 connections...GetConnectionFrom: Checking 2 connections... (pTo = 00262F40)
GetConnectionFrom: Returning 00262F50
GetConnectionFrom: Checking 2 connections... (pTo = 00262F50)
GetConnectionFrom: Returning 00262F40
GetConnectedPin: count=1, forward=0, muxPosition=0
GetNumberOfConnectionsTo: Checking 2 connections...GetConnectionFrom: Checking 2 connections... (pTo = 00262F40)
(continues forever...)

PR #300, opened 2020-09-24
Fix crashes on disconnected pins in wmdks topologies.

cme:
Odd hardware or broken driver installations encountered by some [Hydrogen](https://github.com/hydrogen-music/hydrogen/) users on Windows cause a crash in the application on initialising PortAudio. The WMDKs topology filters on these Windows machines contain pins which are not connected. 

The code asserts that this can never happen. It turns out, it does indeed happen, consistently and repeatably but only on certain machines, causing much frustration to a small number of users (and a smaller number of developers trying to figure out why their app crashes on machines they can't get access to :D ).

The fix removes the erroneous asserts and passes the error condition to the caller, and allows the caller to just consider the topology unusable.

https://music.columbia.edu/pipermail/portaudio/2014-August/016249.html


j-schultz:
I have encountered these assertions myself with the ESI Juli@ PCI sound card and I can confirm that PortAudio works as intended when removing the assertions. We have removed them in OpenMPT as well and this solved issues others users had as well.

sagamusix:
I have encountered these assertions myself with the ESI Juli@ PCI sound card and I can confirm that PortAudio works as intended when removing the assertions. We have removed them in OpenMPT as well and this solved issues others users had as well.

philburk:
I generally approve of this change. I think we should avoid asserts on conditions that are outside our control, like device configuration. Also it seems like a very low risk change.

philburk (approved):
Thanks!

philburk:
This may be related to #245
which mentions an assert(FALSE) and also an infinite loop involving finding Connections.

cme:
> This may be related to #245
> which mentions an assert(FALSE) and also an infinite loop involving finding Connections.

Ah, yes, I saw that before also. I suspect the infinite looping is probably because of a broken topology that contains a cycle, so similar issue but won't be fixed by this fix.

I think the takeaway message is that WMDKS topologies can't be trusted (I *think* they're stored in the registry, so are user-malleable and should be treated very defensively)

sagamusix:
There was [a similar issue with Kernel Streaming on Windows XP](https://sourceforge.net/p/audacity/mailman/message/31421748/) that was fixed a long while ago; maybe the infinite loop with WaveRT is a similar issue?

cme:
Aha, that does look very much like the topology contains a mux which feeds back to itself. It's probably not *supposed* to do that, but again I think PortAudio should deal with the possibility.

A simple std::set of visited nodes is enough to safely detect a cycle and avoid it.

cme:
@sagamusix it looks like the infinite looping issue has been fixed, as there's a limit set on the loop, added in: [d7085cb2](https://github.com/PortAudio/portaudio/commit/d7085cb2) . Less elegant than a map of visited nodes, but it works.

sagamusix:
I didn't have a deep look but is it really fixed for both WaveRT and KS or just KS? Both APIs are handled in the same file but newer systems will use WaveRT. Maybe this discussion should be continued though on the related issue.

RossBencina:
Hi all. I'm not overly familiar with PA/WDMKS, but I agree that calling `assert()` based on external input is to be avoided. This patch corrects that. However I am not (as yet) convinced that this patch resolves the issue -- because the indirect callers of `FindStartConnectionFrom`/`FindStartConnectionTo` also have asserts in them (see below). What do you think?

Analysis: The current patch removes asserts from `FindStartConnectionFrom` and `FindStartConnectionTo`. In both cases this causes a return of 0/NULL to `GetConnectedPin`, which in turn causes `GetConnectedPin` to return `KSFILTER_NODE`. Now `GetConnectedPin` has four call sites (listed below). Given that we now consider it normal behavior to return `NULL` from `FindStartConnection...`, thus returning `KSFILTER_NODE` from `GetConnectedPin`, then the call sites need to deal with it. To my mind, cases 1, and 4 below look fishy. 1884-1886 contain the text "No TOPO pin id ??? This is bad." and line 1974 is "assert(FALSE)". I don't understand the 1884 code -- is it correct? But the assert(FALSE) at 1974 should be replaced by some kind of error logging and cause the function to return paUnanticipatedHostError. What do you think?

## #1 Line 1603

https://github.com/PortAudio/portaudio/blob/8119355f76247f95883ae0940b579877e82de17d/src/hostapi/wdmks/pa_win_wdmks.c#L1603

leads to  line 1985 "No TOPO pin id ??? This is bad":
https://github.com/PortAudio/portaudio/blob/8119355f76247f95883ae0940b579877e82de17d/src/hostapi/wdmks/pa_win_wdmks.c#L1984


## #2 Line 1737

https://github.com/PortAudio/portaudio/blob/8119355f76247f95883ae0940b579877e82de17d/src/hostapi/wdmks/pa_win_wdmks.c#L1737

leads to line 1741 which results in return of paUnanticipatedHostError
https://github.com/PortAudio/portaudio/blob/8119355f76247f95883ae0940b579877e82de17d/src/hostapi/wdmks/pa_win_wdmks.c#L1741

## #3 Line 1811

https://github.com/PortAudio/portaudio/blob/8119355f76247f95883ae0940b579877e82de17d/src/hostapi/wdmks/pa_win_wdmks.c#L1811

lead's to line 1822 "we're done, break;"
https://github.com/PortAudio/portaudio/blob/8119355f76247f95883ae0940b579877e82de17d/src/hostapi/wdmks/pa_win_wdmks.c#L1822

## #4 Line 1919

https://github.com/PortAudio/portaudio/blob/8119355f76247f95883ae0940b579877e82de17d/src/hostapi/wdmks/pa_win_wdmks.c#L1919

leads to "assert(FALSE)" at 1974

https://github.com/PortAudio/portaudio/blob/8119355f76247f95883ae0940b579877e82de17d/src/hostapi/wdmks/pa_win_wdmks.c#L1974

RossBencina:
I've prepared an additional fix to handle `PinNew()` failing: https://github.com/PortAudio/portaudio/pull/309

We still need to fix `PinNew()` as outlined above.

RossBencina:
@cme @sagamusix Hey guys, Phil and I discussed this issue. The four code call sites that I identified remain a concern (lines 1603, 1737, 1811, 1919). We'd like to hear your opinion before deciding whether to merge this PR or not.

sagamusix:
All I can say is that all the devices that I expect to show up are present in PortAudio's device list, although some of them seem to be missing a "friendly name" probably due to the code paths taken, so the name as returned by PortAudio is e.g. "Input (Juli@ Multi-4ch)" for one input device while for another it's "Juli@ Ch12 (Juli@ Ch12)". I tried to dive into the WDM-KS code but I can't quite wrap my head around it without spending a whole lot of time on it. But as mentioned before - We have used this patch for years without problems, so I would assume it's fine.

cme:
I'll try and find some time to review this weekend.

cme:
Agreed, the path from PinNew should use the existing error handling rather than just asserting, so I've incorporated that change.

I've had a quick look through some of the other existing asserts in that code and there are a few others that look as if they're too pessimistic and would be better served by return an error than asserting, but I think addressing those is a separate topic.

PR #304, opened 2020-09-26
Fix CMake build after renaming readme

ibauersachs:


philburk (approved):
Thanks for the fix! I did not anticipate breaking the build by renaming the README file.

Issue #303, opened 2020-09-26
Glitching in WASAPI in shared mode


sveinse:
I'm experiencing significant glitching when using WASAPI from master in shared mode. If using exclusive mode or a full-duplex audio stream, the audio doesn't glitch. 

I'm using PA from Python using the interface library [pyaudio](https://github.com/sveinse/pyaudio). I'm not excluding performance issues from the python environment, but all the other hostAPIs works smoothly without glitching.

The glitching is typically manifested as bursts of 4096 samples played back successfully and then 220-230 samples of pause in between. The pause interval varies. PA is called with `framesPerBuffer` of 1024. It improves the occurrence of glitching by increasing this number, but it doesn't remove the glitching altogether, even with very large values such as 65536.

For reference my environment is:

* Windows 10, 2004 version
* Compiled with MS VS 2019 Community Edition
* Applied patch for building (for MSVC 2019 and to removed ASIO): https://gist.github.com/sveinse/7a3442d6f8444b95c4084a7172ec5fdb
* Compile command (from git bash):

      '/c/Program Files (x86)/Microsoft Visual Studio/2019/Community/MSBuild/Current/Bin/MSBuild.exe' \ 
          build\\msvc\\portaudio.sln \
          //property:GenerateFullPaths=true \
          //t:build \
          //consoleloggerparameters:NoSummary \
          //p:Configuration=Debug \
          //p:Platform=win32

sveinse:
Some observations:

* PortAudio from `master` glitches on WASAPI playback
* PortAudio from `pa_stable_v190600_20161030` does not glitch
* WASAPI used in exclusive mode does not glitch
* WASAPI used in full duplex mode does not glitch

I made a git bisect to determine which commit caused the gliching behavior in WASAPI. The breaking commit commited on Sep 6 2018 is:

c3fee03ff143feca50dcbc53b89670f251bbce24

sveinse:
I think I have found evidence that this is related to sleep timing setup in `ProcThreadPoll()` in `pa_win_wasapi.c` and that the timing amounts have doubled in master causing these glitches.

[...]

PR #305, opened 2020-09-27
Implement fix for glitching WASAPI playback

sveinse:
Fixing #303 by reducing the sleep delays on playback in polled thread mainloop. This reverts the sleep delays back to the same values as used in the stable versions of PA.

### DIFF
--- a/src/hostapi/wasapi/pa_win_wasapi.c
+++ b/src/hostapi/wasapi/pa_win_wasapi.c
@@ -5764,6 +5764,7 @@ PA_THREAD_FUNC ProcThreadPoll(void *param)
 	// Calculate timeout for next polling attempt.
 	sleep_ms_in  = GetFramesSleepTime(stream->in.framesPerHostCallback / WASAPI_PACKETS_PER_INPUT_BUFFER, stream->in.wavex.Format.nSamplesPerSec);
 	sleep_ms_out = GetFramesSleepTime(stream->out.framesPerBuffer, stream->out.wavex.Format.nSamplesPerSec);
+	sleep_ms_out /= 2; // wait only for half of the buffer
 
 	// WASAPI Input packets tend to expire very easily, let's limit sleep time to 2 milliseconds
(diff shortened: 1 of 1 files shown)


dmitrykos (approved):
@sveinse, thank you for a fix! Introducing `sleep_ms_out /= 2` in a separate line is a good approach and falls into the same manner as it was done in the blocking API.

kmatheussen:
Is there a CPU penalty only sleeping 2ms (which happens when recording)? If not, wouldn't make sense to sleep only 2ms for playback as well? Unless windows guarantees hard RT sleeping accuracy here, why take the chance sleeping more?

dmitrykos:
There is always a penalty (more instructions to process) if we call system API frequently (polling). According to the Microsoft docs/example (https://docs.microsoft.com/en-us/windows/win32/coreaudio/rendering-a-stream) we do not need to poll host buffer to often and polling 1/2 of the buffer duration is sufficient to have no underruns. 

Busy polling every 2 ms is too often for playback to my view, with recording there wasn't a choice to do other way as packets were expiring too early on some machines.

kmatheussen:
Of course there is a penalty. My question was rather if the penalty is anything to care about. Also, how can you be sure polling 1/2 of the buffer duration is sufficient when portaudio doesn't report glitches in wasapi? I mean, it's easy to overhear a few glitches now and then.

dmitrykos:
To my view increased CPU consumption can be a problem as it results in a CPU-hungry apps that are disliked by end users. Some systems would even kill such process if it exceeds some certain threshold of CPU consumption for the foreground or background process. Glitches can be solved by optimizes audio rending from the app's side, also you could do audio rendering not inside the PA's thread but in some other thread which would deliver ready buffers/data into the PA's callback thread.

sveinse:
@kmatheussen, @dmitrykos As this is is the the merge request for #303, let's keep the discussion in that issue, please. Otherwise the discussion will be all over the place. Thanks.

Issue #294, opened 2019-03-27
BuildDeviceList - invalid assertion - There may be no ( or hidden/disabled) audio devices on machine.


PortAudio-admin:
*Issue created by Anonymous on Assembla*

*Assertion interferes with debugging using "rr".*

In file:  src/hostapi/alsa/pa_linux_alsa.c
In function: BuildDeviceList
```
    /* Loop over list of cards, filling in info. If a device is deemed unavailable (can't get name),
     * it's ignored.
     *
     * Note that we do this in two stages. This is a workaround owing to the fact that the 'dmix'
     * plugin may cause the underlying hardware device to be busy for a short while even after it
     * (dmix) is closed. The 'default' plugin may also point to the dmix plugin, so the same goes
     * for this.
     */
    PA_DEBUG(( "%s: Filling device info for %d devices\n", __FUNCTION__, numDeviceNames ));
    for( i = 0, devIdx = 0; i < numDeviceNames; ++i )
    {
        PaAlsaDeviceInfo* devInfo = &deviceInfoArray[i];
        HwDevInfo* hwInfo = &hwDevInfos[i];
        if( !strcmp( hwInfo->name, "dmix" ) || !strcmp( hwInfo->name, "default" ) )
        {
            continue;
        }

        PA_ENSURE( FillInDevInfo( alsaApi, hwInfo, blocking, devInfo, &devIdx ) );
    }
    assert( devIdx < numDeviceNames );
```

Line 1453 contained above:
    assert( devIdx < numDeviceNames );

Which is not true if there are no devices to list.

I came across this trying to debugging a program using Mozilla's rr https://rr-project.org/. It records the execution of a program for deterministic debugging reruns, reverse debugging etc. It can not handle shared memory and I understand that it disables pulse-audio. I'm assuming it also disables alsa automatically. This results in the above assertion being triggered.

I compiled and ran with the assertion removed and I was able to record the program execution with rr.

PR #312, opened 2020-10-01
alsa: Option to ignore all ALSA plugin devices.

dmitrykos:
Option to ignore all ALSA plugin devices by using PA_ALSA_IGNORE_ALL_PLUGINS environment variable. 

Fixed logic of guarding assertion in BuildDeviceList() when only one PCM device is available (may occur if PA_ALSA_IGNORE_ALL_PLUGINS is defined).

Ignoring ALSA plugin devices is useful when using using PortAudio with ALSA interface in embedded systems like [Yocto Linux](http://www.yoctoproject.org). Some plugins are causing segfaults and practically may not be needed at all for the developer. By defining PA_ALSA_IGNORE_ALL_PLUGINS developer is able to switch off Alsa Plugin devices, only PCM devices will exist in this case (internal sound card, USB DAC, ...).

### DIFF
--- a/src/hostapi/alsa/pa_linux_alsa.c
+++ b/src/hostapi/alsa/pa_linux_alsa.c
@@ -1066,6 +1066,10 @@ static int IgnorePlugin( const char *pluginId )
     static const char *ignoredPlugins[] = {"hw", "plughw", "plug", "dsnoop", "tee",
         "file", "null", "shm", "cards", "rate_convert", NULL};
     int i = 0;
+
+    if( getenv( "PA_ALSA_IGNORE_ALL_PLUGINS" ) && atoi( getenv( "PA_ALSA_IGNORE_ALL_PLUGINS") ) )
+        return 1;
+
     while( ignoredPlugins[i] )
     {
         if( !strcmp( pluginId, ignoredPlugins[i] ) )
@@ -1327,6 +1331,8 @@ static PaError BuildDeviceList( PaAlsaHostApiRepresentation *alsaApi )
                     paInsufficientMemory );
             snprintf( deviceName, len, "%s: %s (%s)", cardName, infoName, buf );
 
+            PA_DEBUG(( "%s: Found device [%d]: %s\n", __FUNCTION__, numDeviceNames, deviceName ));
+
             ++numDeviceNames;
             if( !hwDevInfos || numDeviceNames > maxDeviceNames )
             {
@@ -1450,7 +1456,7 @@ static PaError BuildDeviceList( PaAlsaHostApiRepresentation *alsaApi )
 
         PA_ENSURE( FillInDevInfo( alsaApi, hwInfo, blocking, devInfo, &devIdx ) );
     }
-    assert( devIdx < numDeviceNames );
+    assert( devIdx <= numDeviceNames );
     /* Now inspect 'dmix' and 'default' plugins */
     for( i = 0; i < numDeviceNames; ++i )
     {


src/hostapi/alsa/pa_linux_alsa.c
```diff
@@ -1063,6 +1063,7 @@ static PaError PaAlsa_StrDup( PaAlsaHostApiRepresentation *alsaApi,
  */
 static int IgnorePlugin( const char *pluginId )
 {
+#ifndef PA_ALSA_IGNORE_ALL_PLUGINS
```
RossBencina, line 1066: Is there a reason why you did the #ifndef here rather than at the call site? :

https://github.com/PortAudio/portaudio/blob/e1f0c230aac0d9aa0d45b4f8600061de6f14c62f/src/hostapi/alsa/pa_linux_alsa.c#L1388

Maybe the debug logging should not be suppressed.

src/hostapi/alsa/pa_linux_alsa.c
```diff
@@ -1450,7 +1456,7 @@ static PaError BuildDeviceList( PaAlsaHostApiRepresentation *alsaApi )
 
         PA_ENSURE( FillInDevInfo( alsaApi, hwInfo, blocking, devInfo, &devIdx ) );
     }
-    assert( devIdx < numDeviceNames );
+    assert( devIdx <= numDeviceNames );
```
RossBencina, line 1459: What is the reason for this change? It is not mentioned in the commit message.

src/hostapi/alsa/pa_linux_alsa.c
dmitrykos: I did not want to pollute the code using an existing `IgnorePlugin` function, in fact modifying `IgnorePlugin` is perfectly fine for the logic of `ifdef` as it already ignoring some plugins and with `#ifndef PA_ALSA_IGNORE_ALL_PLUGINS` will ignore all plugins, that is the goal of the proposed change.

src/hostapi/alsa/pa_linux_alsa.c
dmitrykos: It is mentioned `Fixed logic of guarding assertion in BuildDeviceList() when only one PCM device is available (may occur if PA_ALSA_IGNORE_ALL_PLUGINS is defined).`. 

When `PA_ALSA_IGNORE_ALL_PLUGINS` is defined, e.g. user compiled PA to ignore all ALSA plugins, then `assert( devIdx < numDeviceNames )` will become true and process will stop.

dmitrykos: The algorithm of the failing assert is the following:

Assume device has only USB DAC attached as playback device.

1. `numDeviceNames` becomes 1
2. `FillInDevInfo` (line 1457) increments `devIdx` to 1
3. loop at line 1448 `for( i = 0, devIdx = 0; i < numDeviceNames; ++i )` exits as there is only one PCM device (`numDeviceNames == 1`)
4. assert condition becomes false `assert(1 < 1)`

dmitrykos:
The discussed assert (https://github.com/PortAudio/portaudio/blob/master/src/hostapi/alsa/pa_linux_alsa.c#L1454) is also affecting the issue #294 which can be closed by this patch.

Close #294

CMakeLists.txt
```diff
@@ -297,12 +297,16 @@ ELSE()
       OPTION(PA_USE_ALSA "Enable support for ALSA" OFF)
     ENDIF()
     IF(PA_USE_ALSA)
+      OPTION(PA_ALSA_IGNORE_ALL_PLUGINS "Ignore all ALSA plugin devices" OFF)
```
philburk, line 300: I notice that several ALSA options are controlled by environment variables. See multiple calls to getenv() in "pa_linux_alsa.c". That would allow a user to disable plugin devices for a precompiled app. If the plugin devices are specific to a machine then using getenv() might be better than disabling plugins for the app on all machines.

src/hostapi/alsa/pa_linux_alsa.c
philburk: If we used a getenv() then we could just skip the entire block that starts at 1358.

philburk: If we used a getenv() then we could just skip the entire block that starts at 1363.
Maybe something like:

    if (getenv("PA_ALSA_IGNORE_ALL_PLUGINS")
            && atoi( getenv( "PA_ALSA_IGNORE_ALL_PLUGINS" ) ) )
    {
                PA_DEBUG(( "%s: Ignoring all ALSA plugin devices\n", __FUNCTION__ ));
    }
    else if( ( res = alsa_snd_config_search( *alsa_snd_config, "pcm", &topNode ) ) >= 0 )

philburk: @dmitrykos - What do you think about using getenv()?
Using getenv() makes it machine specific so users can configure it. Plugins are also machine specific so it seems more appropriate to use getenv().

dmitrykos: @philburk , thank you for your comment. I agree, will update implementation accordingly.

dmitrykos:
Changed implementation to the usage of environment variable instead of compile-time define as per our discussion.

philburk (approved):
Thanks for using getenv().

RossBencina (approved):

PR #326, opened 2020-10-12
CMakeLists.txt: Fix missing deps when not using default libs with DSOUND

Nemirtingas:
Fix missing dependencies when building without default libs:
- advapi32
```
lld-link-9: error: undefined symbol: __imp__RegQueryValueExW@24
>>> referenced by /vcpkg/buildtrees/portaudio/src/ebd3ebfacc-58e92047a8/src/hostapi/wdmks/pa_win_wdmks.c:3841
>>>               CMakeFiles/portaudio.dir/src/hostapi/wdmks/pa_win_wdmks.c.obj:(_PaWinWdm_Initialize)

lld-link-9: error: undefined symbol: __imp__RegCloseKey@4
>>> referenced by /vcpkg/buildtrees/portaudio/src/ebd3ebfacc-58e92047a8/src/hostapi/wdmks/pa_win_wdmks.c:3841
>>>               CMakeFiles/portaudio.dir/src/hostapi/wdmks/pa_win_wdmks.c.obj:(_PaWinWdm_Initialize)
```
- user32
```
lld-link-9: error: undefined symbol: __imp__GetDesktopWindow@0
>>> referenced by /vcpkg/buildtrees/portaudio/src/ebd3ebfacc-58e92047a8/src/hostapi/dsound/pa_win_ds.c:2239
>>>               CMakeFiles/portaudio.dir/src/hostapi/dsound/pa_win_ds.c.obj:(_OpenStream)
>>> referenced by /vcpkg/buildtrees/portaudio/src/ebd3ebfacc-58e92047a8/src/hostapi/dsound/pa_win_ds.c:2264
>>>               CMakeFiles/portaudio.dir/src/hostapi/dsound/pa_win_ds.c.obj:(_OpenStream)
```


RossBencina:
@dechamps could you give your opinion on this please?

dechamps:
@Nemirtingas could you please explain what problem you're trying to solve here? I'm not sure I understand the scenario in which these extra declarations are required.

Nemirtingas:
Theses are not 'extra' dependencies, they are linked by default on Windows, but if you disable the default libs in CMake then they are missing when you link portaudio. You use theses functions in your code and such you need to link the libraries that exports them.

dechamps:
@Nemirtingas Sure. Can you explain exactly what you mean by "disable the default libs in CMake"? I can't find that option.

Nemirtingas:
> 
> 
> @Nemirtingas Sure. Can you explain exactly what you mean by "disable the default libs in CMake"? I can't find that option.

Yes, there default libs can be found here: 
https://github.com/Kitware/CMake/blob/ce7e615fe84f4bf9cbcb6dc099cfc3f651304f51/Modules/Platform/Windows-MSVC.cmake#L212-L216

dechamps:
@Nemirtingas I still don't understand how you can get yourself into a situation where default libs are not automatically added, and you need to explicitly specify them in CMakeLists.txt. You pointed to internal CMake code that is responsible for implicitly adding these libraries (thank you for that), but that doesn't explain why you feel the need to add them explicitly. I'm not sure it makes sense to add link parameters that are not actually needed in any reasonable scenario, as that would basically be dead code.

Basically what I'm saying is, I'm still unconvinced that there is anything broken that needs fixing here. It would help if you could describe a scenario (e.g. a specific compilation environment, a specific set of CMake options that a reasonable user might want to pass) in which the current CMakeLists.txt doesn't work.

Nemirtingas:
I'm currently cross-compiling lots of libraries on Linux for Windows and/or MacOS. And when cross-compiling for Windows, CMake is not aware of the default libraries to add, and I had also 1 project that was overriding the CMAKE_C_STANDARD_LIBRARIES_INIT variable, but I don't remember which one.

dechamps:
>  And when cross-compiling for Windows, CMake is not aware of the default libraries to add

I tried to reproduce your problem by cross-compiling PortAudio from Debian Sid to `i686-w64-mingw32` and it worked just fine out-of-the-box.

I'm still unconvinced that there is anything that needs fixing here. If default libs are not being passed, that seems like a bug in CMake or a misconfigured build environment, not a bug in the PortAudio CMake files.

> You could also just declare `/NODEFAULTLIB` in the linker options to disable all default libs.

Sure. But I can't think of a reasonable scenario where someone might want to pass that option unless they're deliberately trying to make things harder for themselves.

Nemirtingas:
Ok, if you don't feel like adding this, then no problem. I've forked the project and added it, that fixes my issue.

PR #328, opened 2020-10-13
Add the install interface to CMakeLists.txt

Nemirtingas:


RossBencina:
@dechamps could you give your opinion on this please?

jcelerier:
as an outsider this looks good and will make it easier to use portaudio from other cmake-based buildsystems, right now one had to add the include paths manually, this will make it unnecessary - just calling `target_link_libraries(my_app PRIVATE portaudio)` will also forward the correct include paths, etc

dechamps:
@RossBencina I think this should be merged; it makes sense, is consistent with CMake best practices, and can help consumer projects locate PortAudio when configuring. I have verified that it doesn't seem to break anything.

Nemirtingas:
Thanks, that will help vcpkg by removing a patch for your library.

Issue #218, opened 2012-11-10
PA/CoreAudio uses some "deprecated" APIs which fail on Mac OS 11.0


PortAudio-admin:
*Issue created by @RossBencina*

We are using a number of APIs that Apple have marked as deprecated. Some were deprecated in OS X 10.5, OS X 10.6, others in OS X 10.8.

We are using the old APIs for compatibility with older OS X versions. At this stage we wish to continue to support OS X 10.4 and later, so upgrading to APIs added in 10.5 and later is not an option.

The use of deprecated APIs presents two minor issues:

1. The compiler is emitting deprecated warnings, so we can't use -Werror as some would like (this can be addressed by removing -Werror, using -Wno-deprecated or using a #pragma, see: http://stackoverflow.com/questions/2622017/suppressing-deprecated-warnings-in-xcode )
1. The requirement to link with Carbon.framework concerns some people

At this stage there are no obvious technical advantages to moving to the newer APIs. Our plan is to deal with these either when the deprecated APIs are removed completely or when we stop supporting older OS X versions for other reasons. The API functions are listed below along with comments on the migration path.

# Core Audio Property Accessor APIs
We use the following old-style HAL Property Accessor functions:

* AudioDeviceGetPropertyInfo
* AudioDeviceGetProperty
* AudioDeviceAddPropertyListener
* AudioDeviceRemovePropertyListener

Since OSX 10.5 these Core Audio HAL APIs have newer direct replacements. 
See http://developer.apple.com/library/mac/#technotes/tn2223/_index.html

# Audio Converter Services AudioConverterFillBuffer

Since OS X 10.5 *AudioConverterFillBuffer* is deprecated in favor of AudioConverterFillComplexBuffer.
See https://developer.apple.com/library/mac/#documentation/MusicAudio/Reference/AudioConverterServicesReference/Reference/reference.html

# Component Manager APIs for instantiating Audio Units (requires Carbon.framework)
These Component Manager functions are used for instantiating Audio Units. They require linking with Carbon.framework. 

*  "_CloseComponent", referenced from:
 *      _CloseStream in libportaudio.a(pa_mac_core.o)
 *    _OpenAndSetupOneAudioUnit in libportaudio.a(pa_mac_core.o)
*  "_FindNextComponent", referenced from:
 *      _OpenAndSetupOneAudioUnit in libportaudio.a(pa_mac_core.o)
*  "_OpenAComponent", referenced from:
[...]

PR #338, opened 2020-10-22
Replace usage of deprecated component manager APIs on 10.6+

hoolrory:
These APIs are deprecated in 10.8+ and are disabled on 11.0(10.16)+ https://developer.apple.com/documentation/coreservices/carbon_core/component_manager?language=objc

Console output trying to use existing portaudio in Xcode 12.2 (beta)
> "This application, or a library it uses, is using the deprecated Component Manager for hosting Audio Components. This is not supported when rebuilding against the 10.16 or later SDK. Also, this makes the host incompatible with version 3 audio units. Please transition to the API's in AudioComponent.h."


RossBencina (approved):

philburk:
For #218

PR #341, opened 2020-10-26
wasapi: Fixed compilation with MinGW32 (original mingw32, not mingw-w64).

dmitrykos:
Fixed compilation with MinGW (mingw32): gcc version 6.3.0 (MinGW.org GCC-6.3.0-1), include /mingw-include folder explicitly (when compiling with mingw-w64 the /mingw-include folder must not be included).


dechamps:
As indicated by the automated Github Actions CI report, this patch breaks cross-compilation to mingw-w64 from Linux.

Since mingw-w64 compilation (at least from Linux) already worked before this PR, I'm not sure I understand what issue this PR is trying to solve. Is this to fix the build on some other variant of MinGW? Or to fix issues that only occur when building directly from Windows, as opposed to cross-compiling?

By the way, there is no need to "fill /mingw-include folder with these missing headers", at least not in the mingw-w64 case. It works out of the box.

dechamps:
Okay, I read this again and it looks like I was confused between mingw32 and mingw-w64 (which are not quite the same thing).

Not sure I see much point in supporting a platform (mingw32) that won't even compile anyway, but I guess it doesn't hurt.

In any case, the mingw-w64 build failures need to be resolved before this can be merged.

dmitrykos:
@dechamps of course :)

There were conflicts but now will be built correctly by mingw-w64 and the original mingw32 (if /mingw-include folder is included).

dmitrykos:
@dechamps 

> By the way, there is no need to "fill /mingw-include folder with these missing headers", at least not in the mingw-w64 case. It works out of the box.

Yes, it is true but original MinGW32 (not W64) is available and this patch makes it possible to compile WASAPI if /mingw-include is included by the user. I have completed necessary includes with missing headers, so user will not need to add any files additionally.

dechamps:
I have no objections to merging this, because all it seems to do is preserve the status quo (as I only just realized, the `mingw-include` folder already exists, this PR merely updates it).

That said, if it was up to me, I would delete the `mingw-include` directory and I would instead tell users that they have to use a build system that has an up-to-date set of Windows headers (such as `mingw-w64` which seems to work perfectly fine) - I don't see why it should be PortAudio's responsibility to provide system headers, and that seems brittle and difficult to maintain. But I know @RossBencina and @philburk disagreed strongly with me in the past on the question of which build systems to maintain support for, so I'll defer to them.

dmitrykos:
@dechamps I see your point. Original MinGW32 is not discontinued (at least not announced) and therefore I do not think it is a big trouble to keep compatibility with it at this moment. 

Later if MinGW32 becomes officially discontinued we can remove /mingw-include from the repository and cleanup the WASAPI's implementation from its traces but users who wish to use that version of GCC due to some reason will still be able to checkout this commit.

PR #344, opened 2020-11-11
Don't crash if alsa_snd_pcm_poll_descriptors fails

### LINKED ISSUE #268: pa_linux_alsa.c:3636: PaAlsaStreamComponent_BeginPolling: Assertion `ret == self->nfds' failed
*Issue created by Anonymous on Assembla*

I am not sure if it is a PortAudio or an fldigi bug, but as a developer, I would say that if an assertion occurs inside the library, it is a bug of the library, unless we are dealing with some memory corruption or race conditions.

So, I can only reproduce the bug with [fldigi](https://sourceforge.net/p/fldigi/fldigi/ci/master/tree/). Reproducibility is 100%. Examples supplied with PortAudio and other PortAudio clients seem to work just fine.
Moreover, this happens only with Alsa 1.1.4.1 (and 1.1.4), but does not happen with the older 1.1.3.

The stacktrace is:

```
#0  0x00007ffff5a9802f in raise () from /lib64/libc.so.6
#1  0x00007ffff5a9942a in abort () from /lib64/libc.so.6
#2  0x00007ffff5a90e87 in ?? () from /lib64/libc.so.6
#3  0x00007ffff5a90f32 in __assert_fail () from /lib64/libc.so.6
#4  0x00007ffff7bb2273 in PaAlsaStreamComponent_BeginPolling (pfds=<optimized out>, self=<optimized out>)
    at /var/tmp/portage/media-libs/portaudio-19.06.00/work/portaudio/src/hostapi/alsa/pa_linux_alsa.c:3636
#5  0x00007ffff7bbfe15 in PaAlsaStreamComponent_BeginPolling (pfds=<optimized out>, self=<optimized out>)
    at /var/tmp/portage/media-libs/portaudio-19.06.00/work/portaudio/src/hostapi/alsa/pa_linux_alsa.c:3926
#6  PaAlsaStream_WaitForFrames (self=self@entry=0x7fffe4001a00, framesAvail=framesAvail@entry=0x7fffe99afe40,
    xrunOccurred=xrunOccurred@entry=0x7fffe99afe3c)
    at /var/tmp/portage/media-libs/portaudio-19.06.00/work/portaudio/src/hostapi/alsa/pa_linux_alsa.c:3797
#7  0x00007ffff7bc05ce in CallbackThreadFunc (userData=0x7fffe4001a00)
    at /var/tmp/portage/media-libs/portaudio-19.06.00/work/portaudio/src/hostapi/alsa/pa_linux_alsa.c:4269
#8  0x00007ffff7998374 in start_thread () from /lib64/libpthread.so.0
#9  0x00007ffff5b4db2f in clone () from /lib64/libc.so.6
```

daschuer:
This fixes #268 

This patch was original proposed by Chameleon here: 
https://portaudio.music.columbia.narkive.com/3V9hsUak/pa-linux-alsa-c-3636-assertion-failed-with-hack-fix

### DIFF
--- a/src/hostapi/alsa/pa_linux_alsa.c
+++ b/src/hostapi/alsa/pa_linux_alsa.c
@@ -3639,16 +3639,15 @@ error:
  */
 static PaError PaAlsaStreamComponent_BeginPolling( PaAlsaStreamComponent* self, struct pollfd* pfds )
 {
-    PaError result = paNoError;
-    int ret = alsa_snd_pcm_poll_descriptors( self->pcm, pfds, self->nfds );
-    /* If alsa returns anything else, like -EPIPE return*/
-    if( ret != self->nfds )
+    int nfds = alsa_snd_pcm_poll_descriptors( self->pcm, pfds, self->nfds );
+    /* If alsa returns anything else, like -EPIPE return */
+    if( nfds != self->nfds )
     {
-      return paUnanticipatedHostError;
+        return paUnanticipatedHostError;
     }
     self->ready = 0;
 
-    return result;
+    return paNoError;
 }
 
 /** Examine results from poll().
--- a/src/hostapi/alsa/pa_linux_alsa.c
+++ b/src/hostapi/alsa/pa_linux_alsa.c
@@ -3642,8 +3642,11 @@ static PaError PaAlsaStreamComponent_BeginPolling( PaAlsaStreamComponent* self, 
     PaError result = paNoError;
     int ret = alsa_snd_pcm_poll_descriptors( self->pcm, pfds, self->nfds );
     (void)ret;  /* Prevent unused variable warning if asserts are turned off */
-    assert( ret == self->nfds );
-
+    /* If alsa returns anything else, like -EPIPE return*/
+    if( ret != self->nfds )
+    {
+      return paUnanticipatedHostError;
+    }
     self->ready = 0;
 
     return result;
@@ -3810,7 +3813,12 @@ static PaError PaAlsaStream_WaitForFrames( PaAlsaStream *self, unsigned long *fr
         {
             /* self->pfds is in effect an array of fds; if necessary, index past the capture fds */
             playbackPfds = self->pfds + (pollCapture ? self->capture.nfds : 0);
-            PA_ENSURE( PaAlsaStreamComponent_BeginPolling( &self->playback, playbackPfds ) );
+            int res = PaAlsaStreamComponent_BeginPolling( &self->playback, playbackPfds );
+            if( res != paNoError)
+            {
+              xrun = 1;
+              goto end;
+            }
             totalFds += self->playback.nfds;
         }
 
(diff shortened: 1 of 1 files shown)


src/hostapi/alsa/pa_linux_alsa.c
```diff
@@ -3642,8 +3642,11 @@ static PaError PaAlsaStreamComponent_BeginPolling( PaAlsaStreamComponent* self,
     PaError result = paNoError;
     int ret = alsa_snd_pcm_poll_descriptors( self->pcm, pfds, self->nfds );
     (void)ret;  /* Prevent unused variable warning if asserts are turned off */
```
philburk, line 3644: Can remove (void) cast because we use ret.

philburk (changes requested):
I will review the rest in more detail.

philburk (approved):
This has been discussed and approved by several people on the mail list.

RossBencina:
I don't feel qualified to comment on this.

src/hostapi/alsa/pa_linux_alsa.c
```diff
@@ -3810,7 +3812,12 @@ static PaError PaAlsaStream_WaitForFrames( PaAlsaStream *self, unsigned long *fr
         {
             /* self->pfds is in effect an array of fds; if necessary, index past the capture fds */
             playbackPfds = self->pfds + (pollCapture ? self->capture.nfds : 0);
-            PA_ENSURE( PaAlsaStreamComponent_BeginPolling( &self->playback, playbackPfds ) );
+            int res = PaAlsaStreamComponent_BeginPolling( &self->playback, playbackPfds );
```
philburk, line 3815: If this change is needed here then would it not also be needed on line 3808?

philburk (changes requested):
I faked an xrun in pa_linux_alsa.c at line 3645 with:

    static int count = 0;
    if( ret != self->nfds || ((count++ % 173) == 150)) // FIXME fake an error

Then I tried running the test paex_read_write_wire and it failed.
But the test aborts when it gets an xrun.
So I modified the test to just print a message when ReadStream or WriteStream return an error.
Now for every xrun I trigger in pa_alsa I get a flood of error returns.

daschuer:
Done

src/hostapi/alsa/pa_linux_alsa.c
philburk: Thanks for the change.

philburk (approved):

philburk:
Fixes #268

antiochtech:
im still getting the error so i suspect my project is still using the old version. its a python project using the raspbian package python3-pyaudio. ive followed the installation recommendations for my device (https://github.com/PortAudio/portaudio/wiki/Platforms_RaspberryPi and http://portaudio.com/docs/v19-doxydocs/compile_linux.html), but i suspect theres another step?

PR #348, opened 2020-11-19
Update pa_jack.h

PaulBoersma:
missing prototype for PaJack_Initialize()

### DIFF
--- a/include/pa_jack.h
+++ b/include/pa_jack.h
@@ -50,6 +50,8 @@
 extern "C" {
 #endif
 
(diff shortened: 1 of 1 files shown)


RossBencina:
Closing as invalid. This pull request has the following problems:

1. It adds a prototype for an implementation-internal function to a public header file.
2. There is no bug. The function in question is declared before use in pa_unix_hostapis.c, which is all that is required by the C99 standard.
3. We don't offer MISRA compliance.

If you are getting compiler warnings, please open an issue ticket with an exact copy of the compiler warning and details of your compiler version and build configuration, then we can discuss further.

PR #349, opened 2020-11-19
Update pa_linux_alsa.h

PaulBoersma:
prototype for PaAlsa_Initialize()
I have changed this by hand with four earlier versions of PortAudio through the years

### DIFF
--- a/include/pa_linux_alsa.h
+++ b/include/pa_linux_alsa.h
@@ -50,6 +50,8 @@
 extern "C" {
 #endif
 
+PaError PaAlsa_Initialize( PaUtilHostApiRepresentation **hostApi, PaHostApiIndex hostApiIndex );
+
 typedef struct PaAlsaStreamInfo
 {
     unsigned long size;


philburk (commented):
I want to understand the need for this change.

This function prototype is currently defined at:
https://github.com/PortAudio/portaudio/blob/09087cf5a63d6fdb6aca139331f017da970f8177/src/os/unix/pa_unix_hostapis.c#L46

It will be called internally when the app calls Pa_Initialize().  An app should not need to call this directly.

What problem are you seeing that needs this fix?

PaulBoersma:
PaAlsa_Initialize() is defined in pa_linux_alsa.c. When compiling pa_linux_alsa.c, my compilers complain that this global function is defined without having been declared ("definition without prototype"). As pa_linux_alsa.c includes pa_linux_alsa.h, including the declaration there solves the problem.

philburk:
I found a site that might b describing the rule thet PortAudio is breaking.
https://lost-contact.mit.edu/afs/rhic.bnl.gov/i386_sl305/app/codewizard-4.3/manuals/items_mi.htm

MISRA-71 says "Functions shall always have visible prototype at both the function definition and call."

So it seems we may be violating MISRA-71 but this is not the right solution because it exposes an internal function prototype.

RossBencina:
Closing as invalid. This pull request has the following problems:

1. It adds a prototype for an implementation-internal function to a public header file.
2. There is no bug. The function in question is declared before use in pa_unix_hostapis.c, which is all that is required by the C99 standard.
3. We don't offer MISRA compliance.

If you are getting compiler warnings, please open an issue ticket with an exact copy of the compiler warning and details of your compiler version and build configuration, then we can discuss further.

PR #346, opened 2020-11-19
Update pa_mac_core_utilities.c

PaulBoersma:
Make definitions equal to delarations.

PR #345, opened 2020-11-19
Update pa_mac_core_utilities.h

PaulBoersma:
Insert "void" twice to turn these two declarations into prototypes.

PR #355, opened 2020-11-26
pa_mac_core_utilities: add void to empty function argument list

RossBencina:
Insert "void" in the argument list to turn initializeXRunListenerList and destroyXRunListenerList declarations into valid prototypes. Also update the function definitions to match. Thanks to Paul Boersma.


RossBencina:
Replaces #345 and #346

PR #356, opened 2020-11-26
CoreAudio add support for new HAL API for applications that target Mac OS X >= 10.5

jmelas:
This pull request addresses item #218 by using the new CoreAudio APIs while keeping compatibility with
applications that need to use the old APIs!
This is done with the following check
#if (MAC_OS_X_VERSION_MIN_REQUIRED >= MAC_OS_X_VERSION_10_5)
#define PA_NEW_HAL 1
#else
#define PA_NEW_HAL 0
#endif
which checks if the application targets Mac OSX >= 10.5, so applications targeting Mac OS X 10.4 will use the old API.

I tried to make the changes as minimal as possible by introducing wrapper functions and #ifdefs
for two reasons:
a) easier review
b) compatibility with old API

### DIFF
--- a/src/hostapi/coreaudio/pa_mac_core.c
+++ b/src/hostapi/coreaudio/pa_mac_core.c
@@ -2669,7 +2669,7 @@ static PaError CloseStream( PaStream* s )
                                                 0,
                                                 false,
                                                 kAudioDeviceProcessorOverload,
-                                                xrunCallback, NULL ); //todo: do we need to pass actual node?
+                                                xrunCallback, NULL ); //no need to pass actual node
        }
        if( stream->inputUnit && stream->outputUnit != stream->inputUnit ) {
           int count = removeFromXRunListenerList( stream );
@@ -2678,7 +2678,7 @@ static PaError CloseStream( PaStream* s )
                                                 0,
                                                 true,
                                                 kAudioDeviceProcessorOverload,
-                                                xrunCallback, NULL ); //todo: do we need to pass actual node?
+                                                xrunCallback, NULL ); //no need to pass actual node
        }
        if( stream->outputUnit && stream->outputUnit != stream->inputUnit ) {
           AudioUnitUninitialize( stream->outputUnit );
--- a/src/hostapi/coreaudio/pa_mac_core.c
+++ b/src/hostapi/coreaudio/pa_mac_core.c
@@ -2386,24 +2386,24 @@ static OSStatus AudioIOProc( void *inRefCon,
          if( stream->inputSRConverter )
          {
                OSStatus err;
-               UInt32 size;
                float data[ inChan * frames ];
-               size = sizeof( data );
                AudioBufferList bufferList;
                bufferList.mNumberBuffers = 1;
                bufferList.mBuffers[0].mNumberChannels = inChan;
-               bufferList.mBuffers[0].mDataByteSize = size;
+               bufferList.mBuffers[0].mDataByteSize = sizeof( data );
                bufferList.mBuffers[0].mData = data;
+               UInt32 packets = frames;
                err = AudioConverterFillComplexBuffer(
                    stream->inputSRConverter,
                    ringBufferIOProc,
                    &stream->inputRingBuffer,
-                   &size,
+                   &packets,
                    &bufferList,
                    NULL);
                if( err == RING_BUFFER_EMPTY )
                { /* the ring buffer callback underflowed */
                   err = 0;
+                  UInt32 size = packets * bytesPerFrame;
                   bzero( ((char *)data) + size, sizeof(data)-size );
                   /* The ring buffer can underflow normally when the stream is stopping.
                    * So only report an error if the stream is active. */
@@ -2576,20 +2576,17 @@ static OSStatus AudioIOProc( void *inRefCon,
          OSStatus err;
          do
          { /* Run the buffer processor until we are out of data. */
-            UInt32 size;
-            long f;
-
-            size = sizeof( data );
             AudioBufferList bufferList;
             bufferList.mNumberBuffers = 1;
             bufferList.mBuffers[0].mNumberChannels = chan;
-            bufferList.mBuffers[0].mDataByteSize = size;
+            bufferList.mBuffers[0].mDataByteSize = sizeof( data );
             bufferList.mBuffers[0].mData = data;
+            UInt32 packets = inNumberFrames;
             err = AudioConverterFillComplexBuffer(
                 stream->inputSRConverter,
                 ringBufferIOProc,
                 &stream->inputRingBuffer,
-                &size,
+                &packets,
                 &bufferList,
                 NULL);
             if( err != RING_BUFFER_EMPTY )
@@ -2599,9 +2596,8 @@ static OSStatus AudioIOProc( void *inRefCon,
                 goto stop_stream;
             }
 
-            f = size / ( chan * sizeof(float) );
-            PaUtil_SetInputFrameCount( &(stream->bufferProcessor), f );
-            if( f )
+            PaUtil_SetInputFrameCount( &(stream->bufferProcessor), packets );
+            if( packets > 0 )
             {
                PaUtil_BeginBufferProcessing( &(stream->bufferProcessor),
                                              &timeInfo,
--- a/src/hostapi/coreaudio/pa_mac_core.c
+++ b/src/hostapi/coreaudio/pa_mac_core.c
@@ -138,7 +138,7 @@ const char *PaMacCore_GetChannelName( int device, int channelIndex, bool input )
 	
 	/* First try with CFString */
 	UInt32 size = sizeof(nameRef);
-	error = pa_AudioDeviceGetProperty( hostApiDevice,
+	error = PaMacCore_AudioDeviceGetProperty( hostApiDevice,
 								   channelIndex + 1,
 								   input,
 								   kAudioDevicePropertyChannelNameCFString,
@@ -148,7 +148,7 @@ const char *PaMacCore_GetChannelName( int device, int channelIndex, bool input )
 	{
 		/* try the C String */
 		size = 0;
-		error = pa_AudioDeviceGetPropertySize( hostApiDevice,
+		error = PaMacCore_AudioDeviceGetPropertySize( hostApiDevice,
 										   channelIndex + 1,
 										   input,
 										   kAudioDevicePropertyChannelName,
@@ -158,7 +158,7 @@ const char *PaMacCore_GetChannelName( int device, int channelIndex, bool input )
 			if( !ensureChannelNameSize( size ) )
 				return NULL;
 
-			error = pa_AudioDeviceGetProperty( hostApiDevice,
+			error = PaMacCore_AudioDeviceGetProperty( hostApiDevice,
 										   channelIndex + 1,
 										   input,
 										   kAudioDevicePropertyChannelName,
@@ -223,7 +223,7 @@ PaError PaMacCore_GetBufferSizeRange( PaDeviceIndex device,
             if( macCoreHostApi->inheritedHostApiRep.deviceInfos[hostApiDeviceIndex]->maxOutputChannels == 0 )
                 isInput = 1;
 
-            result = WARNING(pa_AudioDeviceGetProperty( macCoreDeviceId, 0, isInput, kAudioDevicePropertyBufferFrameSizeRange, &propSize, &audioRange ) );
+            result = WARNING(PaMacCore_AudioDeviceGetProperty( macCoreDeviceId, 0, isInput, kAudioDevicePropertyBufferFrameSizeRange, &propSize, &audioRange ) );
 
             *minBufferSizeFrames = audioRange.mMinimum;
             *maxBufferSizeFrames = audioRange.mMaximum;
@@ -352,7 +352,7 @@ static PaError gatherDeviceInfo(PaMacAUHAL *auhalHostApi)
     auhalHostApi->devIds = NULL;
 
     /* -- figure out how many devices there are -- */
-    pa_AudioHardwareGetPropertySize( kAudioHardwarePropertyDevices,
+    PaMacCore_AudioHardwareGetPropertySize( kAudioHardwarePropertyDevices,
                                   &propsize);
     auhalHostApi->devCount = propsize / sizeof( AudioDeviceID );
 
@@ -364,7 +364,7 @@ static PaError gatherDeviceInfo(PaMacAUHAL *auhalHostApi)
                              propsize );
     if( !auhalHostApi->devIds )
         return paInsufficientMemory;
-    pa_AudioHardwareGetProperty( kAudioHardwarePropertyDevices,
+    PaMacCore_AudioHardwareGetProperty( kAudioHardwarePropertyDevices,
                                   &propsize,
                                   auhalHostApi->devIds );
 #ifdef MAC_CORE_VERBOSE_DEBUG
@@ -383,7 +383,7 @@ static PaError gatherDeviceInfo(PaMacAUHAL *auhalHostApi)
     /* I am not sure how these calls to AudioHardwareGetProperty()
        could fail, but in case they do, we use the first available
        device as the default. */
-    if( 0 != pa_AudioHardwareGetProperty(kAudioHardwarePropertyDefaultInputDevice,
+    if( 0 != PaMacCore_AudioHardwareGetProperty(kAudioHardwarePropertyDefaultInputDevice,
                      &size,
                      &auhalHostApi->defaultIn) ) {
        int i;
@@ -400,7 +400,7 @@ static PaError gatherDeviceInfo(PaMacAUHAL *auhalHostApi)
              }
        }
     }
-    if( 0 != pa_AudioHardwareGetProperty(kAudioHardwarePropertyDefaultOutputDevice,
+    if( 0 != PaMacCore_AudioHardwareGetProperty(kAudioHardwarePropertyDefaultOutputDevice,
                      &size,
                      &auhalHostApi->defaultOut) ) {
        int i;
@@ -435,7 +435,7 @@ static PaError ClipToDeviceBufferSize( AudioDeviceID macCoreDeviceId,
 	UInt32 resultSize = desiredSize;
 	AudioValueRange audioRange;
 	UInt32 propSize = sizeof( audioRange );
-	PaError err = WARNING(pa_AudioDeviceGetProperty( macCoreDeviceId, 0, isInput, kAudioDevicePropertyBufferFrameSizeRange, &propSize, &audioRange ) );
+	PaError err = WARNING(PaMacCore_AudioDeviceGetProperty( macCoreDeviceId, 0, isInput, kAudioDevicePropertyBufferFrameSizeRange, &propSize, &audioRange ) );
 	resultSize = MAX( resultSize, audioRange.mMinimum );
 	resultSize = MIN( resultSize, audioRange.mMaximum );
 	*allowedSize = resultSize;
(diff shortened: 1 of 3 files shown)


philburk (changes requested):
Thank you so much for this Mac OS update. This will make a lot of people happy.

I see that you also cleaned up a lot of trailing white space.   That's great. But unfortunately they obscure the relevant changes. Would it be possible to resubmit this pull request without the white space commit? We can do those as a later pull request. Generally it is best to separate style or format changes from real code changes.

Then we can do the detailed code review.

src/hostapi/coreaudio/pa_mac_core_utilities.h
```diff
@@ -111,9 +110,66 @@
...
 
+#if PA_NEW_HAL
+typedef AudioObjectPropertyListenerProc pa_AudioDevicePropertyListenerProc;
+#else
+typedef AudioDevicePropertyListenerProc pa_AudioDevicePropertyListenerProc;
+#endif
 
-
+OSStatus pa_AudioHardwareGetProperty(
+    AudioHardwarePropertyID inPropertyID,
```
philburk, line 126: Continuation lines shouldbe indented by 8 or line up with the "(".

philburk: Never mind. I see the whole file does this. We can address that later. Better to be consistent for now.

jmelas:
You are welcome Phil!
OK, I've removed the commit with the spaces cleanup!

src/hostapi/coreaudio/pa_mac_core_utilities.h
```diff
@@ -111,9 +111,66 @@
...
+    UInt32*                 ioPropertyDataSize,
+    void*                   outPropertyData);
+OSStatus pa_AudioHardwareGetPropertySize(
```
philburk, line 130: Please put one blank line between calls.

philburk (changes requested):

philburk:
I checked out this fork.
./configure && make clean && make

and got these errors:

libtool: compile:  gcc -c -std=c99 -O2 -Wall -pedantic -pipe -fPIC -DNDEBUG -DPA_LITTLE_ENDIAN -I./include -I./src/common -I./src/os/unix -Wno-deprecated -Werror -arch x86_64 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.15.sdk -mmacosx-version-min=10.4 -DPACKAGE_NAME=\"\" -DPACKAGE_TARNAME=\"\" -DPACKAGE_VERSION=\"\" -DPACKAGE_STRING=\"\" -DPACKAGE_BUGREPORT=\"\" -DPACKAGE_URL=\"\" -DSTDC_HEADERS=1 -DHAVE_SYS_TYPES_H=1 -DHAVE_SYS_STAT_H=1 -DHAVE_STDLIB_H=1 -DHAVE_STRING_H=1 -DHAVE_MEMORY_H=1 -DHAVE_STRINGS_H=1 -DHAVE_INTTYPES_H=1 -DHAVE_STDINT_H=1 -DHAVE_UNISTD_H=1 -DHAVE_DLFCN_H=1 -DLT_OBJDIR=\".libs/\" -DSIZEOF_SHORT=2 -DSIZEOF_INT=4 -DSIZEOF_LONG=8 -DHAVE_CLOCK_GETTIME=1 -DHAVE_NANOSLEEP=1 -DPA_USE_COREAUDIO=1 src/hostapi/coreaudio/pa_mac_core_utilities.c  -fno-common -DPIC -o src/hostapi/coreaudio/.libs/pa_mac_core_utilities.o
src/hostapi/coreaudio/pa_mac_core_utilities.c:75:5: error: use of undeclared identifier 'macErr'
    macErr = AudioHardwareGetProperty(inPropertyID, ioPropertyDataSize, outPropertyData);
    ^
src/hostapi/coreaudio/pa_mac_core_utilities.c:87:5: error: use of undeclared identifier 'macErr'
    macErr = AudioHardwareGetPropertyInfo(inPropertyID, outSize, NULL);
    ^
src/hostapi/coreaudio/pa_mac_core_utilities.c:104:5: error: use of undeclared identifier 'macErr'
    macErr = AudioDeviceGetProperty(inDevice, inChannel, isInput, inPropertyID, ioPropertyDataSize, outPropertyData);

philburk:
Thanks John. With your changes I can now compile using "-mmacosx-version-min=10.4".
I changed that to  "-mmacosx-version-min=10.5" and verified that is is calling your new code.
That also compiles OK.

I ran pa_devs, paex_sine, paex_record, paqa_devs, paqa_errs, paqa_latency and they passed OK.
Note that paqa_devs has some problems with numChannels > 2, but not because of this PR.

philburk (approved):

RossBencina:
Thanks for this. I haven't finished reviewing it but I just want to write down some notes here while I remember:

- Change PA_NEW_HAL to mac-specific naming convention, same for pa_ prefix on new functions
- Don't cut and paste to duplicate almost identical code between new and old listener callbacks -- should share the same code.

I'll explain more later, and perhaps Phil and I can advise for how to improve things.

jmelas:
About #ifdefs and duplication of code:
Do you think it makes sense to drop support for OS X 10.4? If yes then it's trivial to remove old code as it is now.
If not, indeed I will have to introduce new functions to hide the duplications!

Be-ing:
I am writing this comment on a 9 year old Macbook Air stuck on macOS 10.13 because Apple won't let me update it to a newer OS. macOS 10.04 was released in 2004. So I say, yes, drop it.

Be-ing:
@philburk @RossBencina would you be okay with removing the old API now?

philburk:
I tried to find statistics for the current share of machines running 10.4 and it is off the bottom of the chart for everything I could find.  I imagine that people running 10.4 would not expect newly built apps to still run on their machine.
I think most people with old machine have ways to find old versions of apps that will run.

For a new apps built today, even if PortAudio supported 10.4, the app would probably use other APIs that do not run on on 10.4.

10.4 came out in 2005 and machines built in 2005 probably upgraded to later versions.

So I think it is reasonable to drop support for 10.4 and to just note that if people want to support 10.4 then they can use an old stable release or a snapshot.

RossBencina:
I agree to dropping support for 10.4. If we can support 10.6 and later that is good enough. That said, the newer API is kinda sucky compared to the old one, so perhaps keeping the wrapper functions in this patch still makes sense?

RossBencina:
Further to my earlier comments: This may no longer be relevant, but our naming conventions are documented here:

https://github.com/PortAudio/portaudio/wiki/ImplementationStyleGuidelines

With respect to the added functions prefixed `pa_`: static functions should have no special prefix/suffix and module-specific globally visible functions should, in this case, be prefixed with `PaMacCore_`.

The preprocessor symbol naming convention does not appear to be clearly documented but in this case I'd say `PA_NEW_HAL' should instead be `PA_MAC_CORE_NEW_HAL` (since this is a symbol specific to pa mac core, not to PA as a whole.

jmelas:
Thank you all for commenting. I've just dropped support for Tiger and also renamed all helper functions to PaMacCore_.

philburk:
The code looks great. Thanks for making all these changes. But bad news...

I followed the instructions above to try the new code:

    git checkout -b jmelas-bigsur master
    git pull https://github.com/jmelas/portaudio.git bigsur
    ./configure && make clean && make

bin/paex_sine works OK
bin/paex_write_sine works OK
bin/paex_read_write_wire fails with this message and then hangs:

    Wire on. Will run 10 seconds.
    ||PaMacCore (AUHAL)|| Error on line 2409: err=''insz'', msg=Unknown Error

bin/patest_wire has a similar failure

If I switch back to the master branch then they work fine.
So there seems to be a problem with INPUT.

philburk (changes requested):
See comment describing failure and hang in paex_read_write_wire.

philburk:
I tried again with this morning's commit.
paex_read_write_wire now works OK.
patest_wire works the same on master and with your changes.

Also passing:
paqa_errs
patest_underflow
patest_callbackstop
patest_pink
pa_devs
patest_many

This looks great John! Thanks so much for your hard work on this.

philburk (approved):
Nice work!

jmelas:
Thank you @philburk  but my latest commit was simply a rebase from master plus some spacing fixes. I did not change anything.
Are you running the tests with different sound cards?
In my machine the tests run ok but I've managed to reproduce the error by changing the sample rate in paex_read_write_wire 
to 48000 to trigger sample rate conversion. Then I could see the problem you descibed: AudioConverterFillComplexBuffer returns 'insz' (invalid input size). Still investigating!

philburk:
@jmelas - There is clearly a hole in my test strategy. I started to run paqa_devs but it made a horrible sound full volume on my built-in speaker when I had headphones plugged in. So I stopped it. That would have probably have caught the error.

I was able to reproduce the error by opening the Audio MIDI setup,
changing the MacBook Pro Microphone sample rate to 48000,
then running the unmodified paex_read_write_wire.
That triggered the sample rate conversion and caused the error.

philburk (changes requested):
Removing Approval because INPUT is still failing when the sample rate converter is involved.

philburk:
I have been working on paqa_devs.c and I am getting asserts on line 2159 of pa_mac_core.c of this branch.

    PaUtil_GetRingBufferReadRegions( rb, *ioDataSize,
                                    outData, (ring_buffer_size_t *)ioDataSize, 
                                    &dummyData, &dummySize );
    assert( *ioDataSize );

philburk:
The new version of paqa_devs.c in #381 catches this error.

jmelas:
Good news Phil! All tests you described earlier pass now in my machine with sample rate conversion!

src/hostapi/coreaudio/pa_mac_core.c
```diff
@@ -2136,29 +2136,35 @@ static OSStatus ringBufferIOProc(
...
    PaUtilRingBuffer *rb = (PaUtilRingBuffer *) inUserData;
 
-   VVDBUG(("ringBufferIOProc()\n"));
-
    if( PaUtil_GetRingBufferReadAvailable( rb ) == 0 ) {
-      *outData = NULL;
-      *ioDataSize = 0;
+      ioData->mBuffers[0].mData = NULL;
+      ioData->mBuffers[0].mDataByteSize = 0;
+      *ioNumberDataPackets = 0;
```
philburk, line 2146: Returning a value through *ioNumberDataPackets is new behavior. Was that part of the bug?

philburk (commented):
This is passing my tests now.

Can you please describe the nature of the bug?
I see that the size calculation is completely different.

Was there a hidden bug before the conversion to the new API?

src/hostapi/coreaudio/pa_mac_core.c
jmelas: absolutely yes!

src/hostapi/coreaudio/pa_mac_core.c
```diff
@@ -2136,29 +2136,35 @@ static OSStatus ringBufferIOProc(
...
-   assert( *ioDataSize );
-   PaUtil_AdvanceRingBufferReadIndex( rb, *ioDataSize );
-   (*ioDataSize) *= rb->elementSizeBytes ;
+   assert(rbElementsRead > 0);
+   VVDBUG(("RingBuffer read elements %u of %u\n", rbElementsRead, rbElements));
+   PaUtil_AdvanceRingBufferReadIndex( rb, rbElementsRead );
+
+   UInt32 bytesRead = rbElementsRead * rb->elementSizeBytes;
+   ioData->mBuffers[0].mDataByteSize = bytesRead;
+   *ioNumberDataPackets = bytesRead / packetSize;
```
jmelas, line 2167: this line is the most important part!

jmelas:
Previous calculations were not wrong for the old API.
Previous API used one (in,out) variable for data size.
New API uses one (in, out) variable for data size as num of packets
and also requires read data size in bytes to be written in AudoBufferList.
I think I need to add one more commit to pass data size as number-of-packets
in AudioAudioConverterFillComplexBuffer

jmelas:
You can see the change in behaviour with latest commit 650fc6e if you run paex_read_write_wire with very verbose debug messages. Ring Buffer empty message is printed now less times because the data size provided previously was bigger (counted in bytes instead of num of packets)
NB: frames per packets for our stream is 1 so num of packets = num of frames

src/hostapi/coreaudio/pa_mac_core.c
xasopheno: Thank you so much for working on this. I'll be so glad when this lands and I can compile my language again!
https://github.com/xasopheno/WereSoCool

jmelas: You are welcome @xasopheno ! Nice project!

philburk (approved):
It build and runs fine on my Mac OS 10.15.6

paqa_devs passes
paex_read_write_wire passes

philburk:
I am assigning to Ross for final comments.
I will merge this tomorrow if no one objects.

RossBencina:
@philburk I object to you merging before I make my final comments! (I don't anticipate any issues but please let me review it first)

philburk:
> I object to you merging before I make my final comments! 

@RossBencina - OK, you had assigned it back to me without comment. So I thought you did not want to review it.
But rather than merge it, I assigned it back to you and asked again for a review.
I will let you do the merge.

src/hostapi/coreaudio/pa_mac_core.c
```diff
@@ -2646,20 +2665,20 @@ static PaError CloseStream( PaStream* s )
        if( stream->outputUnit ) {
           int count = removeFromXRunListenerList( stream );
           if( count == 0 )
-             AudioDeviceRemovePropertyListener( stream->outputDevice,
+             PaMacCore_AudioDeviceRemovePropertyListener( stream->outputDevice,
                                                 0,
                                                 false,
                                                 kAudioDeviceProcessorOverload,
-                                                xrunCallback );
+                                                xrunCallback, NULL ); //todo: do we need to pass actual node?
```
RossBencina, line 2672: What is the status of this TODO? it would be better to resolve this before merging the PR

src/hostapi/coreaudio/pa_mac_core.c
```diff
@@ -2646,20 +2665,20 @@ static PaError CloseStream( PaStream* s )
...
        if( stream->inputUnit && stream->outputUnit != stream->inputUnit ) {
           int count = removeFromXRunListenerList( stream );
           if( count == 0 )
-             AudioDeviceRemovePropertyListener( stream->inputDevice,
+              PaMacCore_AudioDeviceRemovePropertyListener( stream->inputDevice,
                                                 0,
                                                 true,
                                                 kAudioDeviceProcessorOverload,
-                                                xrunCallback );
+                                                xrunCallback, NULL ); //todo: do we need to pass actual node?
```
RossBencina, line 2681: What is the status of this TODO? it would be better to resolve this before merging the PR

src/hostapi/coreaudio/pa_mac_core.c
```diff
@@ -1131,31 +1132,31 @@ static OSStatus SetupDevicePropertyListeners( PaMacCoreStream *stream, AudioDevi
...
-                                   AudioDevicePropertyGenericListenerProc, stream );
-    
+
+    PaMacCore_AudioDeviceAddPropertyListener( deviceID, 0, isInput, kAudioStreamPropertyLatency,
+                                      AudioDevicePropertyGenericListenerProc, stream );
+    PaMacCore_AudioDeviceAddPropertyListener( deviceID, 0, isInput, kAudioDevicePropertyBufferFrameSize,
+                                      AudioDevicePropertyGenericListenerProc, stream );
+    PaMacCore_AudioDeviceAddPropertyListener( deviceID, 0, isInput, kAudioDevicePropertySafetyOffset,
+                                      AudioDevicePropertyGenericListenerProc, stream );
+
```
RossBencina, line 1145: I think you should fix the whitespace here so that it lines up with the `(` like it used to.

src/hostapi/coreaudio/pa_mac_core.c
```diff
@@ -1131,31 +1132,31 @@ static OSStatus SetupDevicePropertyListeners( PaMacCoreStream *stream, AudioDevi
...
+{
+    PaMacCore_AudioDeviceRemovePropertyListener( deviceID, 0, isInput, kAudioDevicePropertyActualSampleRate,
+                                         AudioDevicePropertyActualSampleRateListenerProc, stream );
+
+    PaMacCore_AudioDeviceRemovePropertyListener( deviceID, 0, isInput, kAudioDevicePropertyLatency,
+                                         AudioDevicePropertyGenericListenerProc, stream );
+    PaMacCore_AudioDeviceRemovePropertyListener( deviceID, 0, isInput, kAudioDevicePropertyBufferFrameSize,
+                                         AudioDevicePropertyGenericListenerProc, stream );
+    PaMacCore_AudioDeviceRemovePropertyListener( deviceID, 0, isInput, kAudioDevicePropertySafetyOffset,
+                                         AudioDevicePropertyGenericListenerProc, stream );
```
RossBencina, line 1159: I think you should fix the whitespace here so that it lines up with the `(` like it used to.

RossBencina (commented):
Question: `ringBufferIOProc()` has always been a complex beast. Are you confident that all code paths inside it have been adequately tested following the changes?

RossBencina:
@jmelas I'm handing this back to you to respond to my review comments.

src/hostapi/coreaudio/pa_mac_core.c
jmelas: It's ok as it is. I removed todo comments.

src/hostapi/coreaudio/pa_mac_core.c
jmelas: Done! Thanks!

jmelas:
About ringBufferIOProc: indeed it was too crypticly implements by reusing one variable for three different purposes.
I had to refactor it a bit to observe its behaviour in order to apply my knowledge to the new version.
@RossBencina yes I have tested both paths and they all work fine, I've also put debug messages in both paths.

jmelas:
Thank you all for feedback!
Another thing before we merge: since we ended up dropping support for the old API, do you think it makes sense to squash
commits "new hal" .. "remove support for OS X 10.4, rename functions pa_ -> PaMacCore_"?

src/hostapi/coreaudio/pa_mac_core.c
jmelas: Done thanks!

sharonlev:
+1 awaiting for this merge. (REF: https://github.com/spatialaudio/python-sounddevice/issues/299)

Be-ing:
I propose making a new release as soon as this is merged.

philburk (approved):

Be-ing:
@RossBencina is there anything left to do here?

PouncySilverkitten:
Similarly waiting for this merge and release!

Be-ing:
ping @philburk this critical bug fix has been sitting for over two weeks

xasopheno:
Awesome this was merged. Does this being merged mean that it's available via something like `brew install portaudio` or do I need to wait for a release. Thanks! D

Be-ing:
I will update vcpkg as soon as the new release is made.

xasopheno:
> I will update vcpkg as soon as the new release is made.

Is there somewhere I track progress towards new release? I'm grateful that folks are working on this. I'm blocked until a new release a cut.

meme:
> Awesome this was merged. Does this being merged mean that it's available via something like brew install portaudio or do I need to wait for a release. Thanks! D

You can `brew install portaudio --HEAD`

philburk:
I found vcpkg at https://vcpkg.readthedocs.io/en/latest/And PortAudio at:    https://github.com/microsoft/vcpkg/tree/master/ports/portaudio
I was not aware of this. We should document this on the Wiki.

> You can brew install portaudio --HEAD

Does that grab the current HEAD from GitHub? Or the old head from Assembla?

meme:
> Does that grab the current HEAD from GitHub? Or the old head from Assembla?

HEAD from GitHub. I used that command to get this fix on my host.

PR #317, opened 2020-10-04
build: add new Mac SDKs to configure

philburk:
Add SDK 10.14 and 10.15 to configure.in
    
Removed i386 and PPC architectures on Mac because they are
deprecated and make fails if they are used.
    
The configure file was generated from the configure.in
file by entering this on Linux:
    
    aclocal
    autoconf
    
Fixes 308

PR #359, opened 2020-12-01
Set macOS version minimum to 10.6 if using the macOS 11.0 SDK

hoolrory:
Required for using the changes made in #338, to get things working on macOS 11.0

Otherwise the configure script was setting the minimum to 10.4 still, which uses APIs incompatible with 11.0 (if compiling against 11.0 SDK)

For #218

### DIFF
--- a/configure
+++ b/configure
@@ -2759,7 +2759,7 @@ if test "${enable_debug_output+set}" = set; then :
   enableval=$enable_debug_output; if test "x$enableval" != "xno" ; then
 
 $as_echo "#define PA_ENABLE_DEBUG_OUTPUT /**/" >>confdefs.h
-
+$as_echo "Debug output enabled"
                   debug_output=yes
                fi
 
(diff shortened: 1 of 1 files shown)


philburk (approved):
Thanks

philburk (changes requested):
Wait. I just realized that the file "configure" should not be edited directly.
The edits should be made in "configure.in" and then "configure" is auto-generated.

See #317 for an example.
If you want, I can generate the files based on your idea. I have the tools installed on Linux.

jmelas:
Please note that current version of the SDK is 11.1 so we also need to add support for that!

jmelas:
Another idea which may help:
`xcrun --sdk macosx --show-sdk-version`
will display the sdk installed

Be-ing:
Obsoleted by #434

Be-ing:
I don't know what the right way to fix this is, but this is a great example of why autogenerated files should not be committed to version control.

Be-ing:
Superseded by #470

philburk:
@hoolrory - I see that you also added "arm64".
Was that to add support for the new ARM Macs?
If so then we should add that after #470 in a separate PR.

philburk:
Closing in favor of #470, which updates configure.in

Be-ing:
@hoolrory Do you have an M1 Mac? If so, can you please test #489? We need someone who has M1 hardware to confirm it works.

Issue #321, opened 2020-10-07
Building on Windows gives portaudio.def: error LNK2001: unresolved external symbol PaWasapi_UpdateDeviceList 


RossBencina:
When building using the `build/msvc/portaudio.sln` I get the following build failure:

    portaudio.def: error LNK2001: unresolved external symbol PaWasapi_UpdateDeviceList 

Steps to reproduce:
1. Clone the repo
2. Move ASIO SDK files in to place
3. Open `build/msvc/portaudio.sln` and click through the upgrade dialog box
4. Build the solution

**Expected behavior**
The project builds and a .dll is produced

**Actual behavior**
Get the error above.

**Desktop (please complete the following information):**
 - Windows 10
 - MSVC 2019 Pro 16.4.3

**Possible fix**

Remove the offending line from the .def file.

RossBencina:
@dmitrykos looks like the problem is this:

https://github.com/PortAudio/portaudio/blob/7db12305c32625bdf907afd55fb3c18c2083992e/src/hostapi/wasapi/pa_win_wasapi.c#L2361

Should be:

''''
PaError PaWasapi_UpdateDeviceList()
{
#if defined(PA_WASAPI_MAX_CONST_DEVICE_COUNT) && (PA_WASAPI_MAX_CONST_DEVICE_COUNT > 0)
	return UpdateDeviceList();
#else 
    return paNoError;
#endif
}
''''

P.S. 4 spaces no tabs!

RossBencina:
See also #323

RossBencina:
Fixed by #363

PR #363, opened 2020-12-14
Fix issue #321 broken MSVC build. 

RossBencina:
Always define PaWasapi_UpdateDeviceList(), even when it is a no-op, because it is mentioned in the .def file.

### DIFF
--- a/src/hostapi/wasapi/pa_win_wasapi.c
+++ b/src/hostapi/wasapi/pa_win_wasapi.c
@@ -2503,7 +2503,7 @@ PaError PaWasapi_UpdateDeviceList()
 #if defined(PA_WASAPI_MAX_CONST_DEVICE_COUNT) && (PA_WASAPI_MAX_CONST_DEVICE_COUNT > 0)
     return UpdateDeviceList();
 #else
-    return paNoError;
+    return paInternalError;
 #endif
 }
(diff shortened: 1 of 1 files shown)


RossBencina:
@dmitrykos waiting on your input on this. The build is broken : (

MichalPetryka:
Can confirm this together with #364 fixes building

dmitrykos:
The idea behind undefined `PaWasapi_UpdateDeviceList` in case `PA_WASAPI_MAX_CONST_DEVICE_COUNT == 0` is that if this function is used anywhere in the code the project will not compile because user was probably willing to use it to update the device list. I did not assume that DEF file would break compilation because symbol can not be weak in this situation.

The proposed change is ok but I would return an error instead of `paNoError` the `paInvalidHostApi` or `paInternalError` so that it will not silently succeed if internal implementation is missing (undefined). @RossBencina could you please adjust to this behavior?

RossBencina:
> could you please adjust to this behavior?

Done.

dmitrykos (approved):
Looks fine!

philburk (approved):
Thanks Ross.
The specific magic text below will tell GitHub to close 321 when this is merged:

Fixes #321

Issue #285, opened 2018-11-18
Deprecating def for exporting functions to a lib/dll.


PortAudio-admin:
*Issue created by Anonymous on Assembla*

Modified port audio to be able to compile and export all functions to a dll/lib file without having to use a def file.
The process will enable easy library usage and makes the code more robust.
Create a header file with the contents below:
```
#pragma once
/*
Additional defines added by alatnet
*/
//check which platform we are compiling to.
#if defined(__WIN32__) || defined(_WIN32) || defined(_WIN64)
    #define PA_WIN_PLATFORM
#elif defined(__APPLE__) || defined(__APPLE_CC__)
    #define PA_APPLE_PLATFORM
#elif defined(__GNUC__) && (!(defined(__WIN32__) || defined(_WIN32)) && !(defined(__APPLE__) || defined(__APPLE_CC__)))
    #define PA_NIX_PLATFORM
#else
    #error No Recognized Platform!
#endif

//see if we are using propriotary compilers.
#ifdef _MSC_VER
    #define PA_VISUAL_C
#elif defined(__BORLANDC__)
    #define PA_BORLANDC
#elif defined(__GNUC__) && (defined(__WIN32__) || defined(_WIN32))
    #define PA_MINGW
    #define PA_GNUC
#elif defined(__GNUC__)
    #define PA_GNUC
#endif

//check if we are compiling to 32 or 64 bit.
#if defined(__x86_64__) || defined(_M_X64) || defined(_WIN64)
    #define PA_ARCH_64
#else
    #define PA_ARCH_32
#endif

//see if we are compiling with unicode support.
#if defined(_UNICODE) || defined(UNICODE)
	#define PA_UNICODE
#endif

#if defined(PA_BUILD_DLL) || defined (PA_BUILD_SO) //set exports for building a dynamic library
    #ifdef PA_WIN_PLATFORM
        #if defined(PA_MINGW) || defined(PA_VISUAL_C)
            #define PA_DLL_DECLARE __declspec(dllexport)
            #define PA_DLL_EXTERN extern
            #define PA_DLL_C_EXTERN extern "C"
        #else
            #define PA_DLL_DECLARE __attribute__ ((dllexport))
            #define PA_DLL_EXTERN extern
            #define PA_DLL_C_EXTERN extern "C"
        #endif
    #else
        #ifdef PA_USE_DECLSPEC
            #define PA_DLL_DECLARE __declspec(dllexport)
            #define PA_DLL_EXTERN extern
[...]

PR #364, opened 2020-12-15
Update def files for #340

MichalPetryka:
PaWasapi_SetDefaultInterfaceId being renamed in #340 caused the project to error out when building with Visual Studio

### DIFF
--- a/build/msvc/portaudio.def
+++ b/build/msvc/portaudio.def
@@ -52,5 +52,5 @@ PaWasapi_ThreadPriorityRevert       @63
 PaWasapi_GetFramesPerHostBuffer     @64
 PaWasapi_GetJackCount               @65
 PaWasapi_GetJackDescription         @66
-PaWasapi_SetDefaultInterfaceId      @67
+PaWasapiWinrt_SetDefaultDeviceId    @67
 PaWasapi_SetStreamStateHandler      @68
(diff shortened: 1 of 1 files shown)


RossBencina:
Was `PaWasapi_SetDefaultInterfaceId` renamed to `PaWasapiWinrt_SetDefaultDeviceId` ?

dmitrykos:
Yes, it was renamed and 'Winrt' was added additionally to the name for purpose to denote that this API is used exclusivity for WinRT/UWP platform. Updating def file was missed, I am sorry for that.

dmitrykos:
In addition to `PaWasapiWinrt_SetDefaultDeviceId` another function `PaWasapiWinrt_PopulateDeviceList` needs to be added into the `portaudio.def` file, @MichalPetryka could you please add it too into the proposed change?

RossBencina:
@dmitrykos  did you see my other fix: https://github.com/PortAudio/portaudio/pull/363 ?

MichalPetryka:
Made the order there reflect the one in the header, not sure if that broke anything since I never used .def files before

RossBencina:
@MichalPetryka Please don't change the number of existing symbols. Once published the numbers should not change.

MichalPetryka:
Sorry, changed it back

RossBencina:
@MichalPetryka You can discuss __declspec here: https://github.com/PortAudio/portaudio/issues/285 My current opinion is that we are stuck with def files because making the necessary changes to support __declspec in portaudio.h is too painful, and if we remove .def we will break all clients that are depending on ordinals.

dmitrykos:
> @dmitrykos did you see my other fix: #363 ?

Sorry I overlooked it, replied in the thread.

The change in the DEF file is ok.

dmitrykos (approved):

PR #369, opened 2020-12-17
Fix warnings when compiling with MSVC 16.8.3

jmelas:
This is a simple PR that fixes the following warnings on Windows:
`pa_debugprint.c(1,1): warning C4828: The file contains a character starting at offset 0xa95 that is illegal in the current source character set (codepage 65001)`
`pa_win_coinitialize.c(136,53): warning C4389: '!=': signed/unsigned mismatch`
`pa_win_wdmks.c(168,1): warning C4005: 'DEFINE_GUIDEX': macro redefinition`
`pa_win_wdmks.c(4787): warning C4701: potentially uninitialized local variable 'hostInputSampleFormat' used`
`pa_win_wdmks.c(4787): warning C4701: potentially uninitialized local variable 'hostOutputSampleFormat' used`
`pa_win_wdmks.c(4562): warning C4701: potentially uninitialized local variable 'wfx' used`


src/os/win/pa_win_coinitialize.c
```diff
@@ -135,7 +135,7 @@ void PaWinUtil_CoUninitialize( PaHostApiTypeId hostApiType, PaWinUtilComInitiali
         DWORD currentThreadId = GetCurrentThreadId();
 		if( comInitializationResult->initializingThreadId != currentThreadId )
 		{
-			PA_DEBUG(("ERROR: failed PaWinUtil_CoUninitialize calling thread[%d] does not match initializing thread[%d]\n",
+			PA_DEBUG(("ERROR: failed PaWinUtil_CoUninitialize calling thread[%u] does not match initializing thread[%u]\n",
```
RossBencina, line 138: `DWORD` is an unsigned long:

https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-dtyp/262627d8-3418-4627-9218-4ffe110850b2

Most of the references that I have seen suggest to use `%lu`, for example:

https://www.davekb.com/browse_programming_tips:win32_printf_a_dword:txt
https://stackoverflow.com/questions/33671758/dword-format-specifier

RossBencina:
Please don't submit multiple unrelated changes as a single pull request. I would appreciate it if you could split this up so that each pull request includes only related changes. I ask you to do this for three reasons:

1. So that we can review and accept changes individually.
2. So that the trivial fixes can be accepted without question or delay.
3. So that our time is not wasted breaking up your requests.

Here is a list of the unrelated changes that you have combined in this request:

1. fix non-standard character in src/common/pa_debugprint.c 
2. #undef DEFINE_GUIDEX in  src/hostapi/wdmks/pa_win_wdmks.c (please explain why)
3. initialize local variables in  src/hostapi/wdmks/pa_win_wdmks.c to suppress compiler warning. (I don't think I accept this change as-is, there is no point initializing local variables just to shut up warnings if the values you are using do not result in reasonable behavior).
4. change PaWinUtilComInitializationResult::initializingThreadId to DWORD to match the native type of WIN32 thread id

jmelas:
OK I am closing this PR for now and will open new PRs.

Issue #370, opened 2020-12-18
WMME x64 build: warning C4312: 'type cast': conversion from 'int' to 'HWAVEIN' of greater size


RossBencina:
Currently 64 bit builds have the following warnings:

         D:\a\portaudio\portaudio\src\hostapi\wmme\pa_win_wmme.c(676,46): warning C4312: 'type cast': conversion from 'int' to 'HWAVEIN' of greater size [D:\a\portaudio\portaudio\build\msvc\portaudio.vcxproj]
         D:\a\portaudio\portaudio\src\hostapi\wmme\pa_win_wmme.c(685,46): warning C4312: 'type cast': conversion from 'int' to 'HWAVEIN' of greater size [D:\a\portaudio\portaudio\build\msvc\portaudio.vcxproj]
         D:\a\portaudio\portaudio\src\hostapi\wmme\pa_win_wmme.c(806,49): warning C4312: 'type cast': conversion from 'int' to 'HWAVEOUT' of greater size [D:\a\portaudio\portaudio\build\msvc\portaudio.vcxproj]
         D:\a\portaudio\portaudio\src\hostapi\wmme\pa_win_wmme.c(815,49): warning C4312: 'type cast': conversion from 'int' to 'HWAVEOUT' of greater size [D:\a\portaudio\portaudio\build\msvc\portaudio.vcxproj]
         D:\a\portaudio\portaudio\src\hostapi\wmme\pa_win_wmme.c(1020,40): warning C4312: 'type cast': conversion from 'UINT' to 'HWAVEIN' of greater size [D:\a\portaudio\portaudio\build\msvc\portaudio.vcxproj]
         D:\a\portaudio\portaudio\src\hostapi\wmme\pa_win_wmme.c(1024,42): warning C4312: 'type cast': conversion from 'UINT' to 'HWAVEOUT' of greater size [D:\a\portaudio\portaudio\build\msvc\portaudio.vcxproj]

RossBencina:
Fix merged by #371

PR #371, opened 2020-12-18
WMME Fix compiler warnings: 'type cast': conversion from 'int' to 'HWAVEIN' of greater size.

RossBencina:
See ticket #370


philburk (approved):
Fixes #370

PR #374, opened 2020-12-18
use variadic arguments in DEFINE_GUID also for MSVC

jmelas:
The fix for DEFINE_GUID that was recently added in pa_win_wdmks.c is also needed for the MSVC compiler.
Latest versions of the compiler support the `/Zc:preprocessor` option which is conforming to C99 and later standards
https://docs.microsoft.com/en-us/cpp/build/reference/zc-preprocessor?view=msvc-160
so it generates the exact same error as Clang (too many arguments error).

### DIFF
--- a/src/hostapi/wdmks/pa_win_wdmks.c
+++ b/src/hostapi/wdmks/pa_win_wdmks.c
@@ -163,7 +163,7 @@ Default is to use the pin category.
 #define DYNAMIC_GUID(data) {data}
 #define _NTRTL_ /* Turn off default definition of DEFINE_GUIDEX */
 #undef DEFINE_GUID
-#ifdef __clang__ /* clang-cl: avoid too many arguments error */
+#if defined(__clang__) || (defined(_MSVC_TRADITIONAL) && !_MSVC_TRADITIONAL) /* clang-cl and new msvc preprocessor: avoid too many arguments error */
   #define DEFINE_GUID(n, ...) EXTERN_C const GUID n = {__VA_ARGS__}
   #define DEFINE_GUID_THUNK(n, ...) DEFINE_GUID(n, __VA_ARGS__)
   #define DEFINE_GUIDEX(n) DEFINE_GUID_THUNK(n, STATIC_##n)
@@ -171,7 +171,7 @@ Default is to use the pin category.
   #define DEFINE_GUID(n, data) EXTERN_C const GUID n = {data}
   #define DEFINE_GUID_THUNK(n, data) DEFINE_GUID(n, data)
   #define DEFINE_GUIDEX(n) DEFINE_GUID_THUNK(n, STATIC_##n)
-#endif /* __clang__ */
+#endif /* __clang__, !_MSVC_TRADITIONAL */
 #endif
 
 #include <setupapi.h>


src/hostapi/wdmks/pa_win_wdmks.c
```diff
@@ -163,15 +163,9 @@ Default is to use the pin category.
...
-#ifdef __clang__ /* clang-cl: avoid too many arguments error */
-  #define DEFINE_GUID(n, ...) EXTERN_C const GUID n = {__VA_ARGS__}
-  #define DEFINE_GUID_THUNK(n, ...) DEFINE_GUID(n, __VA_ARGS__)
-  #define DEFINE_GUIDEX(n) DEFINE_GUID_THUNK(n, STATIC_##n)
-#else
-  #define DEFINE_GUID(n, data) EXTERN_C const GUID n = {data}
-  #define DEFINE_GUID_THUNK(n, data) DEFINE_GUID(n, data)
-  #define DEFINE_GUIDEX(n) DEFINE_GUID_THUNK(n, STATIC_##n)
-#endif /* __clang__ */
+#define DEFINE_GUID(n, ...) EXTERN_C const GUID n = {__VA_ARGS__}
```
philburk, line 166: Are there any compilers where this variadic code will fail?

Maybe the original #ifdef on line 166 should be extended with defined(__MSVC__) or something like that.

jmelas: Yes, we are already inside a `#ifdef _MSC_VER` block!

RossBencina: The variadic code should be avoided with MSVC (standard non-clang front end). That's the whole point of that code

RossBencina:
@jmelas We need to retain backward compatibility with the old preprocessor too. It looks like you have broken that.

The correct fix to the problem that you raise is to detect whether the old or new preprocessor is running, and if the new preprocessor, use the "clang-cl" branch of the #ifdef.

jmelas:
@RossBencina of course I have tested with both versions of the preprocessor and they both work fine!

RossBencina:
@jmelas: how many versions back of MSVC have you tested?

RossBencina:
Discussion of buggy __VARARGS__ in earlier versions of MSVC:

https://stackoverflow.com/questions/9183993/msvc-variadic-macro-expansion

I see no reason to break backward compatibility. Please detect if the new preprocessor is present and use the __clang__ branch in that case.

jmelas:
I have tested with MSVC 16.8.3 both with the /Zc:preprocessor flag and without.
Legacy MSVC behaviour was that it would treat __VA_ARGS__ as one argument, it does not mean that
you could not use __VA_ARGS_ at all. The macros in wdmks do not do weird stuff like the examples posted
so they work ok. They simply relied on the non-conforming behaviour of the MSVC to use just one argument.

jmelas:
There is an `_MSVC_TRADITIONAL` define that detects if the legacy preprocessor is used but I am afraid this define was added in recent version of the compiler after /Zc:preprocessor flag was invented.

RossBencina:
@jmelas What is your rationale for making this change? Does the existing code not work under some conditions?  or are you just doing this to "modernise" things or do you have another motivation?

jmelas:
@RossBencina indeed as I state in the PR description the existing code does not work if you compile with `/Zc:preprocessor`.

RossBencina:
Here are the facts:

I am defacto committer on Windows. I am not the maintainer of PA/WDMKS. (It would appear that we don't have one.) I am therefore inclined to be extremely conservative when accepting patches to code that I am not familiar with.

The proposed patch is intended to fix a problem that only occurs if you enable `/Zc:preprocessor` -- a switch that we do not enable by default. In other words, it is a fix to support an unsupported build configuration.

The patch replaces an MSVC-specific code path that has been known to work for years. I can only assume that the existing code is the way that it is for good reason, probably related to compiler and/or SDK compatibility.

The new patch may or may not be backwards compatible with older versions of MSVC and various versions of the platform SDK (note these macros are not used by PA, they are redefined prior to including platform SDK files). 

The new code has been tested with very few toolchain/SDK combinations. I was able to test on a version 8 (MSVC2005) setup that I still have installed and both the existing and proposed versions worked. However MSVC 5.0 is a common target for building maximally portable DLLs, and I have no way to test that.

To ensure backward compatibility, my preference remains to special-case the new code under the same `#ifdef` as the `__clang__` code. Please try replacing the preprocessor check for the new code path with:

```
#if defined(__clang__) || (defined(_MSVC_TRADITIONAL) and !_MSVC_TRADITIONAL) /* clang-cl and new msvc preprocessor: avoid too many arguments error */
```

jmelas:
Thank you Ross, this patch works for me!
However it's inevitable to drop support for too old compilers as we move forward.
For example if we drop support for MSVC compilers that do not support var args at all
we could then enhance all our PA_DEBUG messages that currently are using double parenthesis
and have them using var args instead!

jmelas:
BTW I've started using MSVC 6.0 with MFC back in 1998.
I remember I was always able to use the TRACE macro of MFC which was not using double parenthesis,
so apparently MSVC 6.0 was able to handle variadic arguments correctly!

RossBencina (approved):

philburk (approved):

PR #376, opened 2020-12-18
wasapi: Fix signed/unsigned mismatch warning.

jmelas:
In file pa_win_wasapi.c we compare signed SystemTimer::granularity with unsigned TIMECAPS::wPeriodMin.
This PR resolves the issue.

### DIFF
--- a/src/hostapi/wasapi/pa_win_wasapi.c
+++ b/src/hostapi/wasapi/pa_win_wasapi.c
@@ -841,8 +841,8 @@ static BOOL SystemTimer_SetGranularity(SystemTimer *timer, UINT32 granularity)
 
     if (timeGetDevCaps(&caps, sizeof(caps)) == MMSYSERR_NOERROR)
     {
-        if (timer->granularity < caps.wPeriodMin)
-            timer->granularity = caps.wPeriodMin;
+        if (timer->granularity < (INT32)caps.wPeriodMin)
+            timer->granularity = (INT32)caps.wPeriodMin;
     }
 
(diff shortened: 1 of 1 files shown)


RossBencina (commented):
Comparing with cast and then assigning without cast on the following line looks incorrect to me. At least use a cast on both lines.

jmelas:
assignment does implicit conversion

RossBencina:
Ok by me.

philburk (approved):

dmitrykos (changes requested):
@jmelas, would you please revert the last change back to d4744e7381d9681aa799b07494e787215ff2beac. We really do not need an additional variable here as it bloats the code.

dmitrykos:
Implicit conversion is ok in this case and does not lead to a compiler warning in my case but if there is something wrong with it would you please comment.

jmelas:
@dmitrykos I agree with you but this is a change requested by @RossBencina and I would not like to go back and forth,
so let's hear @philburk and @RossBencina thoughts on this.

dmitrykos:
> At least use a cast on both lines

@jmelas we can follow this proposal of @RossBencina, e.g. just add an explicit cast inside the assignment like this:

```c
if (timer->granularity < (INT32)caps.wPeriodMin)
    timer->granularity = (INT32)caps.wPeriodMin;
```

Dangling temporary variable just for an assignment looks too bulky to me. Also your commits need to be squashed into one to be merged succesfully.

jmelas:
Done thanks!

dmitrykos (approved):
Looks great, thank you very much!

PR #377, opened 2020-12-18
wdmks: suppress uninitialized variable warnings

jmelas:
It seems an attempt to suppress "uninitialized variable used" warnings was made previously but was not complete.
This PR fixes this.

NB: I always compile with /WX (treat warnings as errors) that's why I am a bit pedantic with warnings.
Another idea if you think it's meaningful to leave some variables uninitialized is to document why and
suppress warnings using `#pragma warning push/#pragma warning( disable: ...)/# pragma warning pop`

### DIFF
--- a/src/hostapi/wdmks/pa_win_wdmks.c
+++ b/src/hostapi/wdmks/pa_win_wdmks.c
@@ -4340,9 +4340,10 @@ static PaError OpenStream( struct PaUtilHostApiRepresentation *hostApi,
     PaWinWdmStream *stream = 0;
     /* unsigned long framesPerHostBuffer; these may not be equivalent for all implementations */
     PaSampleFormat inputSampleFormat, outputSampleFormat;
-    PaSampleFormat hostInputSampleFormat, hostOutputSampleFormat;
+    PaSampleFormat hostInputSampleFormat = paInt16;
+    PaSampleFormat hostOutputSampleFormat = paInt16;
     int userInputChannels,userOutputChannels;
-    WAVEFORMATEXTENSIBLE wfx;
+    WAVEFORMATEXTENSIBLE wfx = { 0 };
 
     PA_LOGE_;
     PA_DEBUG(("OpenStream:sampleRate = %f\n",sampleRate));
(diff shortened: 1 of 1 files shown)


philburk (approved):

RossBencina:
I already rejected this change once. I continue to reject it.

In particular, the following looks wrong. It looks to me like there is a bug. I will check later.

```
 PaSampleFormat hostInputSampleFormat = paInt16;
 PaSampleFormat hostOutputSampleFormat = paInt16;
```

RossBencina:
I do not consider it acceptable to pre-initialize variables with random values in order to "suppress" uninitialized variable warnings. The goal should not be to suppress anything, but to understand what the compiler is telling you and to fix the underlying issue. As there has been no attempt to understand what the compiler is telling you in this instance I am inclined to reject this pull request.

It is possible that this is a *spurious* warning, but in that case the correct fix is to use a `#pragma` to suppress the warning not just assign random values to variables.

RossBencina:
I propose the following as a better fix:

https://github.com/PortAudio/portaudio/pull/383

jmelas:
I completely understand what the compiler is saying: it detected an execution path (a branch) were a variable is used without having been initialized ie it has a completely random value in runtime. This is a very common source of bugs and not a spurious warning.

Another very common way to fix this is to remove the "impossible" branch and replace it with an assert.

RossBencina:
Another way to fix it is to initialize the variable on the branch where it is not initialized, which is what I'm proposing. That also follows the general convention of the surrounding code,

jmelas:
Applied changes suggested by @RossBencina in PR 383.
Now the only thing that puzzles me is if inputSampleFormat and outputSampleFormat 
should also be initialized to zero in the "error" branch.

philburk (commented):
OK, removing my approval. Let's go with #383 instead.

philburk:
I am fine with whatever John and Ross agree on.

RossBencina:
Based on Phil's comments I'm closing this one in favour of #383

PR #379, opened 2020-12-20
Better paqa devs

philburk:
Only test default device.

Use sine wave instead of sawtooth because it is easier to hear glitches.
    
Add some options:
        -i - Test INPUT only.
        -o - Test OUTPUT only.


philburk:
This is polluted with lots of unrelated commits!
CLOSING!

PR #380, opened 2020-12-20
Improve paqa_devs

philburk:
Use sine wave instead of sawtooth because it is easier to hear glitches.
    
Add some options:
        -i - Test INPUT only.
        -o - Test OUTPUT only.


philburk:
Now it is pulling in even more unrelated commits.
CLOSING!

PR #381, opened 2020-12-20
paqa_devs: test default device

philburk:
Use sine wave instead of sawtooth because it is
easier to hear glitches.

Add some options:
    -i - Test INPUT only.
    -o - Test OUTPUT only.

### DIFF
--- a/qa/paqa_devs.c
+++ b/qa/paqa_devs.c
@@ -8,8 +8,8 @@
     the sine wave outputs.
 
     @author Phil Burk  http://www.softsynth.com
-    
-    Pieter adapted to V19 API. Test now relies heavily on 
+
+    Pieter adapted to V19 API. Test now relies heavily on
     Pa_IsFormatSupported(). Uses same 'standard' sample rates
     as in test pa_devs.c.
 */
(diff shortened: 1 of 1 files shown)


qa/paqa_devs.c
```diff
@@ -1,13 +1,13 @@
 /** @file paqa_devs.c
-	@ingroup qa_src
+    @ingroup qa_src
     @brief Self Testing Quality Assurance app for PortAudio
- 	Try to open each device and run through all the
- 	possible configurations. This test does not verify
+    Try to open the default devices and run through all the
+    possible configurations. This test does not verify
```
RossBencina, line 5: If you are going to update this comment, I recommend that you say what it actually does. e.g.:

```
Try to open devices and run through all possible configurations. By default, open only the default devices. Command line options support opening every device, or all input devices, or all output devices.
```

RossBencina (approved):
Seems OK aside from fixing the comment.

I don't like that it is not entirely clear that testing input also works with full duplex devices, and testing output also works with full duplex devices. Does the code correctly zero output when testing input on a full duplex device?

PR #383, opened 2020-12-21
wdmks: Fix 'uninitialized variable used' warnings.

RossBencina:
Alternative solution for PR #377.

### DIFF
--- a/src/hostapi/wdmks/pa_win_wdmks.c
+++ b/src/hostapi/wdmks/pa_win_wdmks.c
@@ -4380,7 +4380,7 @@ static PaError OpenStream( struct PaUtilHostApiRepresentation *hostApi,
     else
     {
         userInputChannels = 0;
-        inputSampleFormat = hostInputSampleFormat = paInt16; /* Suppress 'uninitialised var' warnings. */
+        inputSampleFormat = paInt16; /* Suppress 'uninitialised var' warnings. */
     }
 
     if( outputParameters )
@@ -4415,7 +4415,7 @@ static PaError OpenStream( struct PaUtilHostApiRepresentation *hostApi,
     else
     {
         userOutputChannels = 0;
-        outputSampleFormat = hostOutputSampleFormat = paInt16; /* Suppress 'uninitialized var' warnings. */
+        outputSampleFormat = paInt16; /* Suppress 'uninitialized var' warnings. */
     }
 
     /* validate platform specific flags */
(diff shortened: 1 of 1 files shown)


jmelas:
I confirm this eliminates warnings for hostInputSampleFormat and hostOutputSampleFormat.
Still there is a warning for
`WAVEFORMATEXTENSIBLE wfx;`
Would you be ok with initializing it on declaration ie
`WAVEFORMATEXTENSIBLE wfx = { 0};`

jmelas:
Also do you think it makes sense to use a defined `paSampleFormatUnkown = 0` value instead?

RossBencina:
Regarding the use of `paSampleFormatUnkown = 0 `. The value should never be used, so I think it makes sense to use an invalid value yes. If the 0 were ever used it should (in theory) hit an assert in the buffer processor.

jmelas:
I think the code desperately needs a paSampleFormatNone = 0 define
for the following reasons:

1) In many places paSampleFormat is treated as a bit mask to represent multiple formats. So 0 looks like a great 
choice to indicate "none"

2) In wdmks.OpenStream in the error branches we initialize in/outputSampleFormat to paInt16 (why?)
while @RossBencina suggested hostIn/OutputSampleFormat to 0 usinng a type cast.

3) In wasapi.WaveToPaFormat we return paCustomFormat as a last resort. That looks suspicious.

4) In converters.PaUtil_SelectClosestAvailableFormat we return an error code (paSampleFormatNotSupported)
while the function's return type is PaSampleFormat. We should not mixup types!

RossBencina:
@jmelas please do not start suggesting new features inside of pull requests. If you'd like to create an issue to discuss new changes please feel free. The goal is just to make progress on the issue at hand. If there are no other objections I will merge it.

RossBencina:
I have reviewed the initialization of `wfx` again and I don't think it makes sense to initialize it in order to suppress an uninitialized variable warning. `wfx` is a temporary variable, and each time that it is used, the usage is to pass it to a function name `PaWin_InitializeWaveFormatExtensible`. It is very clear that it is reused multiple times, and that it is initialized on first use. 

Ways forward on this are:

- If there is a pragma that can disable the warning for just this one variable that would be OK.
- If `PaWin_InitializeWaveFormatExtensible` is not initializing the whole struct, (i.e. leaving uninitialized fields) that is a bug and must be fixed within `PaWin_InitializeWaveFormatExtensible`
- Otherwise, it's a spurious compiler warning only thrown with /W4 which we do not support and my vote is to leave it as-is.

philburk (approved):

jmelas:
@RossBencina all calls to `PaWin_InitializeWaveFormatExtensible' are inside `if` blocks or `while` blocks.
If for example you think that `while (hostInputSampleFormat <= paUInt8)` is guaranteed to execute at least once,
it is not obvious to the compiler. Perhaps this `while` should be written as `do {} while`.

jmelas:
BTW there is a FIXME comment in `PaUtil_SelectClosestAvailableFormat` which states that we should ensure that we return
a Sample Format within the valid range.

PR #384, opened 2020-12-21
Fix a GCC warning due to unsigned substraction

jcelerier:
Given

    unsigned int reqRate, setRate, deviation;

then 

    setRate - reqRate

is also an unsigned int - the code works "by chance" as `abs` casts that to an int which overflows and is technically undefined behaviour, for instance for setRate = 44100 and reqRate = 48000.

Thus cast them to int when computing the absolute value to ensure no UB.


RossBencina (approved):
Looks OK to me

philburk (approved):

PR #389, opened 2020-12-22
pa_converters include math if PA_USE_C99_LRINTF

jmelas:
pa_converters.c did not include math.h in case PA_USE_C99_LRINTF is defined
so the compiler could not find the definition for lrintf

### DIFF
--- a/src/common/pa_converters.c
+++ b/src/common/pa_converters.c
@@ -66,6 +66,9 @@
 #include "pa_endianness.h"
 #include "pa_types.h"
 
+#ifdef PA_USE_C99_LRINTF
+#include <math.h>
+#endif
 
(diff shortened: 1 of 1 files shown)


RossBencina (approved):
Seems fine. Although the actual use of lrint looks suspect to me.

RossBencina:
FYI here's what I think about the `lrintf` usage: https://github.com/PortAudio/portaudio/issues/390 I wouldn't enable it.

RossBencina:
This is pending some consensus on #390. I think either we merge this PR or we strip out the `PA_USE_C99_LRINTF` code.

philburk:
I think we should close this and then remove the PA_USE_C99_LRINTF code.

RossBencina:
I've submitted #403 as an alternative to this. I'll close this now.

Issue #200, opened 2011-11-20
PortAudio should use UTF8 for character encoding of public API strings


PortAudio-admin:
*Issue created by @RossBencina*

After some discussion with Rob we concluded that strings passed in PortAudio public APIs should encode multibyte characters (device names etc) as UTF-8.

PortAudio implementations may store strings internally however they like.

This applies to:
- device names
- native error messages

Relevant strings can originate from host API implementations, and from platform-specific common code. As of now, the following APIs definitely return UTF8 strings. The others I'm unsure of. Please discuss below.

Host API code:

- [ ] alsa
- [ ] asihpi
- [ ] asio
- [ ] audioio
- [ ] coreaudio
- [x] dsound
- [ ] jack
- [ ] oss
- [ ] wasapi
- [ ] wdmks
- [x] wmme

OS support code:

- [ ] unix
- [ ] win
- [ ] mac

jmelas:
This can be closed now

RossBencina:
@jmelas Before closing I'd like to be sure that the issue is resolved for all host APIs. I've added a checklist to the head of this ticket. Aside from DSound and MME do you know with confidence which other host APIs are correctly returning UTF8 strings?

arkadijs:
Would be also nice for API documentation to reflect this.

PR #391, opened 2020-12-22
Fix to use UTF-8 text instead of ANSI in MME and DSound

### LINKED ISSUE #224: Patch for Windows DirectSound and WMME UTF-8 device names

from Tobias Erichsen

Patch in attachment

Original discussion thread here:

http://comments.gmane.org/gmane.comp.audio.portaudio.devel/10876

MichalPetryka:
Fixes MME and DSound ANSI characters under MBCS builds. Despite what MS docs say, MME actually does provide non-TCHAR methods (noticed it when Visual Studio IntelliSense suggested them).
While fixing the device names, I've noticed that MME error macros also used ANSI, so i fixed them aswell.

Fixes #224.

Tested on my machine (in my test app which i'm using for integrating PortAudio in our Unity game, thus we use a custom ordering and include the API and Exclusive in the name):
```
VoiceMeeter Output (VB-Audio VoiceMeeter VAIO) (Windows WASAPI)
VoiceMeeter Output (VB-Audio VoiceMeeter VAIO) (Windows WASAPI) - Exclusive
Trust Starzz (Realtek High Definition Audio) (Windows WASAPI)
Trust Starzz (Realtek High Definition Audio) (Windows WASAPI) - Exclusive
Podstawowy sterownik przechwytywania dźwięku (Windows DirectSound)
Trust Starzz (Realtek High Definition Audio) (Windows DirectSound)
VoiceMeeter Output (VB-Audio VoiceMeeter VAIO) (Windows DirectSound)
Mapowanie dźwięku Microsoft - Input (MME)
Trust Starzz (Realtek High Defi (MME)
VoiceMeeter Output (VB-Audio Vo (MME)
Mikrofon (Realtek HD Audio Mic input) (Windows WDM-KS)
FrontMic (Realtek HD Audio Front Mic input) (Windows WDM-KS)
Wejście liniowe (Realtek HD Audio Line input) (Windows WDM-KS)
Miks stereo (Realtek HD Audio Stereo input) (Windows WDM-KS)
VoiceMeeter Output (VoiceMeeter vaio) (Windows WDM-KS)
ASIO4ALL v2 (ASIO)
Voicemeeter Virtual ASIO (ASIO)
```
As you can see, polish characters like ź, ę are all correctly converted.

### DIFF
--- a/src/hostapi/wmme/pa_win_wmme.c
+++ b/src/hostapi/wmme/pa_win_wmme.c
@@ -295,7 +295,7 @@ static signed long GetStreamWriteAvailable( PaStream* stream );
         wchar_t mmeErrorTextWide[ MAXERRORLENGTH ];                     \
         char mmeErrorText[ MAXERRORLENGTH ];                            \
         waveInGetErrorTextW( mmresult, mmeErrorTextWide, MAXERRORLENGTH );   \
-        WideCharToMultiByte( CP_UTF8, WC_COMPOSITECHECK | WC_DEFAULTCHAR,\
+        WideCharToMultiByte( CP_UTF8, 0,                                \
             mmeErrorTextWide, -1, mmeErrorText, MAXERRORLENGTH, NULL, NULL );  \
         PaUtil_SetLastHostErrorInfo( paMME, mmresult, mmeErrorText );   \
     }
@@ -305,7 +305,7 @@ static signed long GetStreamWriteAvailable( PaStream* stream );
         wchar_t mmeErrorTextWide[ MAXERRORLENGTH ];                     \
         char mmeErrorText[ MAXERRORLENGTH ];                            \
         waveOutGetErrorTextW( mmresult, mmeErrorTextWide, MAXERRORLENGTH );  \
-        WideCharToMultiByte( CP_UTF8, WC_COMPOSITECHECK | WC_DEFAULTCHAR,\
+        WideCharToMultiByte( CP_UTF8, 0,                                \
             mmeErrorTextWide, -1, mmeErrorText, MAXERRORLENGTH, NULL, NULL );  \
         PaUtil_SetLastHostErrorInfo( paMME, mmresult, mmeErrorText );   \
     }
--- a/src/hostapi/wmme/pa_win_wmme.c
+++ b/src/hostapi/wmme/pa_win_wmme.c
@@ -290,24 +290,24 @@ static signed long GetStreamWriteAvailable( PaStream* stream );
 
 /* macros for setting last host error information */
 
-#define PA_MME_SET_LAST_WAVEIN_ERROR( mmresult ) \
-    {                                                                   \
-        wchar_t mmeErrorTextWide[ MAXERRORLENGTH ];                     \
-        char mmeErrorText[ MAXERRORLENGTH ];                            \
-        waveInGetErrorTextW( mmresult, mmeErrorTextWide, MAXERRORLENGTH );   \
-        WideCharToMultiByte( CP_UTF8, 0,                                \
-            mmeErrorTextWide, -1, mmeErrorText, MAXERRORLENGTH, NULL, NULL );  \
-        PaUtil_SetLastHostErrorInfo( paMME, mmresult, mmeErrorText );   \
+#define PA_MME_SET_LAST_WAVEIN_ERROR( mmresult )                              \
+    {                                                                         \
+        wchar_t mmeErrorTextWide[ MAXERRORLENGTH ];                           \
+        char mmeErrorText[ MAXERRORLENGTH ];                                  \
+        waveInGetErrorTextW( mmresult, mmeErrorTextWide, MAXERRORLENGTH );    \
+        WideCharToMultiByte( CP_UTF8, 0, mmeErrorTextWide, -1,                \
+            mmeErrorText, MAXERRORLENGTH, NULL, NULL );                       \
+        PaUtil_SetLastHostErrorInfo( paMME, mmresult, mmeErrorText );         \
     }
 
-#define PA_MME_SET_LAST_WAVEOUT_ERROR( mmresult ) \
-    {                                                                   \
-        wchar_t mmeErrorTextWide[ MAXERRORLENGTH ];                     \
-        char mmeErrorText[ MAXERRORLENGTH ];                            \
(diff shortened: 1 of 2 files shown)


src/hostapi/wmme/pa_win_wmme.c
```diff
@@ -298,14 +290,12 @@ static signed long GetStreamWriteAvailable( PaStream* stream );
...
 
-#ifdef UNICODE
-
 #define PA_MME_SET_LAST_WAVEIN_ERROR( mmresult ) \
     {                                                                   \
         wchar_t mmeErrorTextWide[ MAXERRORLENGTH ];                     \
         char mmeErrorText[ MAXERRORLENGTH ];                            \
-        waveInGetErrorText( mmresult, mmeErrorTextWide, MAXERRORLENGTH );   \
-        WideCharToMultiByte( CP_ACP, WC_COMPOSITECHECK | WC_DEFAULTCHAR,\
+        waveInGetErrorTextW( mmresult, mmeErrorTextWide, MAXERRORLENGTH );   \
```
jmelas, line 297: This is great!
I would simply fix the slashes alignment here

jmelas (approved):

RossBencina:
Related #200

src/hostapi/wmme/pa_win_wmme.c
```diff
@@ -298,45 +290,25 @@ static signed long GetStreamWriteAvailable( PaStream* stream );
...
-        wchar_t mmeErrorTextWide[ MAXERRORLENGTH ];                     \
-        char mmeErrorText[ MAXERRORLENGTH ];                            \
-        waveInGetErrorText( mmresult, mmeErrorTextWide, MAXERRORLENGTH );   \
-        WideCharToMultiByte( CP_ACP, WC_COMPOSITECHECK | WC_DEFAULTCHAR,\
-            mmeErrorTextWide, -1, mmeErrorText, MAXERRORLENGTH, NULL, NULL );  \
-        PaUtil_SetLastHostErrorInfo( paMME, mmresult, mmeErrorText );   \
+#define PA_MME_SET_LAST_WAVEIN_ERROR( mmresult )                              \
+    {                                                                         \
+        wchar_t mmeErrorTextWide[ MAXERRORLENGTH ];                           \
+        char mmeErrorText[ MAXERRORLENGTH ];                                  \
```
RossBencina, line 296: If an error is less than `MAXERRORLENGTH` wide characters long is it necessarily less than `MAXERRORLENGTH` bytes long?

src/hostapi/wmme/pa_win_wmme.c
```diff
@@ -298,45 +290,25 @@ static signed long GetStreamWriteAvailable( PaStream* stream );
...
-        wchar_t mmeErrorTextWide[ MAXERRORLENGTH ];                     \
-        char mmeErrorText[ MAXERRORLENGTH ];                            \
-        waveOutGetErrorText( mmresult, mmeErrorTextWide, MAXERRORLENGTH );  \
-        WideCharToMultiByte( CP_ACP, WC_COMPOSITECHECK | WC_DEFAULTCHAR,\
-            mmeErrorTextWide, -1, mmeErrorText, MAXERRORLENGTH, NULL, NULL );  \
-        PaUtil_SetLastHostErrorInfo( paMME, mmresult, mmeErrorText );   \
+#define PA_MME_SET_LAST_WAVEOUT_ERROR( mmresult )                             \
+    {                                                                         \
+        wchar_t mmeErrorTextWide[ MAXERRORLENGTH ];                           \
+        char mmeErrorText[ MAXERRORLENGTH ];                                  \
```
RossBencina, line 306: As above, if an error is less than `MAXERRORLENGTH` wide characters long is it necessarily less than `MAXERRORLENGTH` bytes long?

RossBencina (approved):
I think that this is an improvement over the existing code.

src/hostapi/wmme/pa_win_wmme.c
MichalPetryka: It isn't, to be safe allocating either MAXERRORLENGTH * 4 or dynamically allocating the buffer with required size would be needed.

RossBencina: I'd be happy to take the current change as a positive step forward, but we should probably define real functions that do the conversion with dynamic allocation.

RossBencina:
@philburk could you please take a look over this as a sanity check. I think it's OK to merge, but see the above conversation about static array dimensioning.

src/hostapi/wmme/pa_win_wmme.c
philburk: If the wide character string is close to MAXERRORLENGTH and it contains several extended characters then the UTF8 representation could easily be longer than MAXERRORLENGTH.
So the string may get truncated. I don't know if it will be NUL terminated in that case.
But luckily the strings are coming from the OS and from devices and not from the user. So an exploit is unlikely.
WideCharToMultiByte() will return the length if cbMultiByte is zero. That can be used to do the allocation.

philburk (approved):

dechamps:
I noticed PortAudio master now contains a commit 6f90db9612f80c9dd781969959820cb1610b8184 named "Update pa_win_wmme.c" which is not great in terms of history readability. @philburk next time you might want to double check what the chain of commits looks like before merging; it looks like @MichalPetryka designed this particular pull request to be squashed.

MichalPetryka:
Yeah, in most of projects I commited to PRs like this were always squashed, so I didn't bother with a more explainatory commit name.

philburk:
Some people prefer not to have their commits squashed because they may contains discrete standalone changes. I did not see any comment requesting a squash. So I merged them as is.

Should we have a policy to squash all PRs unless explicitly asked not to squash?

dechamps:
I think the policy should be to do whatever makes sense with respect to the particular PR. In this case, either @MichalPetryka should have been asked to rearrange his commits to make them more readable in context, or the person doing the merge should do a quick review of the commits and realize that they would look better squashed before merging.

Since we're on the topic of merge policy, I would also suggest forbidding the use of "merge commits" altogether, requiring rebases instead. Merge commits make history way more complicated and harder to read. This suggestion is mostly intended for @dmitrykos who has been adding merge commits to master recently (see for example #421). I believe GitHub has an option for disabling merge commits.

PR #394, opened 2020-12-25
Add .editorconfig

### LINKED ISSUE #393: Provide a .editorconfig file
**Describe the bug**
PortAudio currently doesn't provide an .editoconfig file. Editorconfig is an cross-IDE standard providing data about project codestyle such as indentation style. It's supported by many editors out of the box (for example Visual Studio) and many more with an extension (for example Visual Studio Code).

**To Reproduce**
Load into an editor with for example TABs set as default indentation.

**Expected behavior**
.editorconfig overrides TABs with SPACEs for this project

**Actual behavior**
Nothing, tabs are still used

 - OS: n/a
 - OS Version n/a
 - PortAudio version: stable, nightly snapshot (which?), current (please give date and/or Git hash): current master (25.12.2020)
 - If Windows or Linux, which Host API (e.g. WASAPI): n/a

**Additional context**
Providing an editorconfig file would make contributing easier due to removing the need to manually fix indentation and would reduce the amount of PRs with improper formatting.

MichalPetryka:
Fixes #393.
Please verify if all of the settings are correct, i'm mostly unsure about trim_trailing_whitespace and insert_final_newline.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,5 +1,4 @@
 # EditorConfig info: https://editorconfig.org/
-
 root = true
 
 [Makefile*]
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,14 @@
+# EditorConfig info: https://editorconfig.org/
+
+root = true
+
+[Makefile*]
+indent_style = tab
+
+[*.{c,cpp,h}]
+indent_style = space
+indent_size = 4
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+max_line_length = 100
--- a/.editorconfig
+++ b/.editorconfig
@@ -10,4 +10,4 @@ indent_size = 4
 charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
-max_line_length = 100
+max_line_length = 120
--- a/.editorconfig
+++ b/.editorconfig
@@ -4,7 +4,7 @@ root = true
 [Makefile*]
 indent_style = tab
 
-[*.{c,cpp,h}]
+[*.{c,cpp,h,cxx,hxx}]
 indent_style = space
 indent_size = 4
 charset = utf-8


philburk:
Thanks for doing this.

> unsure about trim_trailing_whitespace and insert_final_newline.

I am in favor of both those options.

Is this at the top of tree because editors will traverse up the tree to find a .editorconfig file?

MichalPetryka:
> Thanks for doing this.
> 
> > unsure about trim_trailing_whitespace and insert_final_newline.
> 
> I am in favor of both those options.
> 
> Is this at the top of tree because editors will traverse up the tree to find a .editorconfig file?

```
When opening a file, EditorConfig plugins look for a file named .editorconfig in the directory of the opened file and in every parent directory. A search for .editorconfig files will stop if the root filepath is reached or an EditorConfig file with root=true is found.

EditorConfig files are read top to bottom and the most recent rules found take precedence. Properties from matching EditorConfig sections are applied in the order they were read, so properties in closer files take precedence.
```

So it is supposed to traverse up, although I'm pretty sure that despite GitHubs edit saying it supports editorconfig, it doesn't actually traverse up. `root = true` is supposed to stop traversing as you can see.

MichalPetryka:
Also, what about `charset = utf-8`? Most Linux users will say that utf-8 is the correct choice, but afaik Visual Studio is pretty annoying and still usually adds BOM to files (`charset = utf-8-bom`)

MichalPetryka:
image
Resharper shows that the file is properly loaded.

philburk:
Thanks for the explanation for the search path. 
I did an experiment with BBEdit. It found and used this config file.

I loaded "examples/paex_write_sine_nonint.c" changed one line and saved it.
I notice that the editor cleaned up trailing white space when it saved the file.
That is good but it resulted in many style changes showing up in git diff.

I like to keep style changes separate from read code changes.

I am in favor of merging this file. But I think we should make a pass over the source tree
and cleanup white space style issues before we merge this. Otherwise small commits
will be obscured by all the style fixes.

I don't want to create a lot of merge conflicts.
So maybe the order should be:
1. merge outstanding code changes
2. apply style changes to entire code base
3. merge this change

MichalPetryka:
Planned to cleanup all whitespaces after this gets merged, but sure.

MichalPetryka:
I've done an automated whitespace cleanup, as you can see, it's quite a lot of changes https://github.com/MichalPetryka/portaudio/commit/17cee12df1b8ae12b6a2b9cb86cc815568f0a63e

.editorconfig
```diff
@@ -0,0 +1,13 @@
+# EditorConfig info: https://editorconfig.org/
+root = true
+
+[Makefile*]
+indent_style = tab
+
+[*.{c,cpp,h}]
+indent_style = space
+indent_size = 4
+charset = utf-8
```
RossBencina, line 10: I would have thought that 7-bit ASCII was a safer choice here. The common code needs to be portable to all possible targets. utf-8 does not seem conservative enough to me,

.editorconfig
```diff
@@ -0,0 +1,13 @@
...
+trim_trailing_whitespace = true
+insert_final_newline = true
+max_line_length = 100
```
RossBencina, line 13: I don't think we should set a max line length. We can quibble about what the *guidance* should be, but I think this should be left out of .editorconfig

RossBencina (commented):
Overall I'm in favour of adding a .editorconfig.

However, unless we can be sure that all of the code in the repo already conforms I think adding a .editorconfig right now is a not a good idea, as it will trigger a flurry of garbage whitespace commits interspersed with other non-whitespace changes. 

My thinking is we need the following (which I have been thinking about), added in this order:
1. GitHub action that lints for all whitespace problems, and outputs warnings for each issue.
2. Fix each issue in separate whitespace-only commits.
3. add .editorconfig

This could be "staged" in the sense that we could start with a minimal set of checks (say just the 4-spaces indent rule) and add other changes as we go.

.editorconfig
MichalPetryka: Editorconfig only supports UTF-8, not ASCII, due to it being compatible in ASCII range, checking for non-ASCII characters should be rather done in some GitHub Action

.editorconfig
MichalPetryka: Pretty sure most editors treat this as a suggestion rather than hard limit.

RossBencina:
See also #398

.editorconfig
RossBencina: It looks as though Phil added the "100 char line rule" to our coding guidelines in October without consultation. I don't know where it came from, seems arbitrary to me. Why not 80 or 120? Or take a survey of the code and see what is currently used. Personally I try to stick to 80.

MichalPetryka: Afaik, 120 is a more common limit

RossBencina:
I have started submitting pull requests to clean up whitespace so that it conforms to the specifications of the proposed `.editorconfig`. Once our source is consistent we can merge the `.editorconfig`. I will maintain a list of problem files at #398

.editorconfig
```diff
@@ -0,0 +1,13 @@
+# EditorConfig info: https://editorconfig.org/
+root = true
+
+[Makefile*]
+indent_style = tab
+
+[*.{c,cpp,h}]
```
philburk, line 7: Our current C++ binding files use a .cxx or .hxx suffix. So we should probably add those.

Note that we are probably going to move the C++ bindings to another repo. But there may be C++ code in the host implementations.

philburk (changes requested):
I am OK with whatever line length Ross is OK with.

philburk (approved):

RossBencina:
@MichalPetryka could we go with 120 characters please? `pa_win_wmme.c` has many lines over 100 but only a few over 120. I'll update the Wiki

MichalPetryka:
@RossBencina done

RossBencina (approved):

Issue #398, opened 2020-12-27
Add a CI task that checks whitespace


RossBencina:
Related to the ongoing discussion of adding .editorconfig #393 

I have created a draft script for checking whitespace, here:
https://github.com/RossBencina/portaudio/blob/whitespace-linter/whitelint.py

We could run this as a CI task to verify that new PRs do not have whitespace issues.

Note that it is quite hard to check that all indenting is a multiple of 4 spaces, so I have not yet done that. Problems include:
1. You need to ignore whitespace inside comments
2. Sometimes non-multiple-of-4 is used in mac-style multi-line function calls

Current results of the script are discussed at #474

RossBencina:
Should also check for non-ASCII characters

MichalPetryka:
Also check for a final endline (it's required for some Linux editors).

philburk:
Nice. I think this will be handy as long as it allows for alignment of continuation lines with the first parameter.

dechamps:
My understanding is that clang-format (and maybe astyle, perhaps others) can do all these checks and many more, and wouldn't require a custom script.

Be-ing:
With Mixxx we [use pre-commit](https://github.com/mixxxdj/mixxx/blob/main/.pre-commit-config.yaml) with clang-format and some other tools so issues are automatically fixed locally when committing. We also run pre-commit in a [GitHub Actions workflow](https://github.com/mixxxdj/mixxx/blob/main/.github/workflows/pre-commit.yml) that generates a patch fixing the detected issues as an artifact.

RossBencina:
My understanding is that clang-format is extremely strict in its application of rules and will not give us the flexibility that we need. These tools are really designed for extremely rigid formatting, which is not what we have, nor what we want. Due to the varied contributors our code is way too varied imho, for a strict tool.

I have tried to use both Astyle and clang-format in the past and have given up due to their inflexibility. If you can show that clang-format can check *only* these rules, and not mess up other unrelated stuff then I would be very interested to see a demonstration.

RossBencina:
@Be-ing  I don't like the sound of auto-fixing. The committer should fix the issues. All commits should be clean, and conform to our rules.

RossBencina:
Related #239

RossBencina:
> Also check for a final endline (it's required for some Linux editors).

Done.

Be-ing:
[...]

Issue #239, opened 2015-09-06
Automate checking formatting rules from ImplementationStyleGuidelines


PortAudio-admin:
*Issue created by @RossBencina*

[ImplementationStyleGuidelines](../wiki/ImplementationStyleGuidelines) specifies rules for formating our code.

Our current formatting conventions should be encoded in astyle scripts. We can then generate a diff between "correctly formatted" and our code, and work out how to proceed.

fixdir.bat and fixfile.bat may already be a step in this direction. Need to review.

PR #399, opened 2020-12-27
style: add .astylerc for formatting

philburk:
Also add a badly formatted file for testing astyle.

Test: astyle --project=.astylerc test/style/*.c,h,cpp

### DIFF
--- a/.astylerc
+++ b/.astylerc
@@ -1,0 +1,39 @@
+# Options for astyle, for example:
+#
+#      astyle project=_astylerc test/style/*.c,cpp,h
+#
+# For full docs:
+# http://astyle.sourceforge.net/astyle.html
+#
+# - Preserve file modification date
+--preserve-date
+
+# - Do not create file backups, everything should be VCSed anyway
+--suffix=none
+
+# Align curly brackets
+--style=ansi
+
+# Indent with 4 spaces, convert tabs to spaces
+--indent=spaces=4
+--convert-tabs
+
+# Continuation lines indented 2 levels or aligned
+--indent-continuation=2
+--max-code-length=100
+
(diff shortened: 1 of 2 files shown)


philburk:
See related style issues #393 and #398

dechamps:
Drive-by comment: have you considered [`clang-format`](https://clang.llvm.org/docs/ClangFormat.html)? Not sure how it compares to `astyle` - it might behave better in the face of weird/contrived syntax, I don't know.

philburk:
@dechamps = clang-format looks nice. It is more modern and has more options. 
I installed it on Ubuntu using:
    sudo apt-get install clang-format
Thanks for the tip.

.astylerc
```diff
@@ -0,0 +1,39 @@
...
+# Continuation lines indented 2 levels or aligned
+--indent-continuation=2
+--max-code-length=100
+
+# Put space around operands like + and after commas.
+--pad-oper
+--pad-comma
+
+# Put space inside of parentheses
+--pad-paren-in
```
philburk, line 30: The PortAudio style is to pad the inside of parentheses, like:

    void foo( int x )

But that is an old style and not very popular. What would people think of switching to:

    void foo(int x)

MichalPetryka:
Afaik, Visual Studio and JetBrains products (such as Resharper and CLion) support only clang-format, not astyle.

RossBencina:
Related #239, #398

RossBencina:
Both myself and others have tried to format PortAudio code with astyle and clang-format in the past -- and failed. I don't think either astyle nor clang-format will capture our current formatting guidelines. There are edge-cases where they clash and this will just create a mess of unnecessary and/or unwanted formatting changes.

PR #401, opened 2020-12-27
style: add .clang-format file

philburk:
Also add a badly formatted file for testing clang-format.

Test: clang-format -style=file -i --verbose test/style/*

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,16 @@
+---
+Language:        Cpp
+BasedOnStyle:  WebKit
+
+# These were customized for PortAudio
+AlignEscapedNewlines: Left
+AlignOperands: true
+AllowAllParametersOfDeclarationOnNextLine: false
+BreakBeforeBraces: Allman
+ColumnLimit: 100
+ConstructorInitializerAllOnOneLineOrOnePerLine: true
+ContinuationIndentWidth: 8
+PointerAlignment: Right
+SpacesInParentheses: true
+TabWidth: 4
+...
--- a/test/style/badstyle.c
+++ b/test/style/badstyle.c
@@ -1,0 +1,37 @@
+/**
+ * This is an example of a badly formatted file.
+ * It is used for testing astyle.
+ */
+
+/* bad spaces, bad indentation */
+int foo(float x, int y) {
+  float z = x * y;
+  return z;
+}
+
+/* Very long function declaration. */
+int big_line(int longVariableName, long *variableName, double twiceAsLong,
+             float whenWillItEnd, long anotherOne) {
+  int after_tab = 99;
+  anotherOne = (int)twiceAsLong + longVariableName + *variableName +
+               twiceAsLong + whenWillItEnd + anotherOne + 3.14159265;
+  if (after_tab > longVariableName && *variableName == anotherOne) {
+    return 7;
+  } else
+    return 99;
+}
+
+int use_switch(int choice) {
+  int result = 2;
+  switch (choice) {
+  case 88:
+    return 7;
+  case 99:
+    return 2;
+  case 5: {
+    int x = 5;
+    result = x * choice;
+  } break;
+  }
+  return result;
+}


philburk:
This is an alternative to #399 
I think I like this better. It has much more control.

RossBencina:
My view is that neither AStyle nor ClangFormat are flexible enough to capture our formatting rules. It's a fool's errand.

RossBencina:
Related #239, #398

RossBencina:
The only way I'm prepared to entertain these solutions is if you can show the whole PA code base, reformatted with *no undesired changes* after running through one of these tools.

Please just end the bike-shedding now and forget about astyle and clang-format.

RossBencina:
With respect to PortAudio, I have not changed the position that I stated above. In the interest of having the conversation I will now share my personal opinion.

I believe that formatting consistency is important for readability. This is why we have a style guide.

I am not attached to the particular formatting used by PortAudio, nor am I a great fan, but I am opposed to changing it. That said, I agree that it is not aligned with the most popular modern styles but I also don't think it is unique. We are not the only popular audio library with non-status-quo style.

I too have to make an effort to write code that is formatted in-style, but it doesn't bother me all that much. I have to make a similar effort on almost all code bases I work on -- they usually use different styles.

As a maintainer I view ease of examining the revision history as much more important than a slight increase in difficulty in creating new patches.

I am extremely sensitive about mangling the revision history in ways that obscure the revision history. Such changes makes it difficult to compare arbitrary revisions of the code base (yes, I use a local diff tool that can ignore such formatting, or yes, I could run clang-format on both revisions prior to diffing -- neither of these is as painless as bringing up the diff on the GitHub website). In this sense, I view rewriting the formatting of the entire repository as a catastrophic event. (But yes, similarly catastrophic changes have been made in the past, and yes, it upsets me.)

As far as the "auto-format-everything" approach, obviously it has become a popular approach in recent years. I tried it and I didn't like it. It takes all control away from the developer -- which generally means lower quality, less readable formatting. Personally, for my own code, I have not seen a formatter that allows formatting comments in a way that goes nicely with the code (i.e. they are generally way too rigid in allowing different forms of comments within the code). 

I think that reformatting comments is the showstopping problem with applying these tools to the PA code base (certainly it was when I last tried it). By which I mean that any automated re-format is going to garble the readability of existing comments (and make authoring readable comments more difficult). If there is a tool which does a good job of this I would be interested to see it.

@dechamps I'm not sure what you have in mind, but if the suggestion is to add a .clang-format but have the repo not actually be exhaustively formatted by clang-format using that `.clang-format` I think that's just going to be confusing. On the other hand, as I already said, I'm opposed to performing such an exhaustive rewrite.

dechamps:
> I believe that formatting consistency is important for readability. This is why we have a style guide.

I agree, but I think there's no way PortAudio will get anywhere near consistency any time soon. Yesterday I tried to set up `clang-format` with custom rules like @philburk did in this PR. Looking at the resulting diff revealed tons of problems, not because the formatter itself can't handle the rules, but because there is simply no apparent pattern in the codebase and so it's not even clear what the formatter rules should be. This includes things like spaces in parentheses, `#ifdef` indentation, `*` left-right alignment, `{` placement, etc.

I don't think it's realistic to expect consistency when the current code is not consistent with itself. People won't be able to follow rules if they're confused about what rules to follow.

> I am not attached to the particular formatting used by PortAudio

Me neither. I am fine with any style as long as it's consistent, and hopefully enforceable by an automated formatter. The latter usually follows from the former.

When there are style rules that can be enforced automatically, I stop caring about style because my editor will apply the correct style for me anyway without me having to think about it.

> As a maintainer I view ease of examining the revision history as much more important than a slight increase in difficulty in creating new patches.
> similarly catastrophic changes have been made in the past, and yes, it upsets me

I can relate to that - I also think clean history/diffs is important. Although to be fair, if we end up in a solution where formatting could be enforced by CI, then hopefully we would only have a single "break" in history (that one would have to go around when using e.g. `git blame`) and that's it.

> It takes all control away from the developer -- which generally means lower quality, less readable formatting.

I believe the opposite. Automated formatting enforces consistent formatting, and consistency is what matters most for quality and readability. Manual formatting is error-prone.

> I have not seen a formatter that allows formatting comments in a way that goes nicely with the code
> I think that reformatting comments is the showstopping problem with applying these tools to the PA code base (certainly it was when I last tried it). By which I mean that any automated re-format is going to garble the readability of existing comments (and make authoring readable comments more difficult). If there is a tool which does a good job of this I would be interested to see it.

I believe it is possible to tell `clang-format` to leave comments alone, by disabling options such as [`ReflowComments`](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#reflowcomments). Disabling the [`ColumnLimit`](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#columnlimit) can also help with this. I just gave it a quick try - if I run `clang-format` on, say `pa_win_ds.c` with `ColumnLimit: 0`, it does not touch *any* of the comments at all.

> @dechamps I'm not sure what you have in mind, but if the suggestion is to add a .clang-format but have the repo not actually be exhaustively formatted by clang-format using that `.clang-format` I think that's just going to be confusing.

I don't think it would be any more confusing than the current state of the repo, which is all over the place style-wise. Having that file in there would at least result in some of the future PRs at least being consistent with each other, hopefully gradually migrating the repo to a better, more consistent place over time.

dechamps:
I did some research and came across the [`git-clang-format-per-commit`](https://github.com/marketplace/actions/git-clang-format-check-per-commit) GitHub action, which makes it possible to enforce formatting in CI *without* having to reformat the entire codebase. This is accomplished by only checking the formatting on lines that are changed by a PR. This makes it possible to enforce consistent formatting on *new* code, while leaving existing code alone. I gave it a quick try on one of my projects and it appears to work as expected. (Incidentally, this is also how formatting is enforced at my day job.)

Since there is no requirement that the entire codebase be formatter-compliant at all times, this approach also makes it possible to ignore/override the check on a per-PR basis if the maintainer disagrees with the suggested formatting.

@RossBencina would you be open to this solution? I believe it answers all your concerns:

- It does not require breaking history by suddenly reformatting the entire codebase
- From what I can tell `clang-format` can be configured to leave all comments alone
- If you don't like what `clang-format` is proposing, you can always ignore the check on specific PRs without breaking future checks

At the same time, it will be harder for contributors to accidentally use divergent styles, and editors with `.clang-format` support will spare contributors the burden of formatting their PRs.

(To be clear I still believe that reformatting the entire codebase is worth breaking history - however I understand this is controversial, and I would be fine with the above compromise.)

PR #403, opened 2020-12-28
Remove all alternative sample conversion code using lrintf in pa_converters.c

### LINKED ISSUE #390: pa_converters.c lrintf usage is suspect: does not reproduce the same values as non-lrintf code
When `PA_USE_C99_LRINTF` is defined, `pa_converters.c` uses `lrintf` for truncating floats to integers. For example here is one case:

```
#ifdef PA_USE_C99_LRINTF
        float scaled = *src * 0x7FFFFFFF;
        *dest = lrintf(scaled-0.5f);
#else
        double scaled = *src * 0x7FFFFFFF;
        *dest = (PaInt32) scaled;        
#endif
```
The code in context is here: https://github.com/PortAudio/portaudio/blob/1a5c410f7f09042d2d7ce1509ec3cf0b6091c7f8/src/common/pa_converters.c#L348

I am guessing that the code is intended to implement a fast truncation.

Problem 1: Presumably the code assumes that the rounding mode is set to round to nearest, however this file never checks or sets the rounding mode. Is it set elsewhere? that is not documented. I would expect there to be clear usage of `fesetround` and/or `fegetround`.

Problem 2: The code doesn't work the way the author expected. For example if `scaled == 3.0` then  `lrintf(scaled-0.5f) == 2`. This is because of the way odd/even round to nearest works in IEEE floating point.

The following test program:

```
#include <iostream>
#include <fenv.h>       /* fesetround, FE_* */
#include <math.h>       /* rint */

long int myc99truncate(float x)
{
    return lrintf(x-0.5f);
}

long int mytruncate(float x)
{
    return (long int)x;
}


int main()
{
    fesetround(FE_TONEAREST);
    
    std::cout << 0.0f << "   truncates to " << mytruncate(0.0f) << " or " << myc99truncate(0.0f) << std::endl;
    std::cout << 0.5f << " truncates to " << mytruncate(0.5f) << " or " << myc99truncate(0.5f) << std::endl;
    std::cout << 1.0f << "   truncates to " << mytruncate(1.0f) << " or " << myc99truncate(1.0f) << " *" << std::endl;
    std::cout << 1.5f << " truncates to " << mytruncate(1.5f) << " or " << myc99truncate(1.5f) << std::endl;
    std::cout << 2.0f << "   truncates to " << mytruncate(2.0f) << " or " << myc99truncate(2.0f) << std::endl;
    std::cout << 2.5f << " truncates to " << mytruncate(2.5f) << " or " << myc99truncate(2.5f) << std::endl;
    std::cout << 3.0f << "   truncates to " << mytruncate(3.0f) << " or " << myc99truncate(3.0f) << " *" << std::endl;
    std::cout << 3.5f << " truncates to " << mytruncate(3.5f) << " or " << myc99truncate(3.5f) << std::endl;
    std::cout << 4.0f << "   truncates to " << mytruncate(4.0f) << " or " << myc99truncate(4.0f) << std::endl;
    std::cout << 4.5f << " truncates to " << mytruncate(4.5f) << " or " << myc99truncate(4.5f) << std::endl;
}
```
outputs (non-matching entries marked with *):
```
0   truncates to 0 or 0
0.5 truncates to 0 or 0
1   truncates to 1 or 0 *
1.5 truncates to 1 or 1
2   truncates to 2 or 2
2.5 truncates to 2 or 2
3   truncates to 3 or 2 *
3.5 truncates to 3 or 3
4   truncates to 4 or 4
4.5 truncates to 4 or 4
```

Here is the above test code running in a sandbox:
https://wandbox.org/permlink/ZIZOsQjfIgv3DKAb

If continued use of `lrintf` is desired, the fix that I would recommend is to set the rounding mode to truncate prior to the loop, and reset it to the old rounding mode afterwards. Then simply use `*dest = lrintf(scaled)` in the loop.

RossBencina:
This is code guarded by PA_USE_C99_LRINTF. Fixes #390.

### DIFF
--- a/src/common/pa_converters.c
+++ b/src/common/pa_converters.c
@@ -41,8 +41,6 @@
 
  @brief Conversion function implementations.
 
- If the C9x function lrintf() is available, define PA_USE_C99_LRINTF to use it
-
  @todo Consider whether functions which dither but don't clip should exist,
  V18 automatically enabled clipping whenever dithering was selected. Perhaps
  we should do the same.
@@ -343,13 +341,8 @@ static void Float32_To_Int32(
     while( count-- )
     {
         /* REVIEW */
-#ifdef PA_USE_C99_LRINTF
-        float scaled = *src * 0x7FFFFFFF;
-        *dest = lrintf(scaled-0.5f);
-#else
         double scaled = *src * 0x7FFFFFFF;
         *dest = (PaInt32) scaled;
-#endif
 
         src += sourceStride;
         dest += destinationStride;
@@ -369,17 +362,11 @@ static void Float32_To_Int32_Dither(
     while( count-- )
     {
         /* REVIEW */
-#ifdef PA_USE_C99_LRINTF
-        float dither  = PaUtil_GenerateFloatTriangularDither( ditherGenerator );
-        /* use smaller scaler to prevent overflow when we add the dither */
-        float dithered = ((float)*src * (2147483646.0f)) + dither;
-        *dest = lrintf(dithered - 0.5f);
-#else
         double dither  = PaUtil_GenerateFloatTriangularDither( ditherGenerator );
         /* use smaller scaler to prevent overflow when we add the dither */
         double dithered = ((double)*src * (2147483646.0)) + dither;
         *dest = (PaInt32) dithered;
-#endif
+
         src += sourceStride;
         dest += destinationStride;
     }
@@ -399,15 +386,9 @@ static void Float32_To_Int32_Clip(
     while( count-- )
     {
         /* REVIEW */
-#ifdef PA_USE_C99_LRINTF
-        float scaled = *src * 0x7FFFFFFF;
-        PA_CLIP_( scaled, -2147483648.f, 2147483647.f  );
-        *dest = lrintf(scaled-0.5f);
-#else
         double scaled = *src * 0x7FFFFFFF;
         PA_CLIP_( scaled, -2147483648., 2147483647.  );
         *dest = (PaInt32) scaled;
-#endif
 
         src += sourceStride;
         dest += destinationStride;
@@ -427,19 +408,11 @@ static void Float32_To_Int32_DitherClip(
     while( count-- )
     {
         /* REVIEW */
-#ifdef PA_USE_C99_LRINTF
-        float dither  = PaUtil_GenerateFloatTriangularDither( ditherGenerator );
-        /* use smaller scaler to prevent overflow when we add the dither */
-        float dithered = ((float)*src * (2147483646.0f)) + dither;
-        PA_CLIP_( dithered, -2147483648.f, 2147483647.f  );
-        *dest = lrintf(dithered-0.5f);
-#else
         double dither  = PaUtil_GenerateFloatTriangularDither( ditherGenerator );
         /* use smaller scaler to prevent overflow when we add the dither */
         double dithered = ((double)*src * (2147483646.0)) + dither;
         PA_CLIP_( dithered, -2147483648., 2147483647.  );
         *dest = (PaInt32) dithered;
-#endif
 
         src += sourceStride;
         dest += destinationStride;
@@ -601,13 +574,8 @@ static void Float32_To_Int16(
 
     while( count-- )
     {
-#ifdef PA_USE_C99_LRINTF
-        float tempf = (*src * (32767.0f)) ;
-        *dest = lrintf(tempf-0.5f);
-#else
         short samp = (short) (*src * (32767.0f));
         *dest = samp;
-#endif
 
         src += sourceStride;
         dest += destinationStride;
@@ -631,11 +599,7 @@ static void Float32_To_Int16_Dither(
         /* use smaller scaler to prevent overflow when we add the dither */
         float dithered = (*src * (32766.0f)) + dither;
 
-#ifdef PA_USE_C99_LRINTF
-        *dest = lrintf(dithered-0.5f);
-#else
         *dest = (PaInt16) dithered;
-#endif
 
         src += sourceStride;
         dest += destinationStride;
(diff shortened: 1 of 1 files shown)


RossBencina:
@dmitrykos I am asking for your review just in case you might care about this change. If you don't care, just say so. See #390 for details.

RossBencina:
I posted about this to the mailing list. Let's wait and see whether anyone responds there before merging this one.

dmitrykos:
@RossBencina, sorry I noticed this thread only now. I am in favor to remove lrintf() taking into account your findings in #390 thread. 

The best way would be to use a cast to integer type and modify the rounding mode to - truncate towards zero. It is fast and rounding mode can be set on all platforms (Windows, Apple, Linux).

dmitrykos (approved):

philburk (approved):

Be-ing:
There is a merge conflict now.

RossBencina:
@dmitrykos cast to integer is unaffected by rounding mode. In fact, on x386 target cast to integer is _slow_ because it always resets the rounding mode to truncate, then restores it, on newer architectures it probably uses some SSE that doesn't require setting the rounding mode. if used correctly `lrintf` should be faster than cast to integer -- that is a good reason to use it. But it needs to be done properly, then with benchmarks etc etc.

dmitrykos:
@RossBencina, thank you for correcting me that a cast to integer is guaranteed to be in Round-Towards-Zero mode. Rounding mode though affects `lrint` so the change (removal of `lrint`) made the code portable and unaffected by the rounding mode which could be changed by a user.

I made small experiment and used Visual Studio 2019 (MSVC compiler) targeting Windows 10 (x86, x64).

```c
#include <stdio.h>
#include <math.h>
#include <fenv.h>

int main()
{
	fesetround(FE_UPWARD);

	volatile float vf = 2.2f;

	volatile int vi0 = lrintf(vf);
	volatile int vi1 = (int)vf;

	printf("vi0=%d, vi1=%d", vi0, vi1);

	return 0;
}
``` 

The result is - `vi0 = 3` when using `lrintf` and a cast to the intteger results in `vi1 = 2`, so it confirms that cast works as expected (Round-Towards-Zero mode) and `lrintf` is affected by the rounding mode.

Currently for x86 mode the MSVC compiler enforces the usage of SSE SIMD `cvttss2si` even if SSE is not explicitly specified in the compiler settings when application is targeting Windows 10:

x86:
```asm
volatile int vi1 = (int)vf;
00F61031  movss       xmm0,dword ptr [vf]  
00F61036  cvttss2si   eax,xmm0  
00F6103A  mov         dword ptr [vi0],eax  
```

x64:
```asm
volatile int vi1 = (int)vf;
00007FF7627016B4  movss       xmm0,dword ptr [vf]  
00007FF7627016BA  cvttss2si   eax,xmm0  
00007FF7627016BE  mov         dword ptr [vf],eax
```

In case of `lrintf` compiler is generating a function call and more instructions in x86 mode:

x86
```asm
volatile int vi0 = lrintf(vf);
00591018  movss       xmm0,dword ptr [vf]  
0059101D  movss       dword ptr [esp],xmm0  
00591022  call        dword ptr [__imp__lrintf (0592058h)]  
00591028  mov         dword ptr [ebp-8],eax  
0059102B  add         esp,4  
```

Shorter for x64 but still a function call:

x64:
```asm
volatile int vi0 = lrintf(vf);
00007FF62C8C169D  movss       xmm0,dword ptr [vf]  
00007FF62C8C16A3  call        qword ptr [__imp_lrintf (07FF62C8CA228h)]  
00007FF62C8C16A9  mov         dword ptr [rsp+38h],eax 
```

I also checked what GCC (MinGW.org GCC-6.3.0-1) is doing and the results are similar for a cast to integer when complier is using SSE SIMD. 

x86 + SSE:
```asm
movss	20(%esp), %xmm0	 # vf, vf.2_6
cvttss2si	%xmm0, %eax	 # vf.2_6, vi1.3_7
movl	%eax, 28(%esp)	 # vi1.3_7, vi1
```

For `lrintf` GCC is using single `fistpl` FPU instruction:

```asm
flds	20(%esp)	 # vf
/APP
 # 764 "c:\mingw\include\math.h" 1
fistpl 28(%esp)	 # retval
 # 0 "" 2
/NO_APP
movl	28(%esp), %eax	 # retval, _9
movl	%eax, 24(%esp)	 # _9, vi0
```

If no SSE SIMD requested from the compiler then a cast to integer becomes not optimal because compiler has to set Round-Towards-Zero mode with `fldcw` surrounding `fistpl`, it makes code slower dur to FPU mode switch:

```asm
flds	36(%esp)	 # vf
movb	$12, %ah	 #, tmp94
movw	%ax, 28(%esp)	 # tmp94,
fldcw	28(%esp)	 #
fistpl	24(%esp)	 #
fldcw	30(%esp)	 #
movl	24(%esp), %eax	 #, vi1.3_7
movl	%eax, 44(%esp)	 # vi1.3_7, vi1
```

It is what you mentioned "on x386 target cast to integer is slow because it always resets the rounding mode to truncate, then restores it".

So the conclusion is: 

1. If application is targeting x86 and not requesting SSE from the compiler then cast to integer will be not optimal in terms of performance due to additional usage of `fldcw` instruction surrounding `fistpl`.

2. Usage of `lrintf` when code is compiled with MSVC or GCC and with SSE SIMD enabled is also not optimal because MSVC is doing a function call to `__imp__lrintf` while GCC is falling back to a FPU code that is causing more instructions and also vectorized optimization is no longer possible.

3. Current change to the cast to integer looks optimal for x86/x64 when SSE SIMD is used but less optimal when compiler is using only FPU instructions.

4. `lrint` is affected by the rounding mode and the code should be slower (not benchmarked) on Windows platform if generated by MSVC compiler due to a function call (compiler does not substitude it with a single `fistpl` like GCC does it).

philburk:
There is some interesting code in Android that does lots of PCM format conversions.
https://android.googlesource.com/platform/system/media/+/refs/heads/master/audio_utils/primitives.c

PR #406, opened 2020-12-29
WASAPI: Report thread running so Pa_StartStream can return

bear101:
Report thread running so Pa_StartStream can return prior to PaStreamCallback.

StartStream() calls WaitForSingleObject(stream->hThreadStart, 60*1000)
but it will not be able to return until X number of PaStreamCallback
have been performed by ProcessOutputBuffer().

The above behaviour will cause a deadlock if the user supplied
PaStreamCallback tries to access a shared lock currently held by the
function calling Pa_StartStream().

Also it's a bit weird that you start receiving PaStreamCallbacks
before Pa_StartStream() is able to complete.

### DIFF
--- a/src/hostapi/wasapi/pa_win_wasapi.c
+++ b/src/hostapi/wasapi/pa_win_wasapi.c
@@ -5890,6 +5890,12 @@ PA_THREAD_FUNC ProcThreadEvent(void *param)
         goto thread_error;
     }
 
+    // Signal: stream running
+    stream->running = TRUE;
+
+    // Notify: thread started
+    SetEvent(stream->hThreadStart);
+
     // Initialize event & start INPUT stream
     if (stream->in.clientProc)
     {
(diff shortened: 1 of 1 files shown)


dmitrykos (changes requested):
The same correction is needed for the `ProcThreadEvent`.

dmitrykos:
While checking new changes in both thread functions, to have consistent behavior of Input and Output streams could we move the same code block above `if (stream->in.clientProc)`?, e.g. above lines:

https://github.com/PortAudio/portaudio/blob/3798d94e2a3ebf15b54efe194a186c4a6f1bcafc/src/hostapi/wasapi/pa_win_wasapi.c#L5893-L5894

and:

https://github.com/PortAudio/portaudio/blob/3798d94e2a3ebf15b54efe194a186c4a6f1bcafc/src/hostapi/wasapi/pa_win_wasapi.c#L6174-L6175

I think in this way we won't split the stream start moment of the Input and Output streams. 

If you are ok with this logic would you please make these additional changes.

bear101:
Ok, done! I don't have unit-tests that exercise ProcThreadEvent() but only ProcThreadPoll(), so I hope you're able to test it somehow.

dmitrykos (approved):
Changes look great, thank you very much. I will check if `ProcThreadEvent` is behaving as expected.

dmitrykos:
Tested changes and operation is correct for Event and Poll modes.

PR #419, opened 2020-12-29
Run MSBuild MSVC project CI for pull requests

RossBencina:


philburk (approved):
I am not very familiar with yaml.
But I do not see any problems.

MichalPetryka:
Why not run it on all branches, not just master?

Be-ing:
I think it would be better to remove the MSVC project files: #434

RossBencina:
@MichalPetryka I don't know how to make it run on all branches. Why would we want to run on non-current branches? and how would I configure it to do that?

MichalPetryka:
Making it like this should make it work on all branches, can be useful when working on a branch to see if it compiles correctly without a PR to master.
```
on: [push, pull_request]
```

RossBencina:
@MichalPetryka  I have changed it to the suggested:

`on: [push, pull_request]`

But I am not sure that is different from my previous proposal

```
on:
  push:
    branches: [ master ]
  pull_request:
    branches: [ master ]
```

Which is what the other file c-cpp.yml is using.

PR #422, opened 2021-01-02
Ignore the Visual Studio CMake settings file

dechamps:
When using CMake, this file is written by Visual Studio to hold custom CMake parameters.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,3 +46,6 @@ portaudio-2.0.pc
 
 # Visual Studio CMake build output folder 
 out/
+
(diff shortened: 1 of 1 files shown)


philburk (approved):

Be-ing:
IMO it is inappropriate to add files that are specific to your chosen tooling to a shared project .gitignore. That is what `.git/info/exclude` is for.

RossBencina:
Visual Studio is standard and supported tooling (using CMake with MSVC is not really standard or supported however). In any case I see no problem with this.

PR #428, opened 2021-01-07
wasapi: Device list should not fail with error if no audio device found.

bluecataudio:
When devices count is 0, the API should not fail with an insufficient memory error.

### DIFF
--- a/src/hostapi/wasapi/pa_win_wasapi.c
+++ b/src/hostapi/wasapi/pa_win_wasapi.c
@@ -2110,7 +2110,7 @@ static PaError CreateDeviceList(PaWasapiHostApiRepresentation *paWasapi, PaHostA
 {
     PaUtilHostApiRepresentation *hostApi = (PaUtilHostApiRepresentation *)paWasapi;
     PaError result = paNoError;
-    PaDeviceInfo *deviceInfoArray;
+    PaDeviceInfo *deviceInfoArray = NULL;
     UINT32 i;
     WCHAR *defaultRenderId = NULL;
     WCHAR *defaultCaptureId = NULL;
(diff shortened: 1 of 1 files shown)


dmitrykos (changes requested):
Thank you for the fix proposal. It is incomplete though and will cause a crash on line https://github.com/PortAudio/portaudio/blob/7073b1f2ce240266d9647aa3a18c647348dc7eca/src/hostapi/wasapi/pa_win_wasapi.c#L2269 so we need to implement it better.

1. Would you please adjust code style of the proposed change to match overall style of the WASAPI implementation as such:
```c
if ((paWasapi->deviceCount != 0) && ((deviceInfoArray = AllocateDeviceListMemory(paWasapi)) == NULL))
```

2.  To fix a crash when `paWasapi->deviceCount` is 0 please change line https://github.com/PortAudio/portaudio/blob/7073b1f2ce240266d9647aa3a18c647348dc7eca/src/hostapi/wasapi/pa_win_wasapi.c#L2265 to such:

```c
if ((hostApi->info.deviceCount != 0) && (hostApi->info.deviceCount < PA_WASAPI_MAX_CONST_DEVICE_COUNT))
```

3. To make code more reliable for a future changes let's initialize `deviceInfoArray` on this line: https://github.com/PortAudio/portaudio/blob/7073b1f2ce240266d9647aa3a18c647348dc7eca/src/hostapi/wasapi/pa_win_wasapi.c#L2113

During a commit please amend your last commit, so that we will have just one commit for this PR.

bluecataudio:
It looks like the mix of github desktop app and local git completely screwed-up the branch. I have a fixed file here with all the appropriate changes. Do you mind if I create another pull request with a single commit instead? I apologize for the inconvenience.

dechamps:
@bluecataudio There's no need to create another pull request. Simply delete your old, broken branch, rename your cleaned up branch to take its place, and force-push it to GitHub to overwrite history.

bluecataudio:
Ok, will do. Thanks.

bluecataudio:
The branch has been recreated with a new commit, but it looks like the pull request cannot be re-opened, probably because github desktop delete the remote branch. Is there away to delete the request so that a brand new clean one can be created instead?

dechamps:
I'm not sure if it's possible to recover from a deleted remote branch. You can try re-creating one with the same name, if that doesn't work, you'll have to open another pull request I'm afraid.

bluecataudio:
Ok thanks. Will create another pull request. Sorry for the mess!

Issue #431, opened 2021-01-09
Hifiberry DAC+ADC pro is not listed always


hoene:
**Describe the bug**
Hifiberry soundcard disappears from the list of soundcards if listed with portaudio sometimes.

<pre>
// check all soundcards
  int amountOfAudioDevices = Pa_GetDeviceCount();
  for (PaDeviceIndex i = 0; i < amountOfAudioDevices; i++) {
    const PaDeviceInfo *info = Pa_GetDeviceInfo(i);
    printf("%d \"%s\" \"%s\"", i, info->name, getUSBID(info->name));
 }
</pre>

However, if I use aplay -L all soundcards are available. 

If portaudio fails to list all, then I see an error message:
<pre>Expression 'SetApproximateSampleRate( pcm, hwParams, defaultSr )' failed in 'src/hostapi/alsa/pa_linux_alsa.c', line: 931</pre>

**To Reproduce**
Steps to reproduce the behavior. Include code if applicable.

1. Call
<pre>aplay -Dplughw:CARD=sndrpihifiberry,DEV=0 samplingrate44100channel1.wav</pre>

2. List soundcards as above -> output is 
<pre>
0 "bcm2835 HDMI 1: - (hw:0,0)" ""
1 "bcm2835 Headphones: - (hw:1,0)" ""
2 "snd_rpi_hifiberry_dacplusadcpro: HiFiBerry DAC+ADC Pro HiFi multicodec-0 (hw:2,0)" ""
3 "USB Audio CODEC: - (hw:3,0)" "08bb:2902"
...
</pre>

3. Call
<pre>aplay -Dplughw:CARD=sndrpihifiberry,DEV=0 samplingrate48000channel1.wav</pre>

4. List soundcards as above -> output is 
<pre>
0 "bcm2835 HDMI 1: - (hw:0,0)" ""
1 "bcm2835 Headphones: - (hw:1,0)" ""
2 "USB Audio CODEC: - (hw:3,0)" "08bb:2902"
...
</pre>

5. Call
<pre>aplay -Dplughw:CARD=sndrpihifiberry,DEV=0 samplingrate44100channel1.wav</pre>

6. List soundcards as above -> output is 
<pre>
0 "bcm2835 HDMI 1: - (hw:0,0)" ""
1 "bcm2835 Headphones: - (hw:1,0)" ""
[...]

PR #432, opened 2021-01-09
More relaxed behavior with sample rate mismatch during groping

hoene:
solved https://github.com/PortAudio/portaudio/issues/431


philburk:
I will try to verify this on a Raspberry Pi and on an Ubuntu VM. I have a HiFiBerry DAC.

RossBencina:
@hoene Could you explain what this change does functionally please? I don't understand how it has a different effect from the old code.

hoene:
@RossBencina 
The Hifiberry Pro supports two clock sources, one for 44100 and one for 48000.

Somehow, portaudio, the kernel, or somebody else remembers, which operational mode was selected last time I have run aplay.

Then, when I switch the sampling rate with aplay, portaudio first tests, which sampling rate would be supported. However, during this "groping" the clock source of the Hifiberry is not correctly switched (indeed, during this probing the Hifiberry device driver is not even called). 

My change does not fix this problem. However, it relaxes the grobing function so that even slightly wrong sampling rates are accepted (e.g. 44100 vs. 48000).

philburk (approved):
I tried to reproduce the bug on an RPi4 using a HiFiBerry Dac+.
I could not reproduce the bug. Probably because it is not the same PRO board.

But I applied this patch and nothing seemed to break.

philburk:
@RossBencina - Are you OK if I merge this?

RossBencina:
> My change does not fix this problem. However, it relaxes the grobing function so that even slightly wrong sampling rates are accepted (e.g. 44100 vs. 48000).

I would not call 44100 vs. 48000 "slightly wrong" these are definitely distinct sample rates and should not be automatically interchanged/substituted.

RossBencina:
> Merge remote-tracking branch 'origin/master' into fix-hifiberry-dacan

as a rule, please don't include merge nodes in pull requests. if necessary, rebase your branch onto current master

Be-ing:
> as a rule, please don't include merge nodes in pull requests. if necessary, rebase your branch onto current master

It would help to document this in a CONTRIBUTING.md file in the root of the repository. Every project has different preferences about this.

MichalPetryka:
> > Merge remote-tracking branch 'origin/master' into fix-hifiberry-dacan
> 
> as a rule, please don't include merge nodes in pull requests. if necessary, rebase your branch onto current master

With small PRs like this one using the `Squash and merge` option deals with them, but with bigger ones  rebasing is better if you want to keep them in multiple commits.

trentp-igor:
> I would not call 44100 vs. 48000 "slightly wrong" these are definitely distinct sample rates and should not be automatically interchanged/substituted.

Indeed, it would be completely wrong to silently substitute 48 kHz when 44.1 KHz is requested.

However, I believe the way this code is supposed to work is that it only requests a sample rate approximately near 44.1k.  The hardware is supposed to set the closest rate it can do, which could be off by any amount.  One must query actual rate afterward to know 44.1 vs 48 vs anything else, and PA does do this correctly.  It should work as "set rate near 44.1", "return rate 48", "default rate is 48 kHz".

It looks like a bug in this hardware driver that it fails rather than return the closest rate.  It may be easier to work around this by hacking portaudio instead of fixing driver, but this leaves driver broken for anything which is not portaudio.

philburk:
I am concerned about coercing paInvalidSampleRate to paNoError. It might mask problems on other devices.
Maybe SetApproximateSampleRate() should be modified.
Or we should just try to get them to fix the driver.
I am reluctant to risk a regression in PortAudio just for a workaround for a buggy driver.
Maybe we could check the device ID and make the workaround conditional.

RossBencina:
I don't think this particular change is an acceptable workaround, given that it effects a wide reaching change to behavior. Possibly some more specific fix would be acceptable. (e.g. changing `SetApproximateSampleRate` to accept buggy sample rate changes only on hifiberry hardware).

I'm closing this now, please free to propose an alternate fix.

PR #435, opened 2021-01-13
CMake: use standard BUILD_SHARED_LIBS variable

Be-ing:
Currently CMake builds both shared and static libs. This is bizarre and adds complexity to the CMakeLists.txt code. CMake has a standard variable BUILD_SHARED_LIBS to control this.


Be-ing:
Superseded by #443

Issue #82, opened 2008-02-23
Port PABLIO to V19 blocking i/o API


PortAudio-admin:
*Issue created by @RossBencina*

pablio is not supported in V19. We have mostly-working blocking i/o implementations for all OSes (not quite all APIs yet).

philburk:
This Issue is very old. We should remove PABLIO and just make sure all the blocking I/O APIs are working.

PR #437, opened 2021-01-13
remove deprecated pablio

Be-ing:
This is deprecated code from 1999. Is there any reason to keep this clutter?


philburk (approved):
>  Is there any reason to keep this clutter?

The word "clutter" is a bit harsh. PABLIO was very useful at one point. Hmmm, I wonder if there is such a thing as a "code hoarder"?

PABLIO was deprecated in 2012. I guess it is time to let go.

Be-ing:
> I wonder if there is such a thing as a "code hoarder"?

Don't worry, it's still in the Git history forever. :)

philburk:
Phil will test compile PABLIO. If it does not compile then we can assume it is not being used.
Then we should test the blocking read/write calls in PortAudio to see if they work on all host APIs.
If they all work then we can safely remove PABLIO.

Be-ing:
That seems like a good plan to me.

MichalPetryka:
> Phil will test compile PABLIO. If it does not compile then we can assume it is not being used.
> Then we should test the blocking read/write calls in PortAudio to see if they work on all host APIs.
> If they all work then we can safely remove PABLIO.

I've just tested on Windows, MME, DirectSound and WASAPI seem to work with blocking, WDMKS errors out on init, prob here (for some reason the hostapi error message was empty so I'm not sure): https://github.com/PortAudio/portaudio/blob/master/src/hostapi/wdmks/pa_win_wdmks.c#L4459

philburk:
Thanks Michal. That's very helpful information.

MichalPetryka:
The error code in errorCode is 140732712890608 but the errorText seems to be a null pointer, which is kindof a bad behaviour since the comment in the struct says `a textual description of the error if available, otherwise a zero-length string` and a null is neither of those.

MichalPetryka:
Also tested with ASIO using ASIO4ALL, blocking interface seems really unstable there, `Pa_GetStreamReadAvailable` seems to always return 0 and `Pa_ReadStream` requires a few tries before returning `paNoError` instead of `paTimedOut`.

Be-ing:
Thanks for testing. Maybe we could fix those for 19.8 and remove pablio then?

philburk (approved):
I tried to compile PABLIO. It was never converted to the V19 API. See #82

It uses the old API and so it seems very unlikely anyone that is using it.

Be-ing:
ping @RossBencina

RossBencina (approved):

PR #440, opened 2021-01-13
move Doxyfile & Doxyfile.developer to doc folder

Be-ing:


philburk:
When this is merged, I will also need to update the scripts that generate the autodocs on the server. And, yes, those should be in the repo.

philburk (approved):
I am generally in favor of moving files out of the top folder because it is cluttered.
Alternatively, these files could go in a "scripts" folder.

RossBencina:
I don't get why you would move these, they control documentation generation for the whole source tree, not just the doc folder.

Be-ing:
Oh, I didn't realize that. I'll close this then.

PR #442, opened 2021-01-13
CMake: remove PA_DISABLE_INSTALL hack

Be-ing:
There is no need for this. If the user does not want to install,
then they only need to build the portaudio target.


Be-ing:
Superseded by #443

PR #444, opened 2021-01-13
remove ASIHPI host API

Be-ing:
This was a host API for one specific vendor's audio interfaces. It
has been obsolete since the asihpi ALSA driver was merged into
Linux 2.6.35 in 2010.


philburk (commented):
According to https://audioscience.com/internet/download/sdk/sdk.htm
they recommend ALSA for Linux. So it does seem obsolete. But someone may be using it.

We should send a notice to the PortAudio mailing list that it is being deprecated.
Maybe that has already been done. If no one objects then I am OK with removing it.

philburk:
I sent an email to PortAudio list asking if anyone objects.

Be-ing:
Considering the lack of objections, I think we should go ahead and merge this. But there's no rush to remove this. Let's wait until after a new release with #356.

philburk (approved):

RossBencina (approved):

RossBencina (commented):
There is no evidence that HPI is "obsolete". There are recent (2019) HPI drivers at the link below, and no indication that the HPI is deprecated on Linux.

https://audioscience.com/internet/download/sdk/sdk.htm

philburk:
Upon reviewing the website again, it seems that Audio Science is supporting their proprietary API and that is is sometimes more up-to-date than the ALSA or other APIs.

We will close this for now until until the API is abandoned by Audio Science.

PR #446, opened 2021-01-16
docs: Pa_OpenStream() fails if sampleRate unavailable

### LINKED ISSUE #433: Clarify response to open() using unavailable sample rates
The documentation is not clear about what happens if a stream cannot be opened at the desired sample rate.

http://files.portaudio.com/docs/v19-doxydocs/portaudio_8h.html#a443ad16338191af364e3be988014cbbe

The docs imply it may be necessary to query for stream info to get the actual sample rate.

http://files.portaudio.com/docs/v19-doxydocs/structPaStreamInfo.html

Ross wrote:
> I would expect [paInvalidSampleRate] to be returned if the device can not be
> opened with the specified sample rate.

philburk:
It was not clear in the previous docs.
Fixes #433


Be-ing:
Documenting the current behavior is an improvement, but shouldn't this return an error if the sample rate is not available?

RossBencina (approved):

philburk:
I will merge this after Ross' whitespace changes go in.

philburk:
@Be-ing Yes, I am documenting that the open should fail if the sample rate is unavailable.

PR #448, opened 2021-01-19
OSS: include both /dev/dsp and /dev/dsp0

emaste:
On FreeBSD /dev/dsp is the default device but may not be /dev/dsp0.
PortAudio in the FreeBSD ports collection initially had a patch to
replace the default device with /dev/dsp (for example, with three
devices and /dev/dsp1 as the default we'd have /dev/dsp0, /dev/dsp,
/dev/dsp2).  This caused some trouble (the first device is no longer
default) and was later replaced with this change, leaving /dev/dsp
as the first entry, and just adding each /dev/dsp%d that exists.

FreeBSD bugs 208968 and 236118.

### DIFF
--- a/src/hostapi/oss/pa_unix_oss.c
+++ b/src/hostapi/oss/pa_unix_oss.c
@@ -535,13 +535,13 @@ static PaError BuildDeviceList( PaOSSHostApiRepresentation *ossApi )
      * add it to a linked list.
      * A: Set an arbitrary of 100 devices, should probably be a smarter way. */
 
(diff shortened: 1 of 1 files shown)


emaste:
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=208968
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=236118

philburk (approved):

RossBencina (approved):

PR #450, opened 2021-01-20
OSS: remove old runtime AFMT_S16_NE endianness compat

emaste:
We had a runtime endianness test for OSS systems which do not define
AFMT_S16_NE.  It is unclear if any operating system needs this today,
but if one does this should be implemented as a compile-time #define
instead.

Patches are forthcoming for AFMT_S32_NE for FreeBSD and there is no
value in extending this runtime endianness test.

### DIFF
--- a/src/hostapi/oss/pa_unix_oss.c
+++ b/src/hostapi/oss/pa_unix_oss.c
@@ -927,6 +927,11 @@ static PaError Pa2OssFormat( PaSampleFormat paFormat, int *ossFormat )
         case paInt16:
             *ossFormat = AFMT_S16_NE;
             break;
+#ifdef AFMT_S32_NE
+        case paInt32:
+            *ossFormat = AFMT_S32_NE;
+            break;
+#endif
         default:
             return paInternalError;     /* This shouldn't happen */
     }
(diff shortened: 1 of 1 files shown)


RossBencina:
If there is an improvement on the way, I think the removal of the old code and addition of the new code should be made as a single atomic PR.

emaste:
It turns out that not all systems define `AFMT_S32_NE` (in particular, the GitHub CI Linux environment), so wrap the new cases in `#ifdef AFMT_S32_NE`.

The change does not depend on the first commit in this PR now (run-time endianness check) https://github.com/PortAudio/portaudio/commit/b40acc9a75b0e8266928518c31e19c9b7b306e60 and I can drop it if desired, although I still think it's a reasonable change to make

philburk:
The AFMT_S32_NE changes are fine but unrelated to the original PR.

The original objection was that the first commit removed code for AFMT_S16_NE that sort of worked and replaced it with a #error. If we are 100% certain that AFMT_S16_NE is always defined then this is OK. But the commit message says "It is unclear if any operating system needs this today".

So the only possible impact from that change was that it might break some builds.  The comment says the code should be replaced with a #define. If we are going to remove the old code we should do that instead of the #error. Here is a way to detect Endianness in the preprocessor.

http://esr.ibiblio.org/?p=5095

But that is a bit tricky.

I recommend submitting the AFMT_S32_NE changes as a separate PR and leaving AFMT_S16_NE as is.

philburk:
#462 has the 32 bit support. So closing this one based on previous discussion.

PR #451, opened 2021-01-21
build: Add missing src directory.

orbea:
This fixes out of tree builds when using slibtool (https://dev.midipix.org/cross/slibtool).
```
rdlibtool: error logged in slbt_exec_compile(), line 158: path not found: src/hostapi/skeleton/.libs/.
make: *** [Makefile:227: src/hostapi/skeleton/pa_hostapi_skeleton.lo] Error 2
```
Full log:  [build.log](https://github.com/PortAudio/portaudio/files/5846439/build.log)

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -146,6 +146,7 @@ SRC_DIRS = \
 	src/hostapi/dsound \
 	src/hostapi/jack \
 	src/hostapi/oss \
+	src/hostapi/skeleton \
 	src/hostapi/wasapi \
 	src/hostapi/wdmks \
 	src/hostapi/wmme \


RossBencina (approved):

RossBencina:
Commit message should be changed to something like "autoconf build: add src dir src/hostapi/skeleton to fix out of tree builds."

orbea:
> Commit message should be changed to something like "autoconf build: add src dir src/hostapi/skeleton to fix out of tree builds."

Thanks for the suggestion, I updated the commit message as you suggested.

philburk (approved):

PR #452, opened 2021-01-21
build: Fix builds with parallel make.

orbea:
When building portaudio with `-j2` or greater, slibtool (https://dev.midipix.org/cross/slibtool) and `ccache(1)` the build fails because `mkdir(1)` is slower than slibtool. This does not seem to happen with GNU libtool which is slower than `mkdir`.
```
rdlibtool: error logged in slbt_exec_compile(), line 158: path not found: src/common/.libs/.
make: *** [Makefile:228: src/common/pa_allocation.lo] Error 2
make: *** Waiting for unfinished jobs....
touch lib-stamp
```
Full log: [build.log](https://github.com/PortAudio/portaudio/files/5847183/build.log)

I fixed this by applying the `lib-stamp` as a dependency wherever it might be needed to be sure this doesn't happen again.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -220,16 +220,16 @@ clean:
 distclean: clean
 	$(RM) config.log config.status Makefile libtool portaudio-2.0.pc
 
-%.o: %.c $(MAKEFILE) $(PAINC)
+%.o: %.c $(MAKEFILE) $(PAINC) lib-stamp
 	$(CC) -c $(CFLAGS) $< -o $@
 
-%.lo: %.c $(MAKEFILE) $(PAINC)
+%.lo: %.c $(MAKEFILE) $(PAINC) lib-stamp
 	$(LIBTOOL) --mode=compile $(CC) -c $(CFLAGS) $< -o $@
 
-%.lo: %.cpp $(MAKEFILE) $(PAINC)
+%.lo: %.cpp $(MAKEFILE) $(PAINC) lib-stamp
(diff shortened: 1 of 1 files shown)


RossBencina:
what is `lib-stamp`  ? I can not find any info about it with google.

Redfoxymoon:
https://github.com/PortAudio/portaudio/blob/master/Makefile.in#L242

orbea:
> what is `lib-stamp`? I can not find any info about it with google.

Its a make rule specific to portaudio that creates the required directories and a sentinal file which other make rules can depend on. This is needed since make can't depend on directories directly.

Reference: https://stackoverflow.com/questions/4440500/depending-on-directories-in-make/4614143#4614143

philburk (approved):

RossBencina (approved):

RossBencina (approved):
I'm not qualified to review this, but I'm satisfied that you guys seem to know what you're talking about.

philburk:
I will run autoconf to regenerate the "configure" file to see if it needs to be update in the repo.

philburk:
I cannot reproduce the failure with "make -j2" on Ubuntu. But I believe it could be a problem. Race conditions are intermittent.

I ran autoconf with and without this patch. The Makefile changes but the "configure" file did not.
So it seems configure is using Makefile.in and that autoconf does not.

I think this patch is complete and does not to include a new "configure" file.

PR #454, opened 2021-01-22
Whitespace cleanup in src

MichalPetryka:
Automatically cleaned up whitespaces in src in preparation for #394:
Tabs were converted to spaces using `find ./ -iname '*.c' -type f -exec bash -c 'expand -t 4 "$0" | sponge "$0"' {} \;` for .c, ,h and .cpp.
Lines endings were trimmed and doubled empty lines at file ends removed with my private reformat utility.

PR #458, opened 2021-01-22
Whitespace cleanup in bindings

MichalPetryka:
Same as #454, this one has a lot of changes mostly due to tabs being converted to spaces.

### DIFF
--- a/bindings/cpp/include/portaudiocpp/SampleDataFormat.hxx
+++ b/bindings/cpp/include/portaudiocpp/SampleDataFormat.hxx
@@ -11,21 +11,21 @@ namespace portaudio
 {
 
 
-	//////
-	/// @brief PortAudio sample data formats.
-	///
-	/// Small helper enum to wrap the PortAudio defines.
-	//////
-	enum SampleDataFormat
-	{
-		INVALID_FORMAT	= 0,
-		FLOAT32			= paFloat32,
-		INT32			= paInt32,
-		INT24			= paInt24,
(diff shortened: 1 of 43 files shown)


bindings/cpp/include/portaudiocpp/SampleDataFormat.hxx
```diff
@@ -11,21 +11,21 @@ namespace portaudio
...
-	};
+    //////
+    /// @brief PortAudio sample data formats.
+    ///
+    /// Small helper enum to wrap the PortAudio defines.
+    //////
+    enum SampleDataFormat
+    {
+        INVALID_FORMAT  = 0,
+        FLOAT32         = paFloat32,
```
philburk, line 22: The expand tool seems to handle tabs in the middle of lines very well.

philburk (approved):

bindings/cpp/include/portaudiocpp/SampleDataFormat.hxx
MichalPetryka: Yeah, it has an option to only convert tabs at the start of lines, but after seeing things like this nicely converted I used it without that.

philburk:
Note that we are planning to remove the Java and C++ bindings from PortAudio.

MichalPetryka:
If the plan is to move them to a separate repository, cleaning them up before doing it shouldn't be an issue.

RossBencina (approved):

PR #459, opened 2021-01-22
CMake: build tests and examples

Be-ing:
First step towards removing redundant build systems


test/patest_converters.c
```diff
@@ -215,7 +215,7 @@ float MaximumAbsDifference( float* sourceBuffer, float* referenceBuffer, int cou
     return result;
 }
 
-int main( const char **argv, int argc )
+int main( int argc, const char **argv )
```
philburk, line 218: Wow! That must have been around a while.

philburk (approved):
Thanks for making these improvements to the PortAudio CMake build.

I see that you require CMake 3.0. That was released in June 2014. According to Wikipedia, version 3.0 is the beginning of "Modern CMake".

https://en.wikipedia.org/wiki/CMake

philburk:
@Be-ing - Are you OK with us squashing these commits when we merge?

Be-ing:
You can do whatever you want with the Git history. IMO commits should be as small as possible while still building. This is how I normally work and it makes reviewing much easier.

dechamps:
@philburk The branch history looks clean, so it seems best to rebase, not squash.

Be-ing:
Merge for 19.7? It doesn't matter much to me either way.

Be-ing:
Note that this removed the autotools builds from GitHub Actions CI. If this is not desired or controversial for now, let's hold off on this for 19.7.

Be-ing:
Alternatively I can move the commit that changes the GitHub Actions workflow file to #461.

test/patest_converters.c
philburk: This is now part of #541

RossBencina:
I notice that the 3 changes that I flagged have not yet been moved out of this change set. The changes are quite reasonable but should be submitted separately and merged before this fix.

RossBencina:
> Note that this removed the autotools builds from GitHub Actions CI. If this is not desired

It is not desired. Please don't do that as part of this PR

Be-ing:
> I notice that the 3 changes that I flagged have not yet been moved out of this change set. The changes are quite reasonable but should be submitted separately and merged before this fix.

Why? What difference does it make? They're already cleanly separated into their own commits.

philburk:
> Why? What difference does it make? They're already cleanly separated into their own commits.

There are several reasons.

1. It is good practice to have unrelated changes in separate pull requests so they can be tracked more easily.
2. We generally squash pull requests into one commit instead of merging multiple commits.
3. The reversed argc/argv in patest_converters.c has been fixed in #544.

test/patest_converters.c
philburk: Fixed in #544

test/patest_unplug.c
```diff
@@ -41,6 +41,7 @@
  * license above.
  */
 
+#include <inttypes.h>
```
philburk, line 44: This should be fixed in a separate PR.

qa/paqa_devs.c
```diff
@@ -55,6 +55,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#define _USE_MATH_DEFINES
```
philburk, line 58: Needs to be fixed in a separate PR.

philburk: See #551

philburk:
@Be-ing - I would love to get these cmake changes merged.

The three changes that are not cmake related for patest_converters.c, patest_unplug.c and paqa_devs.c have now been fixed in separate PRs. So those commits can be merged. If I squash and merge then they will go away. I can handle that.

If you, Be-ing, can restore the testing of ./configure to the CI test then we can merge this.

Be-ing:
I rebased and moved the CMake CI to a separate file.

Be-ing:
> We generally squash pull requests into one commit instead of merging multiple commits.

This is only helpful if the Git history is a mess.

philburk:
@Be-ing - Thanks for addressing our comments. This looks great.

Regarding Squashing, often the commit history is a long chain of fixes and tweaks. So Squashing is needed to clean up. But the commits for this PR looks clean and modular. So we will probably just Rebase and Merge. Thanks.

.github/workflows/cmake.yml
```diff
@@ -0,0 +1,56 @@
+name: C/C++ CI
```
RossBencina, line 1: This should probably say `C/C++ CI (CMake)` rather than using the same name in two different workflow scripts.

RossBencina (approved):

qa/paqa_devs.c
```diff
@@ -55,6 +55,7 @@
...
 #include <math.h>
```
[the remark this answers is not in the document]
Be-ing: I don't understand how you're commenting on outdated code?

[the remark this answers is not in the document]
philburk: These were old comments that got stuck in GitHub.

PR #460, opened 2021-01-22
CMake: use GNUInstallDirs instead of hardcoding installation dirs

Be-ing:
CMake automatically installs to the correct paths for each platform if you don't hardcode the wrong path.


Be-ing:
Superseded by #461

PR #462, opened 2021-01-23
OSS: support 32-bit signed format AFMT_S32_NE

emaste:
It may not be provided by all OSS implementations, so wrap the cases in
`#ifdef AFMT_S32_NE`.

### DIFF
--- a/src/hostapi/oss/pa_unix_oss.c
+++ b/src/hostapi/oss/pa_unix_oss.c
@@ -937,6 +937,11 @@ static PaError Pa2OssFormat( PaSampleFormat paFormat, int *ossFormat )
         case paInt16:
             *ossFormat = AFMT_S16_NE;
             break;
+#ifdef AFMT_S32_NE
+        case paInt32:
+            *ossFormat = AFMT_S32_NE;
+            break;
+#endif
         default:
             return paInternalError;     /* This shouldn't happen */
     }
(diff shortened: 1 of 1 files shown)


philburk (approved):

src/hostapi/oss/pa_unix_oss.c
```diff
@@ -937,6 +937,11 @@ static PaError Pa2OssFormat( PaSampleFormat paFormat, int *ossFormat )
         case paInt16:
             *ossFormat = AFMT_S16_NE;
             break;
+#ifdef AFMT_S32_NE
+        case paInt32:
+            *ossFormat = AFMT_S32_NE;
+            break;
+#endif
         default:
             return paInternalError;     /* This shouldn't happen */
```
RossBencina, line 946: From what I understand `/* This shouldn't happen */` is no longer a true statement (in fact it is already false for example calling with `paFloat32`).  Perhaps there should be an `#elseif` to return `paSampleFormatNotSupported`? I'm not sure how this affects the calling code.

emaste: It should come from
```
        PA_ENSURE( GetAvailableFormats( component, &availableFormats ) );
        hostFormat = PaUtil_SelectClosestAvailableFormat( availableFormats, component->userFormat );
```

so I think it is arguably correct as is; GetAvailableFormats populates only `paUInt8`, `paInt8`, `paInt16`, and `paInt32`, returning `paSampleFormatNotSupported` if none of them are available.

It should not be possible to call `Pa2OssFormat` in OSS with e.g. `paFloat32`?

RossBencina (approved):

PR #464, opened 2021-01-24
Fix user-facing and non-user-facing typos 

luzpaz:
Found via `codespell v2.1.dev0`  
`codespell -q 3 -L chello,gord,guid,numer,pres,uint,wirth`


philburk (approved):
Thanks. We recently went through a typo purge. These slipped through.

RossBencina (approved):
All changes are fine. Only thing is that "Wait awhile" is already grammatically correct, it is an idiom of USA English.

luzpaz:
@RossBencina should I revert that?

PR #469, opened 2021-01-24
Add macosx11.1 support to configure

### LINKED ISSUE #378: Change MacOS build mmacosx-version-min to 10.5
PR #356 removes deprecated support for 10.4 and adds support for the newer CoreAudio APIs.
After it merges, we should update configure.in and change

    -mmacosx-version-min=10.4
to

    -mmacosx-version-min=10.5

### LINKED ISSUE #468: Mac: runtime crashes on Big Sur if built using SDK 11
Here is an example error from #395
Error: ||PaMacCore (AUHAL)|| Error on line 1277: err='-66748', msg=Unknown Error

This crash was conflated with Issue #218, which involved deprecated audio APIs. 
But this crash is happening even after the deprecated audio calls have been replaced in PR #356.

jmelas wrote in [#218 comment](https://github.com/PortAudio/portaudio/issues/218#issuecomment-731310548)
> portaudio fails in runtime on macOS 11 only if the program is built against macOS SDK 11.
> Latest XCode 12.2 comes with macOS SDK 11.
> I downgraded Xcode to version 12.1 which uses 10.15 SDK and then portaudio worked fine on Big Sur!

Related Issues:
#395 - Can't open stream on macOS 11.x 
https://github.com/gordonklaus/portaudio/issues/41
https://github.com/tlecomte/friture/issues/154

philburk:
This is needed to support the new XCode 12.3.
I also went ahead and added macosx11.2 because it is surely coming.

Also changed macosx-version-min to 10.6

Fixes #378
Fixes #468

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@ lib-stamp
 lib/*
 libtool
 portaudio-2.0.pc
+autom4te.cache/*
 
 # Precompiled Headers
 *.gch
--- a/configure
+++ b/configure
@@ -723,7 +723,6 @@ infodir
 docdir
 oldincludedir
 includedir
-runstatedir
 localstatedir
 sharedstatedir
 sysconfdir
@@ -822,7 +821,6 @@ datadir='${datarootdir}'
 sysconfdir='${prefix}/etc'
 sharedstatedir='${prefix}/com'
 localstatedir='${prefix}/var'
-runstatedir='${localstatedir}/run'
 includedir='${prefix}/include'
 oldincludedir='/usr/include'
 docdir='${datarootdir}/doc/${PACKAGE}'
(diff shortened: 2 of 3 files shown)


Be-ing:
Why is there an upper limit on the supported SDK version?

philburk:
> Why is there an upper limit on the supported SDK version?

@Be-ing - What limit are you referring to? Are you asking why I stopped at "11.2"?

My answer would be that the elif chain is cumbersome. In the past we only supported released SDKs. I added 11.2 in anticipation of the next release. How many is enough? I don't know.

I also added a TODO that suggested using -showsdks to avoid the chain of elifs and also work automatically with some future releases. But that requires a bit more study. This was a quick fix for a broken build.

Be-ing:
You can use `xcodebuild -version -sdk macosx SDKVersion` to get the installed version. Just check that is >= 10.6 and this PortAudio release will remain buildable on future versions of macOS.

philburk:
Thanks! I need an emoji for head exploding.
I will experiment with that for an alternative PR.
I may go ahead and just merge this as an emergency fix.

Be-ing:
https://emojipedia.org/exploding-head/

philburk:
I am closing this in favor of #470.  It is much simpler and will last longer without needing changes.
By the time it gets to the new code it should be SDK 1.6 or later.

PR #473, opened 2021-01-27
ASIO: Handle buffer size changes

### LINKED ISSUE #472: Handling ASIO buffer size changes
I've been working on adding portaudio support to Jamulus (corrados/jamulus#870).

One thing I found was that portaudio's API doesn't provide a way to handle changes in ASIO buffer size. The result is that when the user changes the buffer size in the ASIO driver control panel, the sound all of a sudden gets an annoying buzzing distortion. Jamulus'  current ASIO code handles this correctly.

I will make a pull request from the patch I used to fix this problem shortly.

[A user on the mailing list pointed out a similar issue affects WASAPI device sample rate](https://listserv.cuit.columbia.edu/scripts/wa.exe?A2=PORTAUDIO;ca8d968a.2101D&S=); that doesn't affect my use-case since Jamulus only supports 48kHz and so there won't be any use in reacting to sample rate changes.

npostavs:
This fixes #472. I don't know if it's the best approach.

```
When kAsioBufferSizeChange or kAsioResetRequest is received, remember
it and update the buffer sizes the next time
PaAsio_GetAvailableBufferSizes() is called.  Add callbacks for those
messages so that the library user can be aware of them.
```

### DIFF
--- a/include/pa_asio.h
+++ b/include/pa_asio.h
@@ -80,8 +80,9 @@ typedef void (*PaAsioResetRequestCallback)();
 
  @param callback The function to call on ASIO reset request.  This is called
  when the driver's buffer size has changed or it needs a reset, usually in
- response to a user changing the configuration.  Pass NULL to unregister a
- callback.
+ response to a user changing the configuration.  Call Pa_Terminate followed by
+ Pa_Initialize to have the changes reflected in Portaudio's API.  Pass NULL to
+ unregister a callback.
 */
 void PaAsio_RegisterResetRequestCallback( PaAsioResetRequestCallback callback );
 
--- a/include/pa_asio.h
+++ b/include/pa_asio.h
@@ -71,6 +71,20 @@ PaError PaAsio_GetAvailableBufferSizes( PaDeviceIndex device,
         long *minBufferSizeFrames, long *maxBufferSizeFrames, long *preferredBufferSizeFrames, long *granularity );
 
 
+typedef void (*PaAsioResetRequestCallback)();
+
+/** Register callback to detect reset requests.
+
+ This correponds to the ASIO kAsioResetRequest and kAsioBufferSizeChange
+ messages.
+
+ @param callback The function to call on ASIO reset request.  This is called
+ when the driver's buffer size has changed or it needs a reset, usually in
+ response to a user changing the configuration.  Pass NULL to unregister a
+ callback.
+*/
+void PaAsio_RegisterResetRequestCallback( PaAsioResetRequestCallback callback );
+
 /** Backwards compatibility alias for PaAsio_GetAvailableBufferSizes
 
  @see PaAsio_GetAvailableBufferSizes
--- a/src/hostapi/asio/pa_asio.cpp
+++ b/src/hostapi/asio/pa_asio.cpp
@@ -276,7 +276,7 @@ typedef struct PaAsioDriverInfo
     ASIODriverInfo asioDriverInfo;
     long inputChannelCount, outputChannelCount;
     long bufferMinSize, bufferMaxSize, bufferPreferredSize, bufferGranularity;
-    bool postOutput, bufferSizeOutdated;
+    bool postOutput;
 }
 PaAsioDriverInfo;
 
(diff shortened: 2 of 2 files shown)


RossBencina:
I'm not sure whether it is the best approach but it is not bad. `bufferSizeOutdated` needs to be atomic.

I wonder whether a single generic callback would be better than individual callbacks for each event type.

npostavs:
> `bufferSizeOutdated` needs to be atomic.

Ah, it's been a while since I've had to think about threads. Do you mean [`std::atomic`](https://en.cppreference.com/w/cpp/atomic/atomic)?

> I wonder whether a single generic callback would be better than individual callbacks for each event type.

Hmm, probably yes. I started out thinking that just `kAsioBufferSizeChange` would give me the all the info I wanted. But then it turns out my device sends `kAsioResetRequest` when I change the buffer size (I assume this is typical). And I end up ignoring the `newBufferSize` parameter in the code anyway.

src/hostapi/asio/pa_asio.cpp
```diff
@@ -928,6 +928,29 @@ PaError PaAsio_GetAvailableBufferSizes( PaDeviceIndex device,
             PaAsioDeviceInfo *asioDeviceInfo =
                     (PaAsioDeviceInfo*)hostApi->deviceInfos[hostApiDevice];
 
+            PaAsioHostApiRepresentation *asioApi = (PaAsioHostApiRepresentation*) hostApi;
+            if ( hostApiDevice == asioApi->openAsioDeviceIndex &&
+                 asioApi->openAsioDriverInfo.bufferSizeOutdated )
+            {
+                ASIOError asioError;
+                if( (asioError = ASIOGetBufferSize(&asioApi->openAsioDriverInfo.bufferMinSize,
```
philburk, line 936: Maybe declare some variable like:

    auto info = asioApi->openAsioDriverInfo;

then use that to shorten the long lines below.

include/pa_asio.h
```diff
@@ -70,6 +70,26 @@ extern "C"
...
+/** Register callback to detect changes to buffer size.
+
+ This correponds to the ASIO kAsioBufferSizeChange message.  Note that an ASIO
+ driver may send a more general reset request when the user changes the buffer
+ size.
+
+ @param callback The function to call when ASIO buffer size changes.
+ It will receive the new buffer size as a parameter.
+*/
+void PaAsio_RegisterBufferSizeChangeCallback( void (*callback)(long newBufferSize) );
```
philburk, line 82: You could declare a typedef for the callback function, similar to PaStreamCallback.

https://github.com/PortAudio/portaudio/blob/f7efb6b53cf134ec9165da7f6a3c0172371b820a/include/portaudio.h#L830

include/pa_asio.h
```diff
@@ -70,6 +70,26 @@ extern "C"
 PaError PaAsio_GetAvailableBufferSizes( PaDeviceIndex device,
         long *minBufferSizeFrames, long *maxBufferSizeFrames, long *preferredBufferSizeFrames, long *granularity );
 
+/** Register callback to detect changes to buffer size.
+
+ This correponds to the ASIO kAsioBufferSizeChange message.  Note that an ASIO
+ driver may send a more general reset request when the user changes the buffer
+ size.
+
+ @param callback The function to call when ASIO buffer size changes.
```
philburk, line 79: Can this be called with nullptr to Unregister the callback?

src/hostapi/asio/pa_asio.cpp
```diff
@@ -3199,6 +3223,33 @@ static void sampleRateChanged(ASIOSampleRate sRate)
     PA_DEBUG( ("sampleRateChanged : %d \n", sRate));
 }
 
+static void (*bufferSizeChangeCallback)(long newBufferSize);
```
philburk, line 3226: According to Naming conventions:
https://github.com/PortAudio/portaudio/wiki/ImplementationStyleGuidelines

> All static variables end with _

philburk (commented):
Thanks for contributing.

include/pa_asio.h
npostavs: Yes, now added a note about that to the doc

npostavs:
> > `bufferSizeOutdated` needs to be atomic.
> 
> Ah, it's been a while since I've had to think about threads. Do you mean [`std::atomic`](https://en.cppreference.com/w/cpp/atomic/atomic)?

According to https://github.com/PortAudio/portaudio/wiki/ImplementationStyleGuidelines, C++ code should be C++98 so `std::atomic` wouldn't be available. But perhaps the guidelines are out of date?

I'm also unsure about the thread safety of `ASIOGetBufferSize`.

npostavs:
I'm thinking perhaps it would be simpler to just drop all the attempts at updating the buffer size, leaving just the callbacks. The library user would be responsible for calling `Pa_Terminate()` and `Pa_Initialize()` to get updated info.

This would push all the threading trouble onto the caller. And there can be changes other than buffer size anyway.

RossBencina:
@npostavs do you think #519 would supersede this?

npostavs:
As I mentioned there, #519 doesn't seem to cover the case of messages sent before a stream is opened.

RossBencina:
@npostavs I don't think it is possible to receive callbacks before the stream is opened. Your code does not handle that either: the callbacks `asioCallbacks_ ` are set during `OpenStream`.

npostavs:
> I don't think it is possible to receive callbacks before the stream is opened.

Oh, you're right. I managed to confuse myself into thinking it could work (possibly I got a bit mixed up between opening vs starting the stream). I believe #519 supersedes this one then.

RossBencina:
Closing as superseded by #519.

PR #478, opened 2021-01-28
pa_mac_core.c: add missing definition to fix build on older systems

ra1nb0w:
thanks to @kencu

### DIFF
--- a/src/hostapi/coreaudio/pa_mac_core.c
+++ b/src/hostapi/coreaudio/pa_mac_core.c
@@ -71,6 +71,10 @@
 #include "pa_mac_core_utilities.h"
 #include "pa_mac_core_blocking.h"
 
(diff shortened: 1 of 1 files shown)


ra1nb0w:
ping

philburk (approved):

ra1nb0w:
thank you. FYI: now it works fine on the new Apple M1 (arm64).

Be-ing:
@ra1nb0w can you please clarify what build system you used that worked on M1? Autotools or CMake? Can you test #489? We need someone with M1 hardware to confirm it works.

ra1nb0w:
Ops. Sorry. Wrong comment, I messed up. These fix make build available on system less than 10.6 that doesn't have the define.
At macports we are using autotools. I don't have an M1 yet; probably in a few days I can test it if can wait..

Issue #392, opened 2020-12-23
Improper usage of FormatMessage


MichalPetryka:
**Describe the bug**
I've found 3 files that use FormatMessage macro that expect UTF8 data, which will result with broken non ASCII text when compiling with MBCS and totally unreadable data with UNICODE:
https://github.com/PortAudio/portaudio/blob/master/src/os/win/pa_win_coinitialize.c#L93
https://github.com/PortAudio/portaudio/blob/master/src/hostapi/wmme/pa_win_wmme.c#L345
https://github.com/PortAudio/portaudio/blob/master/src/hostapi/asio/pa_asio.cpp#L218

**To Reproduce**
n/a, look at linked code

**Expected behavior**
Code should use FormatMessageW and convert the data with WideCharToMultiByte into a temporary buffer (since the buffer size can't be constant, it'll need to be allocated and freed with one of the helpers)

**Actual behavior**
The code ignores the encoding and passes the received data into host error text.

**Desktop (please complete the following information):**
 - OS: Windows
 - OS Version n/a
 - PortAudio version: stable, nightly snapshot (which?), current (please give date and/or Git hash): current master (23.12.2020), probably all versions
 - If Windows or Linux, which Host API (e.g. WASAPI): MME, ASIO, also CoInitialize helper

RossBencina:
Related #200

MichalPetryka:
Best solution would be to make a helper function in some OS specific header, that would dynamically allocate a buffer for UTF-8 and convert from UTF-16 to it. Current code in MME and DS could be replaced with that.

jmelas:
This can be closed now

Issue #176, opened 2011-05-10
consolodate windows error handling functions to improve error reporting quality


PortAudio-admin:
*Issue created by @RossBencina*

Windows uses a common error code system across most APIs (not sure about WMME wave* functions). PortAudio error reporting could be improved by providing standard functions for translating these error codes into PA errors.

Solution: create a pa_win_error.c/h module with routines for translating Windows error codes to PA error codes. Including support for translating unexpected host errors.

pa_win_wmme.c  contains a function PaMme_SetLastSystemError that could be used as a basis.

Function signature would be something like:

```
/**
 Convert a Windows error code into a PaError. Sets the host-api specific error
 information if needed.

 @param hostApiType The calling host api type. Used when reporting paUnanticipatedHostError

 @param winError A Windows error code.
*/
PaError PaWinUtil_ConvertSystemError( PaHostApiTypeId hostApiType, DWORD winError );
```

jmelas:
This can be closed now

PR #495, opened 2021-02-06
PaWinUtil_SetLastSystemErrorInfo

jmelas:
This PR addresses items #176 and #392.

For #176:
It creates a new function `PaWinUtil_ConvertSystemError` which gets the text of a win32 error and passes it
to `PaUtil_SetLastHostErrorInfo`. It changes calls to all win hostapis to call the new function which greatly reduces code duplication.

For #392:
`PaWinUtil_ConvertSystemError` explicitly calls `FormatMessageW` and converts the message buffer.

Additional fixes:
1. `PaWinUtil_ConvertSystemError` does not allocate memory anymore. It uses a stack buffer of size 1024 which is the size of the buffer we use in `PaUtil_SetLastHostErrorInfo` (`PA_LAST_HOST_ERROR_TEXT_LENGTH_==1024`).
2. It adds the `FORMAT_MESSAGE_IGNORE_INSERTS` flag which maybe needed in some cases as explained by Raymond Chen here: https://devblogs.microsoft.com/oldnewthing/20071128-00/?p=24353

### DIFF
--- a/src/hostapi/dsound/pa_win_ds.c
+++ b/src/hostapi/dsound/pa_win_ds.c
@@ -83,6 +83,7 @@
 #include "pa_process.h"
 #include "pa_debugprint.h"
 
+#include "pa_win_util.h"
 #include "pa_win_ds.h"
 #include "pa_win_ds_dynlink.h"
 #include "pa_win_waveformat.h"
@@ -203,9 +204,14 @@ static signed long GetStreamReadAvailable( PaStream* stream );
 static signed long GetStreamWriteAvailable( PaStream* stream );
 
 
-/* FIXME: should convert hr to a string */
+#if _WIN32_WINNT >= 0x0602 // Windows 8 and above
+#define PA_DS_SET_LAST_DIRECTSOUND_ERROR( hr ) \
+    PaWinUtil_SetLastSystemErrorInfo( paDirectSound, hr )
+#else
+/* FIXME: should use DXGetErrorString/DXGetErrorDescription for Windows 7 and below */
 #define PA_DS_SET_LAST_DIRECTSOUND_ERROR( hr ) \
     PaUtil_SetLastHostErrorInfo( paDirectSound, hr, "DirectSound error" )
+#endif
 
 /************************************************* DX Prototypes **********/
 static BOOL CALLBACK CollectGUIDsProcW(LPGUID lpGUID,
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -126,6 +126,7 @@ IF(WIN32)
     src/os/win/pa_win_wdmks_utils.c
     src/os/win/pa_win_coinitialize.c)
   SET(PA_PLATFORM_INCLUDES
+    src/os/win/pa_win_util.h
     src/os/win/pa_win_coinitialize.h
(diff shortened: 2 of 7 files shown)


MichalPetryka:
`FormatMessageA` won't give correct results due to it using ANSI, not UTF-8, the code should use `FormatMessageW` and convert the text top UTF-8 with `WideCharToMultiByte`.

jmelas:
> `FormatMessageA` won't give correct results due to it using ANSI, not UTF-8, the code should use `FormatMessageW` and convert the text to UTF-8 with `WideCharToMultiByte`.

Personally I would prefer to enusre that error messages are always in english. Asio library does that too.
However I am not 100% sure that `MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT)` assures english language.
Maybe we can replace it with `MAKELANGID(LANG_ENGLISH, SUBLANG_ENGLISH_US)`

MichalPetryka:
1. Not sure if forcing english is a good idea.
2. Even then, things like paths could still contain unicode characters, *A functions are both a bad practice and slower (cause most of them convert all data from Unicode to ANSI and vice versa).

src/hostapi/dsound/pa_win_ds.c
```diff
@@ -203,9 +204,8 @@ static signed long GetStreamReadAvailable( PaStream* stream );
 static signed long GetStreamWriteAvailable( PaStream* stream );
 
 
-/* FIXME: should convert hr to a string */
 #define PA_DS_SET_LAST_DIRECTSOUND_ERROR( hr ) \
-    PaUtil_SetLastHostErrorInfo( paDirectSound, hr, "DirectSound error" )
+    PaWinUtil_ConvertSystemError( paDirectSound, hr )
```
MichalPetryka, line 208: According to stackoverflow (https://stackoverflow.com/questions/13543195/how-can-i-retrieve-error-strings-for-directx-9 and https://walbourn.github.io/wheres-dxerr-lib/) FormatMessage seems to only work with DirectSound on Windows 8 and higher, before that deprecated DXGetErrorString() and DXGetErrorDescription() seem the correct way.

RossBencina: Windows 7 is still a valid target so this is reasonable feedback. Probably best option is to use the DirectSound deprecated APIs for pa_win_ds.c, and possibly provide a conditional compliation to use the new `PaWinUtil_ConvertSystemError` function if and only if targeting Windows 8 and higher.

RossBencina:
`PaWinUtil_ConvertSystemError` is not the correct name, because it does not simply "convert", it changes global state by calling `PaUtil_SetLastHostErrorInfo`. It should be called something like `PaWinUtil_SetLastSystemErrorInfo` -- I'm open to better suggestions.

src/hostapi/dsound/pa_win_ds.c
jmelas: Unfortunately I do not have a Windows 7 machine to test DXGetErrorString so let's leave it for a different PR!
I've provided conditional compilation so that for Windows 7 and earlier old code is used.

philburk (approved):

RossBencina (approved):

PR #496, opened 2021-02-09
wasapi: add PaWasapi_GetIMMDevice function

rouseabout:
The IMMDevice pointer enables access to audio device capabiltiies not exposed by PortAudio.
https://docs.microsoft.com/en-us/windows/win32/api/mmdeviceapi/nn-mmdeviceapi-immdevice

### DIFF
--- a/build/msvc/portaudio.def
+++ b/build/msvc/portaudio.def
@@ -55,3 +55,4 @@ PaWasapi_GetJackDescription         @66
 PaWasapi_SetStreamStateHandler      @68
 PaWasapiWinrt_SetDefaultDeviceId    @67
 PaWasapiWinrt_PopulateDeviceList    @69
+PaWasapi_GetIMMDevice               @70
--- a/cmake_support/template_portaudio.def
+++ b/cmake_support/template_portaudio.def
@@ -58,3 +58,4 @@ PaUtil_SetDebugPrintFunction        @55
 @DEF_EXCLUDE_WASAPI_SYMBOLS@PaWasapi_SetStreamStateHandler      @68
 @DEF_EXCLUDE_WASAPI_SYMBOLS@PaWasapiWinrt_SetDefaultDeviceId    @67
 @DEF_EXCLUDE_WASAPI_SYMBOLS@PaWasapiWinrt_PopulateDeviceList    @69
+@DEF_EXCLUDE_WASAPI_SYMBOLS@PaWasapi_GetIMMDevice               @70
--- a/include/pa_win_wasapi.h
+++ b/include/pa_win_wasapi.h
@@ -437,6 +437,16 @@ int PaWasapi_GetDeviceMixFormat( void *pFormat, unsigned int formatSize, PaDevic
 int/*PaWasapiDeviceRole*/ PaWasapi_GetDeviceRole( PaDeviceIndex device );
 
 
+/** Get device IMMDevice pointer
+
+ @param device Device index.
+ @param pAudioClient Pointer to pointer of IMMDevice.
+
+ @return Error code indicating success or failure.
+*/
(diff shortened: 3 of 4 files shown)


dmitrykos:
@rouseabout, would you please remove the change to the Makefile.in file as it also affects WinMME which is outside the scope of the proposed change, it should be separate PR I believe.

For the addition of PaWaspi_GetIMMDevice symbol would you please modify https://github.com/PortAudio/portaudio/blob/master/build/msvc/portaudio.def instead by adding new function to its end.

Please squash all changes into a single commit.

MichalPetryka:
@dmitrykos You can use this to have GitHub squash it for you.
image

dmitrykos:
@MichalPetryka I am aware of this functionality but if I am not mistaken squashing commits by the author of the PR is the rule of thumb in PortAudio Git. I could not find the reference to it in the Wiki and may be mistaken though, may be @RossBencina could clarify it.

Besides that there are changes requested that do not allow me to merge this PR right away.

dechamps:
> squashing commits by the author of the PR is the rule of thumb in PortAudio Git

There is no "rule of thumb" in PortAudio Git. We had this discussion several times already in various PRs. My personal suggestion would just be to do whatever seems appropriate depending on the state of the PR: either rebase (if the commits are atomic and individually self-consistent) or squash (if the branch looks messy). (I would prefer merge commits never be used unless in exceptional circumstances, as they make history much harder to read.)

dmitrykos:
@dechamps thank you for your input. Personally I like a single commit with all changes, especially if there were changes due to the PR review process. It is much easier to read the PR later on if there is a need to look back for a changes.

dechamps:
> Personally I like a single commit (by PR author) with all changes, especially if there were changes due to the PR review process.

Again, that depends.

Usually, when I send a PR, I respond to feedback by changing the commits themselves (i.e. changing history) and force-pushing. This means the PR always has a clean history at all times and every commit has an up-to-date, consistent diff. In this case it is better to rebase the PR because that preserves the rich, fine-grained history information. The PR is always kept in a "rebase-ready" state.

Other authors prefer to address feedback on PRs by adding commits instead of changing history. This results in a branch with a "messy" history (e.g. commits that are not consistent and might not even build or pass CI). It is better to squash such CLs to avoid polluting the history of the master branch, provided the PR is small.

And then there are large messy PRs that we don't want to squash into a single commit because the resulting commit would be too large and unwieldy. In that case the reviewer can simply ask the author to clean up their history, and rebase once that's done.

dmitrykos:
@dechamps I agree that approach needs to be flexible depending on the situation and do not argue about it.

For this given PR I prefer to have clean single commit without a history because it is minor and is an addition of the new single function. The reference to Makefile.in should not be in it because it is outside the scope of the change. Moreover https://github.com/PortAudio/portaudio/blob/master/build/msvc/portaudio.def needs to be changed accordingly.

So I hope @rouseabout will update the PR soon so that I would be able to merge it.

rouseabout:
Will update. Wait. Portaudio has two def files, one for cmake one for msvc. But the symbol @ numbers different? I feel that needs to be fixed first before I add new symbols.

https://github.com/PortAudio/portaudio/blob/master/build/msvc/portaudio.def

https://github.com/PortAudio/portaudio/blob/master/cmake_support/template_portaudio.def

dmitrykos:
@rouseabout for both defs I guess we will need the same symbol Ids. 

It looks that `template_portaudio.def` is inconsistent regarding symbols Ids in comparison with `portaudio.def`. May be @dechamps  could comment about whether we need to fix it for WASAPI case in a separate PR as he modified that file recently.

dmitrykos:
@rouseabout just noticed new PR #499 regarding `template_portaudio.def` fix.

dmitrykos:
@rouseabout, looks fine!

PR #499, opened 2021-02-09
cmake_support/template_portaudio.def: make symbol numbers consistent …

rouseabout:
…with build/msvc/portaudio.def

Portaudio has two dll export def files, one for cmake one for msvc. But the symbol @ numbers different across files? Lets assume the msvc is more authorative, since that existed before the cmake build system.

https://github.com/PortAudio/portaudio/blob/master/build/msvc/portaudio.def

https://github.com/PortAudio/portaudio/blob/master/cmake_support/template_portaudio.def


dmitrykos:
@rouseabout thank you for fixing it.

PR #500, opened 2021-02-09
configure/make: export PaWasapi_* and PaWasapiWinrt_* symbols

rouseabout:
(i will create a separate pull request later that exports PaWinMME_* symbols and ammend the portaudio.defs)

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -44,7 +44,7 @@ PALIB = libportaudio.la
 PAINC = include/portaudio.h
 
 PA_LDFLAGS = $(LDFLAGS) $(SHARED_FLAGS) -rpath $(libdir) -no-undefined \
-	     -export-symbols-regex "(Pa|PaMacCore|PaJack|PaAlsa|PaAsio|PaOSS)_.*" \
+	     -export-symbols-regex "(Pa|PaMacCore|PaJack|PaAlsa|PaAsio|PaOSS|PaWasapi|PaWasapiWinrt)_.*" \
 	     -version-info $(LT_CURRENT):$(LT_REVISION):$(LT_AGE)
 
 COMMON_OBJS = \


Be-ing:
Maintaining this list of symbols will be obsolete after #461 using [CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS](https://cmake.org/cmake/help/latest/variable/CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS.html#variable:CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS).

rouseabout:
> Maintaining this list of symbols will be obsolete after #461 using [CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS](https://cmake.org/cmake/help/latest/variable/CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS.html#variable:CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS).

portaudio has three different build systems (cmake,  vcproj,  and configure/make). Does that mean you are planning to drop the other build systems?

rouseabout:
> Maintaining this list of symbols will be obsolete after #461 using [CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS](https://cmake.org/cmake/help/latest/variable/CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS.html#variable:CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS).

portaudio has three different build systems (cmake,  vcproj,  and configure/make). Does that mean you are planning to drop the other build systems?

Be-ing:
IMO it would be a big waste of effort and cause a lot of trouble for downstream users to maintain the redundant build systems.

RossBencina:
> IMO it would be a big waste of effort and cause a lot of trouble for downstream users to maintain the redundant build systems.

There are no plans to deprecate the other build systems.

philburk (approved):

PR #504, opened 2021-02-12
JACK PipeWire fixes (escape regex chars)

Be-ing:
This fixes various bugs in the JACK host API exposed by PipeWire. Refer to the commit messages for details. I think this should be included in the upcoming 19.7 release because Linux distros are about to switch to PipeWire by default in the next few months (Fedora 34 and Ubuntu 21.04 at least).

I usually write C++ and don't have much experience with C so please point out if I'm doing something wrong. The `replace_string` function might not be the most efficient, but it works, and this is not performance critical code or we wouldn't be using regexes here already.

### DIFF
--- a/src/hostapi/jack/pa_jack.c
+++ b/src/hostapi/jack/pa_jack.c
@@ -452,48 +452,36 @@ BlockingWaitEmpty( PaStream *s )
 
 /* ---- jack driver ---- */
 
-static void replace_string(char *string, size_t string_buffer_size, const char *to_find, const char *replacement)
+/* copy null terminated string source to destination, escaping regex characters with '\\' in the process */
+static void copy_string_and_escape_regex_chars( char *destination, const char *source, size_t destbuffersize )
 {
-    char *found_position = strstr(string, to_find);
-    while( found_position ) {
-        size_t original_length = strlen(string);
-        size_t found_length = strlen(to_find);
-        size_t replacement_length = strlen(replacement);
+    assert( destination != source );
+    assert( destbuffersize > 0 );
 
-        size_t destination_length = original_length - found_length + replacement_length;
-        if( destination_length >= string_buffer_size ) {
-            PaUtil_DebugPrint( "Cannot replace %s with %s in %s, buffer size is too small\n",
-                               to_find, replacement, string );
-            return;
+    char *dest = destination;
+    /* dest_stop is the last location that we can null-terminate the string */
+    char *dest_stop = destination + (destbuffersize - 1);
+
+    const char *src = source;
+
+    while ( *src != '\0' && dest != dest_stop )
+    {
+        const char c = *src;
+        if ( strchr( "\\()[]{}*+?|$^.", c ) != NULL )
+        {
+            if( (dest + 1) == dest_stop )
+                break; /* only proceed if we can write both c and the escape */
+
+            *dest = '\\';
+            dest++;
         }
+        *dest = c;
+        dest++;
 
-        memmove(found_position + replacement_length,
-                found_position + found_length,
-                // add 1 to include the \0 terminator
-                string + original_length - (found_position + found_length) + 1);
-        memcpy(found_position, replacement, replacement_length);
-
-        found_position = strstr(found_position + replacement_length, to_find);
+        src++;
     }
-}
 
-static void escape_regex_chars(char* string, size_t buffersize)
-{
-    // Escaping \ must be first because the following replacements insert \ chars into the string.
-    replace_string(string, buffersize, "\\", "\\\\");
-    replace_string(string, buffersize, "(", "\\(");
-    replace_string(string, buffersize, ")", "\\)");
-    replace_string(string, buffersize, "[", "\\[");
-    replace_string(string, buffersize, "]", "\\]");
-    replace_string(string, buffersize, "{", "\\{");
-    replace_string(string, buffersize, "}", "\\}");
-    replace_string(string, buffersize, "*", "\\*");
-    replace_string(string, buffersize, "+", "\\+");
-    replace_string(string, buffersize, "?", "\\?");
-    replace_string(string, buffersize, "|", "\\|");
-    replace_string(string, buffersize, "$", "\\$");
-    replace_string(string, buffersize, "^", "\\^");
-    replace_string(string, buffersize, ".", "\\.");
+    *dest = '\0';
 }
 
 /* BuildDeviceList():
@@ -637,8 +625,9 @@ static PaError BuildDeviceList( PaJackHostApiRepresentation *jackApi )
 
         /* To determine how many input and output channels are available,
          * we re-query jackd with more specific parameters. */
-        strncpy( device_name_regex_escaped, client_names[client_index], device_name_regex_escaped_size );
-        escape_regex_chars( device_name_regex_escaped, device_name_regex_escaped_size );
+        copy_string_and_escape_regex_chars( device_name_regex_escaped,
+                            client_names[client_index],
+                            device_name_regex_escaped_size );
         strncpy( port_regex_string, device_name_regex_escaped, port_regex_size );
         strncat( port_regex_string, port_regex_suffix, port_regex_size );
 
@@ -1304,10 +1293,9 @@ static PaError OpenStream( struct PaUtilHostApiRepresentation *hostApi,
         int err = 0;
 
         /* Get output ports of our capture device */
-        strncpy( regex_escaped_client_name,
+        copy_string_and_escape_regex_chars( regex_escaped_client_name,
                  hostApi->deviceInfos[ inputParameters->device ]->name,
                  regex_escaped_client_name_length );
-        escape_regex_chars( regex_escaped_client_name, regex_escaped_client_name_length );
         strncpy( regex_pattern, regex_escaped_client_name, regex_size );
         strncat( regex_pattern, port_regex_suffix, regex_size );
         UNLESS( jack_ports = jack_get_ports( jackHostApi->jack_client, regex_pattern,
@@ -1333,10 +1321,9 @@ static PaError OpenStream( struct PaUtilHostApiRepresentation *hostApi,
         int err = 0;
 
         /* Get input ports of our playback device */
-        strncpy( regex_escaped_client_name,
+        copy_string_and_escape_regex_chars( regex_escaped_client_name,
                  hostApi->deviceInfos[ outputParameters->device ]->name,
                  regex_escaped_client_name_length );
-        escape_regex_chars( regex_escaped_client_name, regex_escaped_client_name_length );
         strncpy( regex_pattern, regex_escaped_client_name, regex_size );
         strncat( regex_pattern, port_regex_suffix, regex_size );
         UNLESS( jack_ports = jack_get_ports( jackHostApi->jack_client, regex_pattern,
--- a/src/hostapi/jack/pa_jack.c
+++ b/src/hostapi/jack/pa_jack.c
@@ -520,7 +520,8 @@ static PaError BuildDeviceList( PaJackHostApiRepresentation *jackApi )
     char *port_regex_string = NULL;
     char *device_name_regex_escaped = NULL;
     // In the worst case scenario, every character would be escaped, doubling the string size.
-    size_t device_name_regex_escaped_size = jack_client_name_size() * 2;
+    // Add 1 for null terminator.
+    size_t device_name_regex_escaped_size = jack_client_name_size() * 2 + 1;
     size_t port_regex_size = device_name_regex_escaped_size + strlen(port_regex_suffix);
     int port_index, client_index, i;
     double globalSampleRate;
@@ -1140,10 +1141,11 @@ static PaError OpenStream( struct PaUtilHostApiRepresentation *hostApi,
     PaJackStream *stream = NULL;
     char *port_string = PaUtil_GroupAllocateMemory( jackHostApi->deviceInfoMemory, jack_port_name_size() );
     // In the worst case every character would be escaped which would double the string length.
-    size_t regex_escaped_client_name_length = jack_client_name_size() * 2;
+    // Add 1 for null terminator
+    size_t regex_escaped_client_name_length = jack_client_name_size() * 2 + 1;
     char *regex_escaped_client_name = PaUtil_GroupAllocateMemory(
         jackHostApi->deviceInfoMemory, regex_escaped_client_name_length );
-    unsigned long regex_size = jack_client_name_size() * 2 + strlen(port_regex_suffix);
+    unsigned long regex_size = jack_client_name_size() * 2 + 1 + strlen(port_regex_suffix);
     char *regex_pattern = PaUtil_GroupAllocateMemory( jackHostApi->deviceInfoMemory, regex_size );
     const char **jack_ports = NULL;
     /* int jack_max_buffer_size = jack_get_buffer_size( jackHostApi->jack_client ); */
--- a/src/hostapi/jack/pa_jack.c
+++ b/src/hostapi/jack/pa_jack.c
@@ -452,7 +452,7 @@ BlockingWaitEmpty( PaStream *s )
 
 /* ---- jack driver ---- */
 
-void replace_string(char *string, size_t string_buffer_size, const char *to_find, const char *replacement)
+static void replace_string(char *string, size_t string_buffer_size, const char *to_find, const char *replacement)
 {
     char *found_position = strstr(string, to_find);
     while( found_position ) {
@@ -477,7 +477,7 @@ void replace_string(char *string, size_t string_buffer_size, const char *to_find
     }
 }
 
-void escape_regex_chars(char* string, size_t buffersize)
+static void escape_regex_chars(char* string, size_t buffersize)
 {
     // Escaping \ must be first because the following replacements insert \ chars into the string.
     replace_string(string, buffersize, "\\", "\\\\");
--- a/src/hostapi/jack/pa_jack.c
+++ b/src/hostapi/jack/pa_jack.c
@@ -76,6 +76,7 @@
 static pthread_t mainThread_;
 static char *jackErr_ = NULL;
 static const char* clientName_ = "PortAudio";
+static const char* port_regex_suffix = ":.*";
 
 #define STRINGIZE_HELPER(expr) #expr
 #define STRINGIZE(expr) STRINGIZE_HELPER(expr)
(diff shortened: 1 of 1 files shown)


src/hostapi/jack/pa_jack.c
```diff
@@ -451,6 +451,49 @@ BlockingWaitEmpty( PaStream *s )
 
 /* ---- jack driver ---- */
 
+void replace_string(char *string, char *to_find, char *replacement)
+{
+    char *found_position = strstr(string, to_find);
+    while( found_position ) {
+        size_t sizeof_original = strlen(string);
```
daschuer, line 458: The name of the variable is misleading, because sizeof normally includes the null terminator, len not.
```suggestion
        size_t original_len = strlen(string);
```

src/hostapi/jack/pa_jack.c
```diff
@@ -451,6 +451,49 @@ BlockingWaitEmpty( PaStream *s )
...
+        size_t sizeof_destination = sizeof_original - sizeof_found + sizeof_replacement;
+        char *buffer = (char *)malloc((sizeof_destination + 1) * sizeof(char));
+
+        size_t length_before_find = found_position - string;
+        memcpy(buffer, string, length_before_find);
+        memcpy(buffer + length_before_find, replacement, sizeof_replacement);
+        memcpy(buffer + length_before_find + sizeof_replacement,
+              string + length_before_find + sizeof_found,
+              sizeof_original - length_before_find + sizeof_found);
+        memcpy(string, buffer, sizeof_destination);
```
daschuer, line 471: This is unsave, because you potentially write behind the original string without permission.
You should take a buffer size parameter as well to be save.

src/hostapi/jack/pa_jack.c
```diff
@@ -451,6 +451,49 @@ BlockingWaitEmpty( PaStream *s )
...
+void replace_string(char *string, char *to_find, char *replacement)
+{
+    char *found_position = strstr(string, to_find);
+    while( found_position ) {
+        size_t sizeof_original = strlen(string);
+        size_t sizeof_found = strlen(to_find);
+        size_t sizeof_replacement = strlen(replacement);
+
+        size_t sizeof_destination = sizeof_original - sizeof_found + sizeof_replacement;
+        char *buffer = (char *)malloc((sizeof_destination + 1) * sizeof(char));
```
daschuer, line 463: You can get around this malloc and free, by copy the tailing part byte by byte to the new location. memmove() is your friend here.

Be-ing:
Regarding the colons in device names, I have [opened an issue on PipeWire's tracker](https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/714) because this violates an old JACK convention. Carla also has problems with `XONE:K2`

src/hostapi/jack/pa_jack.c
```diff
@@ -451,6 +451,49 @@ BlockingWaitEmpty( PaStream *s )
...
+              sizeof_original - length_before_find + sizeof_found);
+        memcpy(string, buffer, sizeof_destination);
+        string[sizeof_destination] = '\0';
+        free(buffer);
+
+        found_position = strstr(string + length_before_find + sizeof_replacement, to_find);
+    }
+}
+
+void escape_regex_chars(char* string) {
```
daschuer, line 479: This should also take a buffer size parameter.

src/hostapi/jack/pa_jack.c
```diff
@@ -584,11 +636,13 @@ static PaError BuildDeviceList( PaJackHostApiRepresentation *jackApi )
 
         /* To determine how many input and output channels are available,
          * we re-query jackd with more specific parameters. */
-
-        sprintf( regex_pattern, "%s:.*", client_names[client_index] );
+        strcpy( device_name_regex_escaped, client_names[client_index] );
```
daschuer, line 639: strcpy is not save, because it does no length check. A one stop shopping function from the standard library is snprintf() with "%s".

daschuer (changes requested):

src/hostapi/jack/pa_jack.c
```diff
@@ -451,6 +451,49 @@ BlockingWaitEmpty( PaStream *s )
 
 /* ---- jack driver ---- */
 
+void replace_string(char *string, char *to_find, char *replacement)
```
daschuer, line 454:
```suggestion
void replace_string(char *string, size_t size, const char *to_find, const char *replacement)
```

Be-ing:
> Regarding the colons in device names, I have opened an issue on PipeWire's tracker because this violates an old JACK convention. Carla and QJackCtl also have problems with XONE:K2

This was fixed in PipeWire.

src/hostapi/jack/pa_jack.c
Be-ing: It now checks the buffer size before memmove. Is this an acceptable alternative?

Be-ing:
Ugh I have been testing the wrong library because the CMake implementation installs to the wrong directory (`lib` instead of `lib64`). The last few commits are broken, working on it...

Be-ing:
Okay I think the code is safe now. I cleaned up the Git history and rebased. Please review again.

Be-ing:
With the current PipeWire master branch, this is working well with all my audio interfaces.

src/hostapi/jack/pa_jack.c
```diff
@@ -451,6 +452,50 @@ BlockingWaitEmpty( PaStream *s )
 
 /* ---- jack driver ---- */
 
+void replace_string(char *string, size_t buffer_length, char *to_find, char *replacement)
```
daschuer, line 455: buffer_length is ambiguous here, "buffer" is unrelated to \0 while the therm length usually excludes it.  Without const, we cannot use string literals. 
```suggestion
void replace_string(char *string, size_t string_size, const char *to_find, const char *replacement)
```

src/hostapi/jack/pa_jack.c
```diff
@@ -451,6 +452,50 @@ BlockingWaitEmpty( PaStream *s )
...
+        size_t replacement_length = strlen(replacement);
+
+        size_t destination_length = original_length - found_length + replacement_length;
+        if( destination_length >= buffer_length ) {
+            PaUtil_DebugPrint( "Cannot replace %s with %s in %s, buffer size is too small\n",
+                               to_find, replacement, string );
+            return;
+        }
+
+        size_t before_find_length = found_position - string;
```
daschuer, line 470: This looks wrong, because string is a negative pointer. You add string below anyway in almost all cases.

src/hostapi/jack/pa_jack.c
```diff
@@ -451,6 +452,50 @@ BlockingWaitEmpty( PaStream *s )
...
+        memmove(string + before_find_length + replacement_length,
+                string + before_find_length + found_length,
+                original_length - before_find_length);
```
daschuer, line 473: I think the copied num is wrong. We can also move the null terminator here. How about this. 
memmove(found_position + replacement_length,
                found_position + found_length,
                string + original_length - (found_position + found_length) + 1);

src/hostapi/jack/pa_jack.c
```diff
@@ -451,6 +452,50 @@ BlockingWaitEmpty( PaStream *s )
...
+        memcpy(string + before_find_length, replacement, replacement_length);
```
daschuer, line 474: simplify to: 
```suggestion
        memcpy(found_position, replacement, replacement_length);
```

src/hostapi/jack/pa_jack.c
```diff
@@ -451,6 +452,50 @@ BlockingWaitEmpty( PaStream *s )
...
+        string[destination_length] = '\0';
```
daschuer, line 475: This is redundant if we copy the null terminator above.

src/hostapi/jack/pa_jack.c
```diff
@@ -451,6 +452,50 @@ BlockingWaitEmpty( PaStream *s )
...
+
+        found_position = strstr(string + before_find_length + replacement_length, to_find);
```
daschuer, line 477:
```suggestion
        found_position = strstr(found_position + replacement_length, to_find);
```

daschuer (changes requested):

src/hostapi/jack/pa_jack.c
Be-ing: This works though

src/hostapi/jack/pa_jack.c
Be-ing: What we are concerned with here is not the chars + `\0`, it is the size we have malloced. So I think `string_buffer_size` is better than just `string_size`.

src/hostapi/jack/pa_jack.c
Be-ing: Good idea, this simplifies the code

src/hostapi/jack/pa_jack.c
Be-ing: yes that works to copy the `\0` terminator

Be-ing:
Thank you for the suggestions @daschuer. That simplified the code.

Be-ing:
@RossBencina @philburk I think this is ready to merge. Please review.

daschuer (approved):

daschuer:
I am not sure about removing the monitor channels. There are there for a reason and if is not the responsibility of a wrapper library to hide them. I think we need to add a renaming layer that makes them usable with Portaudio. Alternative we may make the ignore list accessible from the outside.

Can you put that in another PR to not slow down the integration process about this?

Be-ing:
I am quite confident about hiding the monitor channels. I can't actually figure out a reason why they exist. It doesn't make any sense to me and I've asked multiple times on the PipeWire issue tracker. Loopback inputs are a hack around the ability to arbitrarily route audio between applications, so they're unnecessary in PipeWire. Exposing them in PortAudio would be confusing because there is no context. In JACK applications like Catia or QJackCtl, the user can see the port names. In PortAudio all we have are devices with a number of channels without any names for the channels.

Be-ing:
Not hiding the monitor channels from would IMO be a significant regression compared to using PortAudio with jackd.

Be-ing:
Not hiding the monitor channels from would IMO be a significant regression compared to using PortAudio with jackd.

daschuer:
I still don't get the point why the application using portaudio should not decide over this. But anyway this is unrelated to the original PR. Please split that out to not delay both.

Be-ing:
The reason is that the PortAudio application *can't* decide about it because PortAudio's API only has concepts of devices with numbers of channels with no additional metadata.

Be-ing:
https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/736

daschuer:
I think a separate monitor device as proposed by @saivert is the way to go.
Let's postpone the issue until upstream has decided.

philburk:
Thanks for this PR.

> I can't actually figure out a reason why they exist.

Loopback channels are not often used.  But they can be handy for doing automated testing.

We are very close to getting 19.7 out. So I am hesitant to include this new code right before a stable snapshot.
Hopefully we will move to a more regular cadence for stable releases.

Be-ing:
PortAudio apps will be broken when Fedora 34 and Ubuntu 21.04 release in a few months without this.

Be-ing:
I can split the last commit to another branch, but the regex escaping is unambiguously something PortAudio needs to do.

Be-ing:
It would be better if PipeWire didn't create the confusing monitor ports by default that probably few users will use but confuse lots of others so PortAudio doesn't have to hack around them.

src/hostapi/jack/pa_jack.c
```diff
@@ -490,7 +539,9 @@ static PaError BuildDeviceList( PaJackHostApiRepresentation *jackApi )
      * associated with the previous list */
     PaUtil_FreeAllAllocations( jackApi->deviceInfoMemory );
 
-    regex_pattern = PaUtil_GroupAllocateMemory( jackApi->deviceInfoMemory, jack_client_name_size() + 3 );
+    port_regex_string = PaUtil_GroupAllocateMemory( jackApi->deviceInfoMemory, port_regex_size );
+    device_name_regex_escaped = PaUtil_GroupAllocateMemory(
```
philburk, line 543: jack_client_name_size() include the NUL character so this is OK.

philburk (approved):

philburk (commented):
I changed my approval to a comment that the code looks good. But we should not merge it into 19.7 because it hides the monitor channels. That may be a problem but it is not clear.

Be-ing:
I removed the commit hiding the monitor channels.

philburk (changes requested):
Hmm. It did not remove my approval. I will try clearing that by requesting changes.

philburk (commented):
Now setting back to "comment". We may or may not want to hide the monitor channels. That is not clear yet.

Be-ing:
@philburk can you clarify if it is okay with you to merge this for 19.7 without the commit hiding the loopback monitor ports?

philburk (approved):
Ah! Your commits and my reviews crossed in the mail.
I see you removed the code that hides the channels.
As you say, this code fixes a problem that may break apps in the near future.
I am approving it and we should consider this for 19.7.
Thanks for the contribution.

Be-ing:
The [usability issue with the monitor ports](https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/736) has been solved in PipeWire by presenting them to JACK as separate input devices.

RossBencina:
Looks OK at a glance, but still I need to review it properly. But this is too big a change to risk making days before a stable release.

philburk:
I agree that it is a bit too risky for merging right before a release. Let's address it ASAP after we tag the 19.7 release.

Be-ing:
Now that the usability issue with the loopback monitoring ports has been resolved in PipeWire, this PR only affects a minority of users who have devices that happen to have a regex token in their device name. So it is probably not a big deal to wait on it.

Be-ing:
It seems this is not as esoteric of an issue as I thought. More users [are affected by this](https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/788). Please reconsider merging this for 19.7.

dbenjaminmiller:
Yeah, this issue prevents me from using any PortAudio programs with pw-jack.

Holzhaus:
> Yeah, this issue prevents me from using any PortAudio programs with pw-jack.

Same here.

Be-ing:
I am once again asking you to reconsider merging this for 19.7. Please consider the impact not merging this will have. It would cause user support hassles and reflect badly upon all PortAudio applications on Linux, PipeWire, and Linux distributions. Telling users to compile an unmerged branch of one of an application's dependencies to make the application usable is not a good situation. This may not matter to you personally, but lots of Linux users still have bad memories of the rollout of PulseAudio and are quite concerned that they'll have to experience that again with PipeWire. Please don't contribute to making the rollout of PipeWire rough.

Be-ing:
I would be happy to reach out to the maintainers of the Fedora and Ubuntu portaudio packages and urge them to update to 19.7 for their upcoming releases that will enable PipeWire by default if this gets merged for 19.7.

prez:
I see that this did not make it into 19.7, which is a shame. 
Could you please merge this ASAP? In search for a solution to my audio problems, I switched to pipewire, and this seems to be the final piece preventing it (I use a couple of PortAudio applications) from being a viable alternative.

Be-ing:
@RossBencina can you please review this? It is only +84 −10.

rolandlo:
Having this PR merged would also benefit Xournal++. We depend on Portaudio working fine. More and more users will use PipeWire. It's sad if they get into conflict with apps using Portaudio.

Swiftb0y:
Considering PipeWire is getting increasingly popular (fedora switching to it as a default for example), it would be a shame if this patch won't be included.

uklotzde:
The in-place regex escape code might not be optimal but it looks reasonable and has been confirmed to work. I guess performance is not an issue here and could be improved later.

What matters more is that, as multiple users already confirmed, without this patch PortAudio + Jack would be dysfunctional in all distributions that are switching to PipeWire. It is expected to cause a lot of hassle and support request for applications that rely on PortAudio. We are the ones that users will blame and contact first if things are not working as expected.

Holzhaus:
@RossBencina Please reconsider including this fix in the next stable release. Without this, portaudio apps are unusable with JACK via pipewire on my machine.

src/hostapi/jack/pa_jack.c
```diff
@@ -451,6 +452,50 @@ BlockingWaitEmpty( PaStream *s )
...
+                found_position + found_length,
+                // add 1 to include the \0 terminator
+                string + original_length - (found_position + found_length) + 1);
+        memcpy(found_position, replacement, replacement_length);
+
+        found_position = strstr(found_position + replacement_length, to_find);
+    }
+}
+
+void escape_regex_chars(char* string, size_t buffersize)
```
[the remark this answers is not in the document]
Be-ing: Sure, I don't care about the implementation details as long as it works and doesn't crash.

Be-ing:
Thank you for reviewing. I will take care of the review comments tonight. It is unfortunate that something as common as find and replacing in a string is so complicated and error prone in C.

I have just tested installing Mixxx in a fresh Fedora 34 beta virtual machine and finally it Just Works with the PortAudio JACK backend using PipeWire without grabbing exclusive access to the ALSA device or having to configure anything. Of course my RME Babyface Pro does not work in the VM because of this bug so I am eager to get this fix merged.

src/hostapi/jack/pa_jack.c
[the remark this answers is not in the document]
Be-ing: My original implementation allocated a temporary buffer but I changed it to memmove at @daschuer's suggestion. I don't really care either way.

src/hostapi/jack/pa_jack.c
```diff
@@ -472,7 +517,11 @@ static PaError BuildDeviceList( PaJackHostApiRepresentation *jackApi )
 
     const char **jack_ports = NULL;
     char **client_names = NULL;
-    char *regex_pattern = NULL;
+    char *port_regex_string = NULL;
+    char *device_name_regex_escaped = NULL;
+    // In the worst case scenario, every character would be escaped, doubling the string size.
+    size_t device_name_regex_escaped_size = jack_client_name_size() * 2;
```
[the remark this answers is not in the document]
Be-ing: Good point. Fixed: ca361337d5ce8df34710fd04ce7662cd3fa7bd2b

src/hostapi/jack/pa_jack.c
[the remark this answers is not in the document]
Be-ing: I am not sure your proposed method is a better approach. It only works for this one very specific task of escaping regex chars whereas my current implementation can be reused in the future for replacing arbitrary strings in arbitrary strings. Nevertheless I don't want to bikeshed over this. If you insist on changing the approach that you proposed, I will do it.

Be-ing:
@RossBencina The last commit changes to your implementation with some small changes for readability. In my experience, I find it confusing to manipulate a variable with the `++` operator or call a function with side effects in a variable assignment statement, so I moved the `dest++` calls to their own lines. I would prefer to merge everything except the last commit because the `replace_string` function is more flexible and could be used for other situations in the future. But I'm not going to bikeshed over this implementation detail; both implementations work. I tested the current state of this branch with the following devices using the latest PipeWire master branch (db85339f509e8c1bc39bb90fd6a4e6327b257964) and confirm they all work with Mixxx using the PortAudio JACK backend:

```
Babyface Pro (Class Compliant Mode) Pro
Traktor Kontrol S4 MK3 Pro
Built-in Audio Analog Stereo
XONE K2 Analog Surround 4.0
```

src/hostapi/jack/pa_jack.c
```diff
@@ -451,6 +452,43 @@ BlockingWaitEmpty( PaStream *s )
...
+        const char c = *src;
+        if ( strchr( "\\()[]{}*+?|$^.", c ) != NULL )
+        {
+            if( (dest + 1) == dest_stop )
+                break; /* only proceed if we can write both c and the escape */
+
+            *dest = '\\';
+            dest++;
+
+            *dest = c;
```
daschuer, line 478: This is redundant with the else branch

daschuer (approved):
Only a minor nits. the rest LGTM.

src/hostapi/jack/pa_jack.c
Be-ing: Good point. I simplified this and force pushed.

philburk (approved):
Thanks Be-ing for finding and fixing the problem. Also thanks for explaining the impact of not fixing this bug.
Thanks to the reviewers.
And thanks Ross for the simplified regex escape function. It is very clear, which reduces the risk of this late change.

Be-ing:
Great, thank you for merging. It will be easier to get distributions to package this fix now that it has been merged upstream instead of trying to convince them to ship a patch. I have opened a bug report for Fedora to update. @Holzhaus could you notify the Arch packagers to update PortAudio when 19.7 is released? Does anyone want to volunteer to reach out to Debian & Ubuntu?

Holzhaus:
On Arch, it's possible to flag a package as out of date here: https://archlinux.org/packages/extra/x86_64/portaudio/
I can do that when portaudio 19.7 is relased, but in my experience the Arch package maintainers deal with updates in less than a day anyway.

Be-ing:
This fix will be included in Fedora 34 on day 0: https://bugzilla.redhat.com/show_bug.cgi?id=1939749#c4

PR #506, opened 2021-02-12
coreaudio: remove pa_mac_core_old.c

### LINKED ISSUE #493: Consider removing pa_mac_core_old.c
`pa_mac_core_old.c` is not used anymore so we could consider removing it

philburk:
It has not been used for years.

Fixes #493


philburk:
Ross pre-approved this.

PR #513, opened 2021-02-18
pablio whitespace cleanup

MichalPetryka:
Since #437 seems to be on hold for 19.7, I've decided to do this for now.


philburk:
We are going to delete PABLIO so we no point in cleaning it up.

PR #519, opened 2021-02-23
Add support for ASIO messages

EvanBalster:
This PR addresses PortAudio's bad habit of ignoring certain highly actionable messages from ASIO drivers by providing a callback mechanism to the application.  Applications may receive the following events:

- Reset Request
- Resync Request
- Sample Rate Changed
- Buffer Size Changed
- Latencies Changed

In particular, Reset Request is commonly used to notify the ASIO client when the stream should be reset in order to put configuration changes into effect or resume the interrupted flow of audio.  Implementing automatic reset allows the user to interactively re-configure the ASIO stream by way of a driver-specific UI without superfluous manual steps in the client application.

I have been using this functionality in production for over four years in an application with an integrated crash forensics system and haven't identified any faults.

### DIFF
--- a/include/pa_asio.h
+++ b/include/pa_asio.h
@@ -88,6 +88,75 @@ PaError PaAsio_ShowControlPanel( PaDeviceIndex device, void* systemSpecific );
 
 
 
+/** ASIO message types.
+    
+ These mostly correspond with asioMessage calls from the ASIO SDK.
+ ASIO's sampleRateDidChange is adapted to use this callback.
+ Refer to ASIO SDK documentation for complete information.
+*/
+typedef enum PaAsioMessageType
+{
+    /** The driver requests that it be reset (by closing and re-opening the stream).
+        Typically dispatched when the user changes driver settings.
+        Recommend closing, re-opening and restarting stream, and always returning 1.
+            
+        params:
+            none.
+    */
+    paAsioResetRequest      = 1,
+    
+    /** Informs the application that a sample-rate change was detected.
+        Recommend noting the new sample-rate, but no action is needed.
+        
+        params:
+            opt[0] -- the new sample-rate.
+    */
+    paAsioSampleRateChanged = 2,
+    
+    /** Informs the application that the driver has a new preferred buffer size.
+        Recommend handling like paAsioResetRequest.
+        
+        params:
+            value -- the new preferred buffer size.
+    */
+    paAsioBufferSizeChange  = 3,
+    
+    /** Informs the application that the driver has gone out of sync, invalidating timestamps.
+        Recommend handling like paAsioResetRequest.
+        
+        params:
+            none.
+    */
+    paAsioResyncRequest     = 4, 
+    
+    /** Informs the application that the driver's latencies have changed.
+        FIXME: The only way to query the new latencies is to reset the stream.
+        Recommend ignoring unless latency reporting is critical.
+        
+        params:
+            none.
+    */
+    paAsioLatenciesChanged = 5,
+    
+} PaAsioMessageType;
+/** ASIO message callback, set in PaAsioStreamInfo.
+    Do not call PortAudio or PaAsio functions inside this callback!
+ 
+ @param value Message-specific integer value.
+ Indicates buffer size in paAsioBufferSizeChange.
+ 
+ @param message Message-specific pointer value.
+ Unused as of the ASIO 2.2 SDK.
+ 
+ @param opt Message-specific double value.
+ opt[0] indicates sample rate in paAsioSampleRateChange.
+ 
+ @param userData The value of a user supplied pointer passed to
+ Pa_OpenStream() intended for storing synthesis data etc.
+    
+ @return True if the application handled the message, false otherwise.
+*/
+typedef long PaAsio_MessageCallback( long messageType, long value, void *message, double *opt, void *userData );
 
 /** Retrieve a pointer to a string containing the name of the specified
  input channel. The string is valid until Pa_Terminate is called.
@@ -125,7 +194,7 @@ PaError PaAsio_SetStreamSampleRate( PaStream* stream, double sampleRate );
 typedef struct PaAsioStreamInfo{
     unsigned long size;             /**< sizeof(PaAsioStreamInfo) */
     PaHostApiTypeId hostApiType;    /**< paASIO */
-    unsigned long version;          /**< 1 */
+    unsigned long version;          /**< 2 */
 
     unsigned long flags;
 
@@ -140,6 +209,13 @@ typedef struct PaAsioStreamInfo{
         result.
     */
     int *channelSelectors;
+
+    /** ASIO message callback and user-defined parameter.
+        Unsupported in Blocking I/O mode.
+        Set to NULL if unused.
+        If a callback is supplied for both input and output, it will be called twice!
+    */
+    PaAsio_MessageCallback *messageCallback;
 }PaAsioStreamInfo;
 
 
--- a/src/hostapi/asio/pa_asio.cpp
+++ b/src/hostapi/asio/pa_asio.cpp
@@ -1304,6 +1304,16 @@ PaError PaAsio_Initialize( PaUtilHostApiRepresentation **hostApi, PaHostApiIndex
                 continue;
             }
 
+            // Blacklist dynamic loading of JackRouter in 32-bit applications.
+            // This causes an instant crash due to a long-standing address translation error.
+            //   https://github.com/jackaudio/jack2/issues/332
+
+            if (sizeof(void*)>4 && strcmp (names[i],"JackRouter") == 0)
+            {
+                PA_DEBUG(("BLACKLISTED!!!\n"));
+                continue;
+            }
+
 
             if( IsDebuggerPresent_ && IsDebuggerPresent_() )
             {
@@ -1638,6 +1648,9 @@ typedef struct PaAsioStream
     PaStreamCallbackFlags callbackFlags;
 
     PaAsioStreamBlockingState *blockingState; /**< Blocking i/o data struct, or NULL when using callback interface. */
+
+    /* Message callbacks copied from PaAsioStreamInfo */
+    PaAsio_MessageCallback *messageCallback[2];
 }
 PaAsioStream;
 
@@ -1880,22 +1893,38 @@ static PaError ValidateAsioSpecificStreamInfo(
 {
     if( streamInfo )
     {
-        if( streamInfo->size != sizeof( PaAsioStreamInfo )
+        switch( streamInfo->version )
                 || streamInfo->version != 1 )
         {
+        {
+        case 1:
+            /* NOTE: V1 structure's size is smaller by one pointer. */
+            if( streamInfo->size < sizeof( PaAsioStreamInfo ) - sizeof(PaAsio_MessageCallback) )
+                return paIncompatibleHostApiSpecificStreamInfo;
+        }
+            break;
+            
+        case 2:
+            if( streamInfo->size != sizeof( PaAsioStreamInfo ) )
+                return paIncompatibleHostApiSpecificStreamInfo;
+            break;
+            
+        default:
             return paIncompatibleHostApiSpecificStreamInfo;
         }
 
         if( streamInfo->flags & paAsioUseChannelSelectors )
+        {
             *channelSelectors = streamInfo->channelSelectors;
 
-        if( !(*channelSelectors) )
-            return paIncompatibleHostApiSpecificStreamInfo;
+            if( !(*channelSelectors) )
+                return paIncompatibleHostApiSpecificStreamInfo;
 
-        for( int i=0; i < streamParameters->channelCount; ++i ){
-            if( (*channelSelectors)[i] < 0
-                    || (*channelSelectors)[i] >= deviceChannelCount ){
-                return paInvalidChannelCount;
+            for( int i=0; i < streamParameters->channelCount; ++i ){
+                if( (*channelSelectors)[i] < 0
+                        || (*channelSelectors)[i] >= deviceChannelCount ){
+                    return paInvalidChannelCount;
+                }
             }
         }
     }
@@ -2011,7 +2040,7 @@ static PaError OpenStream( struct PaUtilHostApiRepresentation *hostApi,
     PaError result = paNoError;
     PaAsioHostApiRepresentation *asioHostApi = (PaAsioHostApiRepresentation*)hostApi;
     PaAsioStream *stream = 0;
-    PaAsioStreamInfo *inputStreamInfo, *outputStreamInfo;
+    PaAsioStreamInfo *inputStreamInfo = NULL, *outputStreamInfo = NULL;
     unsigned long framesPerHostBuffer;
     int inputChannelCount, outputChannelCount;
     PaSampleFormat inputSampleFormat, outputSampleFormat;
@@ -2750,6 +2779,19 @@ static PaError OpenStream( struct PaUtilHostApiRepresentation *hostApi,
     stream->postOutput = driverInfo->postOutput;
     stream->isStopped = 1;
     stream->isActive = 0;
+    
+    if( usingBlockingIo )
+    {
+        /* Disable message callback due to policy of overwriting userData above. */
+        stream->messageCallback[0] = NULL;
+        stream->messageCallback[1] = NULL;
+    }
+    else
+    {
+        /* Message callback may be defined in versions >= 2 of ASIO streamInfo. */
+        stream->messageCallback[0] = ( (  inputStreamInfo &&  inputStreamInfo->version >= 2 ) ?  inputStreamInfo->messageCallback : NULL );
+        stream->messageCallback[1] = ( ( outputStreamInfo && outputStreamInfo->version >= 2 ) ? outputStreamInfo->messageCallback : NULL );
+    }
 
     asioHostApi->openAsioDeviceIndex = asioDeviceIndex;
 
@@ -3185,6 +3227,26 @@ previousTime = paTimeInfo.currentTime;
 }
 
 
+static long fireMessageCallback( PaAsioStream *asioStream, long messageType, long value, void* message, double* opt )
+{
+    void *userData = asioStream->streamRepresentation.userData;
+        
+    long ret = 0;
+    
+    /* Fire callbacks and determine whether one and/or the other handles the message. */
+    if( asioStream->messageCallback[0] )
+    {
+        if( (*asioStream->messageCallback[0])( messageType, value, message, opt, userData ) ) ret = 1;
+    }
+    if( asioStream->messageCallback[1] )
+    {
+        if( (*asioStream->messageCallback[1])( messageType, value, message, opt, userData ) ) ret = 1;
+    }
+    
+    return ret;
+}
+
+
 static void sampleRateChanged(ASIOSampleRate sRate)
 {
     // TAKEN FROM THE ASIO SDK
@@ -3195,14 +3257,25 @@ static void sampleRateChanged(ASIOSampleRate sRate)
     // AES/EBU or S/PDIF digital input at the audio device.
     // You might have to update time/sample related conversion routines, etc.
 
-    (void) sRate; /* unused parameter */
-    PA_DEBUG( ("sampleRateChanged : %d \n", sRate));
+    double opt = sRate;
+
+    PA_DEBUG( ("asioSampleRateChanged : %f hz\n", opt));
+    
+    if (!theAsioStream)
+    {
+        /* Some ASIO drivers will fire messages during OpenStream.  We ignore them. */
+        PA_DEBUG( ("    ...message blocked, device not open yet.\n"));
+        return;
+    }
+    
+    fireMessageCallback( theAsioStream, paAsioSampleRateChanged, 0, 0, &opt );
 }
 
 static long asioMessages(long selector, long value, void* message, double* opt)
 {
 // TAKEN FROM THE ASIO SDK
     // currently the parameters "value", "message" and "opt" are not used.
+    int doClientCallback = 0, paAsioMessageType = 0;
     long ret = 0;
 
     (void) message; /* unused parameters */
@@ -3210,13 +3283,26 @@ static long asioMessages(long selector, long value, void* message, double* opt)
 
     PA_DEBUG( ("asioMessages : %d , %d \n", selector, value));
 
+    if (!theAsioStream)
+    {
+        /* Some ASIO drivers will fire messages during OpenStream.  We ignore them. */
+        PA_DEBUG( ("    ...message blocked, device not open yet.\n"));
+        return 0;
+    }
+
     switch(selector)
     {
         case kAsioSelectorSupported:
             if(value == kAsioResetRequest
-            || value == kAsioEngineVersion
             || value == kAsioResyncRequest
             || value == kAsioLatenciesChanged
+            || value == kAsioBufferSizeChange)
+            {
+                /* Did the client register a callback? */
+                if( theAsioStream->messageCallback[0] || theAsioStream->messageCallback[1] )
+                    ret = 1L;
+            }
+            if(value == kAsioEngineVersion
             // the following three were added for ASIO 2.0, you don't necessarily have to support them
             || value == kAsioSupportsTimeInfo
             || value == kAsioSupportsTimeCode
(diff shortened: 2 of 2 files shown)


EvanBalster:
Additional context:  This PR is a recreation of an old merge request on PortAudio's previous Assembla-hosted repository.  readers can find that Merge Request here...

https://app.assembla.com/spaces/portaudio/git/merge_requests/3939463

src/hostapi/asio/pa_asio.cpp
```diff
@@ -1304,6 +1304,16 @@ PaError PaAsio_Initialize( PaUtilHostApiRepresentation **hostApi, PaHostApiIndex
                 continue;
             }
 
+            // Blacklist dynamic loading of JackRouter in 32-bit applications.
+            // This causes an instant crash due to a long-standing address translation error.
+            //   https://github.com/jackaudio/jack2/issues/332
+
+            if (sizeof(void*)>4 && strcmp (names[i],"JackRouter") == 0)
```
MichalPetryka, line 1311: Won't this also block it in 64bit apps? Since a pointer is 8B there.

src/hostapi/asio/pa_asio.cpp
```diff
@@ -1880,22 +1893,35 @@ static PaError ValidateAsioSpecificStreamInfo(
 {
     if( streamInfo )
     {
-        if( streamInfo->size != sizeof( PaAsioStreamInfo )
-                || streamInfo->version != 1 )
+        switch( streamInfo->version )
         {
+        case 1:
+            /* NOTE: V1 structure's size is smaller by one pointer. */
+            if( streamInfo->size < sizeof( PaAsioStreamInfo ) - sizeof(PaAsio_MessageCallback*) )
```
MichalPetryka, line 1900: Shouldn't this do != instead of <?

src/hostapi/asio/pa_asio.cpp
EvanBalster: My memory is vague but it may actually be the comment that's incorrect.  As I recall, the defect in JackRouter involves attempting to store a 64-bit pointer from the application in a 32-bit value, later expanding it into an invalid pointer.  Some (re-)testing may be in order.

src/hostapi/asio/pa_asio.cpp
EvanBalster: This particular check is difficult as we can't perfectly predict what the size of the original PaAsioStreamInfo structure would be without fully declaring it.  I use `<` instead of `!=` in case some padding is placed before the new field (although this is unlikely because the previous field is also a pointer).

Interested in input from maintainers on this issue.

src/hostapi/asio/pa_asio.cpp
EvanBalster: Confirmed based on discussion here:  https://github.com/jackaudio/jack2/issues/275

64-bit applications will crash immediately upon opening the JackRouter ASIO driver.  I'll amend the comment.

src/hostapi/asio/pa_asio.cpp
MichalPetryka: Also, you've increased the size of the struct by 2 pointers, not one.

EvanBalster: The internal struct is increased by two pointers (callbacks assigned to input and output stream).  I've only added one function pointer to PaAsioStreamInfo:  https://github.com/PortAudio/portaudio/pull/519/files#diff-d0136a1e498ae2f8b83825ed61b878a90f8904151f365099b498358d1937d863R218

MichalPetryka: Nvm, I've looked at PaAsioStream not PaAsioStreamInfo.

src/hostapi/asio/pa_asio.cpp
dyfer: I believe that [Jack 1.9.17 on Windows](https://github.com/jackaudio/jack2/releases/tag/v1.9.17) provides a new, recompiled version of JackRouter ASIO driver, which I believe doesn't have this issue anymore. I can't find the relevant discussion thread though...

EvanBalster: Unfortunately older installations of JACK are still extant on users' machines.  Considering this problem creates an immediate crash in Pa_Initialize where it does occur, the driver should remain blacklisted for a while (unless there's a way to detect its version number here, which might be out of scope).

npostavs:
I have a PR that overlaps somewhat with this, so we should probably coordinate that (#473). The difference is that I only cover `kAsioResetRequest` and `kAsioBufferSizeChange`, and the callback isn't tied to a stream. What happens with our code if the user changes the buffer size before the application is playing/recording sound?

RossBencina:
I still need to review this properly, but two obvious things:

1. This extension should be selected using a flag (same as other host API extensions) not (only) by inspecting the struct version
2.  it needs to be documented is that it is the client's problem to deal with synchronising notification callbacks that occur during Pa_CloseStream() to avoid race conditions.

include/pa_asio.h
```diff
@@ -125,7 +194,7 @@ PaError PaAsio_SetStreamSampleRate( PaStream* stream, double sampleRate );
 typedef struct PaAsioStreamInfo{
     unsigned long size;             /**< sizeof(PaAsioStreamInfo) */
     PaHostApiTypeId hostApiType;    /**< paASIO */
-    unsigned long version;          /**< 1 */
+    unsigned long version;          /**< 2 */
 
     unsigned long flags;
```
RossBencina, line 199: Use the flags field to indicate whether a feature in the struct is to be used. There is currently a flag called `paAsioUseChannelSelectors`. A new flag `paAsioUseMessageCallback` should be added, and that should be used rather than struct size+version, or NULL to indicate presence of the `messageCallback`. That's the way that we do it elsewhere and I think it's best to be consistent here.

src/hostapi/asio/pa_asio.cpp
```diff
@@ -1304,9 +1304,10 @@ PaError PaAsio_Initialize( PaUtilHostApiRepresentation **hostApi, PaHostApiIndex
                 continue;
             }
 
-            // Blacklist dynamic loading of JackRouter in 32-bit applications.
+            // Blacklist dynamic loading of JackRouter ASIO driver by 64-bit applications.
             // This causes an instant crash due to a long-standing address translation error.
             //   https://github.com/jackaudio/jack2/issues/332
+            //   https://github.com/jackaudio/jack2/issues/275
```
Be-ing, line 1310: This is tangential to this PR, but I got PortAudio to build on Windows with the JACK host API: https://github.com/PortAudio/portaudio/pull/618 Considering the ASIO backend explicitly prevents using JackRouter, I think this would be the only way for PortAudio applications to use JACK on Windows.

It builds but it doesn't seem to work on Windows. I have not had a chance to debug it yet. Help would be appreciated if you are interested.

src/hostapi/asio/pa_asio.cpp
Be-ing: This is tangential to this PR, but I got PortAudio to build on Windows with the JACK host API: https://github.com/PortAudio/portaudio/pull/618 Considering the ASIO backend explicitly prevents using JackRouter, I think this would be the only way for PortAudio applications to use JACK on Windows.

It builds but it doesn't seem to work on Windows. I have not had a chance to debug it yet. Help would be appreciated if you are interested.

EvanBalster:
I notice this was tagged for inclusion in the next release of PortAudio.

I'm currently working on ASIO functionality in my company's application and expect I will have an opportunity to revisit this Pull Request in the coming weeks.

dyfer:
If I understand correctly, this PR would prevent PortAudio apps from using JackRouter (routing audio through JACK) on Windows, is this correct? If so, such change would affect SuperCollider Windows users, if they use JACK. I understand that one can't make everyone happy, but I think this would be an unfortunate regression.

To clarify, the issue only occurs with older versions of JACK (pre-2021); it does not occur with the current version.

EvanBalster:
> If I understand correctly, this PR would prevent PortAudio apps from using JackRouter (routing audio through JACK) on Windows, is this correct? If so, such change would affect SuperCollider Windows users, if they use JACK. I understand that one can't make everyone happy, but I think this would be an unfortunate regression.
> 
> To clarify, the issue only occurs with older versions of JACK (pre-2021); it does not occur with the current version.
> 
> EDIT: maybe blocking JackRouter could be enabled/disabled at compile time with a CMake option? That way those building PortAudio themselves could make the appropriate decision as they see fit.

This PR blocks interfacing with JACK through its ASIO adaptor, "JackRouter", in order to avoid a problem that historically rendered the JACK ASIO adaptor nonfunctional for 64-bit apps.  PortAudio also supports interfacing directly with the JACK API on Windows (ie, natively rather than through ASIO/JackRouter) which will understandably be the better approach to making Windows applications with Jack support.

If the bug in JackRouter has been fixed, and that fix has propagated widely, the guard could be removed -- perhaps with a CMake option as you suggest.  But be advised that the JackRouter problem causes applications to crash when _initializing_ PortAudio on any system with the faulty ASIO driver; this usually means an app without this guard will be unopenable on affected systems.  The nature of the crash also makes inspection of the debug trace rather difficult.

dyfer:
I'm aware of the bug, and that it prevents the application from starting.

Is building PortAudio with the JACK backend on Windows actually operational? I don't think that was the case in the past. Also, would the application be then linked against the Jack library? IMO it's quite important to allow the same binary to run both on systems that do and don't have JACK installed.

If JACK on Windows works, AND if the JACK library would be dynamically loaded, then indeed it is a viable workaround.

As I wrote, the fix has been around for over two years now. I don't have the data on how well it is propagated. Since SuperCollider uses PortAudio on Windows, this has come up in SuperCollider's issue tracker, so it is likely that there are users who use PortAudio + Jack (with fixed JackRouter) on Windows.

src/hostapi/asio/pa_asio.cpp
```diff
@@ -1304,6 +1304,17 @@ PaError PaAsio_Initialize( PaUtilHostApiRepresentation **hostApi, PaHostApiIndex
                 continue;
             }
 
+            // Blacklist dynamic loading of JackRouter ASIO driver by 64-bit applications.
```
RossBencina, line 1307: Please move any changes unrelated to the purpose of this PR to a separate PR. In particular, blacklisting a device is unrelated to adding support for ASIO messages.

EvanBalster: I've removed the JackRouter blacklisting rule.  The issues relating to it on their GitHub remain closed but two years on it's my understanding that bug has long since been fixed and few computers will have the old broken version installed.

EvanBalster:
Pardon my two year (minus one day) delay.

- Removed the rule blacklisting the JackRouter virtual ASIO driver.
- Added `PaAsioUseMessageCallback` flag as requested by @RossBencina, and added validation code.

Note that the latter change will break compatibility with any existing applications that use my branch (the message callback will not be used because the flag is not set).

I considered checking for this in `ValidateAsioSpecificStreamInfo` but that would represent a departure from the way channelSelectors is validated (ie, non-null channelSelectors is simply ignored if the flag is not set).  This gives me the impression the library is meant to tolerate the case where unused fields in the structure are uninitialized rather than zero-initialized.

include/pa_asio.h
EvanBalster: Done.  Make note of the validation code.

src/hostapi/asio/pa_asio.cpp
```diff
@@ -1871,24 +1874,43 @@ static PaError ValidateAsioSpecificStreamInfo(
...
+        case 2:
+            if( streamInfo->size != sizeof( PaAsioStreamInfo ) )
+                return paIncompatibleHostApiSpecificStreamInfo;
+            break;
+            
+        default:
             return paIncompatibleHostApiSpecificStreamInfo;
         }
 
         if( streamInfo->flags & paAsioUseChannelSelectors )
```
RossBencina, line 1894: It looks like you've fixed a bug here by making all channel selector handling depend on the condition `streamInfo->flags & paAsioUseChannelSelectors`. This seems like a valid fix. Could you extract this out to a separate PR please? We can merge that immediately.

include/pa_asio.h
```diff
@@ -88,6 +88,75 @@ PaError PaAsio_ShowControlPanel( PaDeviceIndex device, void* systemSpecific );
...
+    /** Informs the application that the driver's latencies have changed.
+        FIXME: The only way to query the new latencies is to reset the stream.
+        Recommend ignoring unless latency reporting is critical.
+        
+        params:
+            none.
+    */
+    paAsioLatenciesChanged = 5,
+    
+} PaAsioMessageType;
```
RossBencina, line 141:
```suggestion
} PaAsioMessageType;

```

insert blank line

RossBencina:
Please fix the trailing whitespace issues. Maybe consider installing an editor plugin to strip trailing whitespace on save.

```
Run python ./pa_whitelint.py
error: src/hostapi/asio/pa_asio.cpp(1884) trailing whitespace:
b'            '
error: src/hostapi/asio/pa_asio.cpp(1889) trailing whitespace:
b'            '
error: src/hostapi/asio/pa_asio.cpp(3219) trailing whitespace:
b'        '
error: src/hostapi/asio/pa_asio.cpp(3221) trailing whitespace:
b'    '
error: src/hostapi/asio/pa_asio.cpp(3231) trailing whitespace:
b'    '
error: src/hostapi/asio/pa_asio.cpp(3249) trailing whitespace:
b'    '
error: src/hostapi/asio/pa_asio.cpp(3256) trailing whitespace:
b'    '
error: src/hostapi/asio/pa_asio.cpp(3358) trailing whitespace:
b'    '
error: src/hostapi/asio/pa_asio.cpp(3361) trailing whitespace:
b'    '
error: include/pa_asio.h(92) trailing whitespace:
b'    '
error: include/pa_asio.h(102) trailing whitespace:
b'            '
error: include/pa_asio.h(107) trailing whitespace:
b'    '
error: include/pa_asio.h([11](https://github.com/PortAudio/portaudio/actions/runs/8013040990/job/21926914230?pr=519#step:4:12)0) trailing whitespace:
b'        '
error: include/pa_asio.h(115) trailing whitespace:
b'    '
error: include/pa_asio.h(118) trailing whitespace:
b'        '
error: include/pa_asio.h([12](https://github.com/PortAudio/portaudio/actions/runs/8013040990/job/21926914230?pr=519#step:4:13)3) trailing whitespace:
b'    '
error: include/pa_asio.h(126) trailing whitespace:
b'        '
error: include/pa_asio.h([13](https://github.com/PortAudio/portaudio/actions/runs/8013040990/job/21926914230?pr=519#step:4:14)0) trailing whitespace:
b'    paAsioResyncRequest     = 4, '
error: include/pa_asio.h(131) trailing whitespace:
b'    '
error: include/pa_asio.h(135) trailing whitespace:
b'        '
error: include/pa_asio.h([14](https://github.com/PortAudio/portaudio/actions/runs/8013040990/job/21926914230?pr=519#step:4:15)0) trailing whitespace:
b'    '
error: include/pa_asio.h(144) trailing whitespace:
b' '
error: include/pa_asio.h(147) trailing whitespace:
b' '
error: include/pa_asio.h([15](https://github.com/PortAudio/portaudio/actions/runs/8013040990/job/21926914230?pr=519#step:4:16)0) trailing whitespace:
b' '
error: include/pa_asio.h(1[53](https://github.com/PortAudio/portaudio/actions/runs/8013040990/job/21926914230?pr=519#step:4:54)) trailing whitespace:
b' '
error: include/pa_asio.h(1[56](https://github.com/PortAudio/portaudio/actions/runs/8013040990/job/21926914230?pr=519#step:4:57)) trailing whitespace:
b'    '
```

src/hostapi/asio/pa_asio.cpp
EvanBalster: I did not change the behavior of the paAsioUseChannelSelectors flag.  That condition was there before.

EvanBalster:
I went ahead and implemented the forward-compatibility as it seems unlikely to cause problems compared to the alternative.  Implementors may need a fallback to deal with the possibility of an older portaudio DLL that rejects non-1 versions though.  https://github.com/PortAudio/portaudio/pull/519/commits/0308b1e714055104faad16658751f896d117b3c3

src/hostapi/asio/pa_asio.cpp
```diff
@@ -1871,24 +1874,50 @@ static PaError ValidateAsioSpecificStreamInfo(
...
-                return paInvalidChannelCount;
+            for( int i=0; i < streamParameters->channelCount; ++i ){
+                if( (*channelSelectors)[i] < 0
+                        || (*channelSelectors)[i] >= deviceChannelCount ){
+                    return paInvalidChannelCount;
+                }
             }
         }
+
+        if( streamInfo->flags && paAsioUseMessageCallback )
```
nidefawl, line 1916: Shouldn't this be a bitwise & ?

EvanBalster: Well that's an embarrassing oversight on my part!  Fixed.

src/hostapi/asio/pa_asio.cpp
RossBencina: You did change the behavior: you added braces on lines 1902-1920 so the scope of the conditional is changed.

src/hostapi/asio/pa_asio.cpp
```diff
@@ -1871,24 +1874,50 @@ static PaError ValidateAsioSpecificStreamInfo(
...
+                return paIncompatibleHostApiSpecificStreamInfo;
+            break;
+
+        case 2:
+            if( streamInfo->size != sizeof( PaAsioStreamInfo ) )
+                return paIncompatibleHostApiSpecificStreamInfo;
+            break;
+
+        default:
+            /* Newer versions of the struct must be larger. */
```
RossBencina, line 1895:
```suggestion
            /* Newer versions of the struct must not be smaller. */
```

It's ok if the sizes are equal, and note that this can happen even if you add a field to a struct, depending on the alignof of the struct.

src/hostapi/asio/pa_asio.cpp
```diff
@@ -1871,24 +1874,50 @@ static PaError ValidateAsioSpecificStreamInfo(
...
-                || streamInfo->version != 1 )
+        switch( streamInfo->version )
         {
+        case 0:
+            /* NOTE: Struct version 0 is invalid, possibly indicative of uninitialized data. */
             return paIncompatibleHostApiSpecificStreamInfo;
+
+        case 1:
+            /* NOTE: V1 structure's size is smaller by one pointer. */
+            if( streamInfo->size < sizeof( PaAsioStreamInfo ) - sizeof(PaAsio_MessageCallback*) )
```
RossBencina, line 1885: Structly speaking, `<` is incorrect. If when defining the v2 struct, the v1 struct has to be padded to append an aligned `PaAsio_MessageCallback*` then the v1 struct will be less than `sizeof( PaAsioStreamInfo ) - sizeof(PaAsio_MessageCallback*)`. In practice this is probably not an issue since `alignof(v1struct)` is probably equal to `alignof(PaAsio_MessageCallback*)`.

philburk:
The CI is still finding some trailing white space.

src/hostapi/asio/pa_asio.cpp
EvanBalster: Alternatives would be to declare a V1 version of the structure, formulate a more elaborate size test based on normative compiler behavior, or not test at all.  The previous field is a pointer too.  So I'm inclined to leave this imperfect heuristic as it is.

src/hostapi/asio/pa_asio.cpp
EvanBalster: OK, I've created PR https://github.com/PortAudio/portaudio/pull/885 comprising just this fix and removed it from here.  We'll want to merge 885 first.

EvanBalster:
At [Ross' suggestion](https://github.com/PortAudio/portaudio/pull/519#discussion_r1501241157) I've split off #885 which fixes a bug in stream info validation.  885 should be merged before this PR; otherwise it won't be possible to use ASIO messages without also specifying a channel mapping.

src/hostapi/asio/pa_asio.cpp
RossBencina: I think we all agree that it is imperfect but we'll live with it for now.

src/hostapi/asio/pa_asio.cpp
```diff
@@ -2742,6 +2769,21 @@ static PaError OpenStream( struct PaUtilHostApiRepresentation *hostApi,
...
+    stream->messageCallback[0] = NULL;
+    stream->messageCallback[1] = NULL;
+    if( usingBlockingIo )
+    {
+        /* Disable message callback due to policy of overwriting userData above. */
+    }
+    else
+    {
+        /* Message callback may be defined in versions >= 2 of ASIO streamInfo. */
+        if(  inputStreamInfo && ( inputStreamInfo->flags & paAsioUseMessageCallback) )
```
RossBencina, line 2781:
```suggestion
        if( inputStreamInfo && ( inputStreamInfo->flags & paAsioUseMessageCallback) )
```

whitespace tweak

RossBencina:
#885 is merged now

src/hostapi/asio/pa_asio.cpp
```diff
@@ -3186,28 +3248,52 @@ static void sampleRateChanged(ASIOSampleRate sRate)
...
     long ret = 0;
 
     (void) message; /* unused parameters */
     (void) opt;
 
     PA_DEBUG( ("asioMessages : %d , %d \n", selector, value));
 
+    if (!theAsioStream)
+    {
+        /* Some ASIO drivers will fire messages during OpenStream.  We ignore them. */
```
RossBencina, line 3279: This seems like incomplete code (same in the  `sampleRateChanged` handler). Is there a rationale for dropping these messages?

To me, it seems that if we're going to pass through driver messages, we should pass through all of them. All it would take would be to stash the required callback data for use here even before the stream is fully created.

EvanBalster: It has been quite a while since I visited this code.  My faint recollection is that some ASIO drivers were generating spurious ResetRequest messages at this time but I'd need to look into it.

philburk:
According to Gemini, this cryptic command might trigger a CI test.

$ bors r+

philburk:
My command did not trigger a CI test. I think this PR needs to be rebased on master.

RossBencina:
Discussion invited:

This PR has stalled because there is a policy decision to resolve about what to do with ASIO driver callbacks while the stream is being opened. I don't like the idea of suppressing messages in a low-level pass-through API -- thatshould be a decision for higher layers to make.

The current PR code suppresses ASIO message callbacks "while the stream is being opened," in-fact only up to the sequence point in `pa_asio.cpp` `OpenStream()` that sets `theAsioStream = stream;` [1]. Note that `pa_front.c` `Pa_OpenStream` does more work after that before returning to the client. So the currently implemented mechanism may still result in callbacks being made prior to when `Pa_OpenStream` returns. This is unavoidable and consistent with other asynchronous callbacks in PortAudio.

If I understand correctly, the purpose of suppressing callbacks is to avoid issues with the client getting stuck in a loop of infinite resets while the stream is being opened.  On the other hand, if a legitimate reset happens after `LoadAsioDriver` but before  `Pa_OpenStream` returns the current PR code will suppress that event. This is undesirable.

Baseline proposal: Don't suppress any callbacks. The client will have to do the work to ignore callbacks up to some sequence point after Pa_OpenStream has returned, either using a flag or a timer. This has effectively the same behavior as I described in the previous paragraph but gives the client full control over exactly when the stop-ignoring sequence point happens. Document that the notification callback can occur prior to` Pa_OpenStream` returning and client's should avoid getting into infinite reset loops.

Enhancement: add a flags parameter to the message notification callback. Define at least two flag fields: `paAsioLoadingDriver`, `paAsioConfiguringDriver` . The first is set for notifications that occur prior to `LoadAsioDriver` returning. The second is set prior to `OpenStream()` returning.  This allows the client to distinguish between "reset while opening ASIO driver" callbacks and "reset after opening" callbacks, which greatly reduces or eliminates race conditions.

[1] NB assignment should be atomic/have release memory barrier if this mechanism is used

EvanBalster:
I think the flags parameter would probably be the best way to do it.  Or adding a couple new event codes corresponding to the transitions between those stages of initialization.

Apologies for leaving this untouched for so long; I was in poor health throughout 2023-2024.

RossBencina:
@EvanBalster sorry to hear about your health issues. Phil and I are discussing moving this PR off the 19.8 release milestone. I think there's still time to ship it if you (or someone else) would like to work on it now, but there's no pressure, we can just take care of it later.

EvanBalster:
Those health issues (post-COVID syndrome) made an unexpected comeback so I can't commit to anything short term.

RossBencina:
@EvanBalster no problems. I wish you the best for your recovery.

Unless someone else is interested in working on this PR we will move it to v19.9

Issue #534, opened 2021-03-25
JACK regex escape fix has excess/unnecessary string allocation and copies


RossBencina:
The following code improvement was overlooked in PR #405 which has now been committed.

The device string is copied and escaped into `device_name_regex_escaped` and then that string is immediately copied into a new string `port_regex_string`.

Near identical code is used in three locations:

https://github.com/PortAudio/portaudio/blob/eec7bb739771381ab698a6a611d933bdd72cbd8f/src/hostapi/jack/pa_jack.c#L628

https://github.com/PortAudio/portaudio/blob/eec7bb739771381ab698a6a611d933bdd72cbd8f/src/hostapi/jack/pa_jack.c#L1296

https://github.com/PortAudio/portaudio/blob/eec7bb739771381ab698a6a611d933bdd72cbd8f/src/hostapi/jack/pa_jack.c#L1324

Operations could be performed directly to `port_regex_string`. This would avoid allocation the intermediate `device_name_regex_escaped` buffer, and avoid the extra string copy.

The fix is to remove all mention of `device_name_regex_escaped` and perform string operations directly in  `port_regex_string`.

RossBencina:
@Be-ing what do you think about this? Would you like to include a fix in 19.7? Phil and I have no way to test a fix.

Be-ing:
I could test but I will be quite busy the next couple days so I don't know if I could write the code.

RossBencina:
I've proposed PR: #536

philburk:
Because this is only an optimization, let's punt this to 19.8.

Issue #405, opened 2020-12-29
python3: src/hostapi/alsa/pa_linux_alsa.c:3641: PaAlsaStreamComponent_BeginPolling: Assertion `ret == self->nfds' failed.


antiochtech:
rpi 3b, raspbian buster

error occurs within a minute or two.

doesnt matter if im installing pyaudio and portaudio from packages or building them myself. ive tried the last stable release of pa, as well as the nightly build.

https://pastebin.com/id2QaedQ

[Edit: Reproducing example at above link is
```
#!/usr/bin/env python3
 
import sys
import pyaudio
import time
import RPi.GPIO as GPIO
from array import array
 
GPIO.setmode(GPIO.BCM)
GPIO.setwarnings(False)
GPIO.setup(24, GPIO.OUT)
 
THRESHOLD = 327
MAX_SILENCE = 62
CHUNK_SIZE = 1024
FORMAT = pyaudio.paInt16
RATE = 44100
 
def is_silent(snd_data):
    "Returns 'True' if below the 'silent' threshold"
    return max(snd_data) < THRESHOLD
 
def listen():
    num_silent = 0
    snd_started = False
 
    while True:
        # little endian, signed short
        snd_data = array('h', stream.read(CHUNK_SIZE))
        if sys.byteorder == 'big':
            snd_data.byteswap()
 
        silent = is_silent(snd_data)
 
        if silent:
            if snd_started:
                num_silent += 1
                GPIO.output(24,0)
                sys.stdout.write("data: yellow\n\n")
            elif not snd_started:
                GPIO.output(24,1)
                sys.stdout.write("data: red\n\n")
            sys.stdout.flush()
        elif not silent:
            if snd_started:
                num_silent = 0
            elif not snd_started:
                snd_started = True
            GPIO.output(24,0)
            sys.stdout.write("data: green\n\n")
            sys.stdout.flush()
 
        if snd_started and num_silent > MAX_SILENCE:
            snd_started = False
            num_silent = 0
            sys.stdout.write("data: red\n\n")
            sys.stdout.flush()
            GPIO.output(24,0)
 
        if snd_started and silent and num_silent < MAX_SILENCE:
            GPIO.output(24,1)
 
if __name__ == '__main__':
    p = pyaudio.PyAudio()
    stream = p.open(format=FORMAT, channels=1, rate=RATE,
    input=True, output=True,
    frames_per_buffer=CHUNK_SIZE)
[...]

PR #536, opened 2021-03-29
Improve JACK regex pattern escape code (see ticket #534). 

RossBencina:
Improves previously merged PR #405 master commit eec7bb7 by performing string conversion directly in to the target buffer, thus avoiding one allocation and one strncpy per client/device name.

NOTE: I have not tested this change.


RossBencina:
@Be-ing Please could you review and test this?

Be-ing (approved):
I tested and this still works.

philburk (approved):

PR #546, opened 2021-04-12
Export "PaUtil_*" symbols from shared library (fixes #423)

### LINKED ISSUE #423: application level debug output
I am having difficulty assigning a callback function for portaudio debug strings:
  CXX      soundcard/fldigi-sound.o
  CXXLD    fldigi
/usr/bin/ld: soundcard/fldigi-sound.o: in function `SoundPort::initialize()':
/home/dave/SF/fldigi.git/src/soundcard/sound.cxx:1160: undefined reference to `PaUtil_SetDebugPrintFunction(void (*)(char const*))'

Any help at all would be much appreciated.

Thank you.

David

mmuetzel:
"PaUtil_SetDebugPrintFunction()" and related functions aren't exported from the shared library.
The proposed change should allow programs that link to the shared library to use that and related functions.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -44,7 +44,7 @@ PALIB = libportaudio.la
 PAINC = include/portaudio.h
 
 PA_LDFLAGS = $(LDFLAGS) $(SHARED_FLAGS) -rpath $(libdir) -no-undefined \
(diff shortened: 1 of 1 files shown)


RossBencina:
PaUtil_ is a private API, we can't export all of it.

philburk (commented):
Thanks for the PR. It is clear now what the issue is.
We do not want to expose the PaUtil_ functions.
But we do need to provide some way to access the debug logs.
So we will consider adding something official to the API.

PR #559, opened 2021-04-29
Have PaUtil_GetTime() use monotonic time on Unix

xyzzy42:
The currently used clock, CLOCK_REALTIME, will have discontinuous jumps, forward
and backward, when the system clock is adjusted.  This is bad for synchronizing
audio.  CLOCK_MONOTONIC does not have jumps, and never moves backward, but does
have the continuous adjustments done by NTP to increase clock accuracy.

Since kernel 2.6.25 ALSA has use CLOCK_MONOTONIC by default for timestamps
(commit b751eef1 from Dec 13 2007).  This change means the timestamps provided
in the PA callbacks can be compared to PaUtil_GetTime(), which currently does
not work.

The function used on MacOS, mach_absolute_time(), behaves like CLOCK_MONOTONIC.

If the system does not have CLOCK_MONOTONIC, fallback to CLOCK_REALTIME, which
is what alsa-lib does for such systems.  Which at this point would need to be
very old.

### DIFF
--- a/src/os/unix/pa_unix_util.c
+++ b/src/os/unix/pa_unix_util.c
@@ -160,7 +160,11 @@ PaTime PaUtil_GetTime( void )
     return mach_absolute_time() * machSecondsConversionScaler_;
 #elif defined(HAVE_CLOCK_GETTIME)
     struct timespec tp;
+#if defined(CLOCK_MONOTONIC)
+    clock_gettime(CLOCK_MONOTONIC, &tp);
+#else
     clock_gettime(CLOCK_REALTIME, &tp);
+#endif
     return (PaTime)(tp.tv_sec + tp.tv_nsec * 1e-9);
 #else
     struct timeval tv;
--- a/src/os/unix/pa_unix_util.c
+++ b/src/os/unix/pa_unix_util.c
@@ -280,7 +280,7 @@ PaError PaUnixThread_New( PaUnixThread* self, void* (*threadFunc)( void* ), void
     memset( self, 0, sizeof (PaUnixThread) );
     PaUnixMutex_Initialize( &self->mtx );
     PA_ASSERT_CALL( pthread_condattr_init( &cattr ), 0 );
-#if defined(CLOCK_MONOTONIC)
+#if defined(CLOCK_MONOTONIC) && !defined(__APPLE__)
     PA_ASSERT_CALL( pthread_condattr_setclock( &cattr, CLOCK_MONOTONIC ), 0 );
 #endif
     PA_ASSERT_CALL( pthread_cond_init( &self->cond, &cattr), 0 );
--- a/src/os/unix/pa_unix_util.c
+++ b/src/os/unix/pa_unix_util.c
@@ -274,11 +274,16 @@ PaError PaUnixThread_New( PaUnixThread* self, void* (*threadFunc)( void* ), void
 {
     PaError result = paNoError;
     pthread_attr_t attr;
+    pthread_condattr_t cattr;
     int started = 0;
 
     memset( self, 0, sizeof (PaUnixThread) );
     PaUnixMutex_Initialize( &self->mtx );
-    PA_ASSERT_CALL( pthread_cond_init( &self->cond, NULL ), 0 );
+    PA_ASSERT_CALL( pthread_condattr_init( &cattr ), 0 );
+#if defined(CLOCK_MONOTONIC)
+    PA_ASSERT_CALL( pthread_condattr_setclock( &cattr, CLOCK_MONOTONIC ), 0 );
+#endif
+    PA_ASSERT_CALL( pthread_cond_init( &self->cond, &cattr), 0 );
 
     self->parentWaiting = 0 != waitForChild;
 


dmitrykos (approved):

dmitrykos:
For sure, `CLOCK_REALTIME` can be affected by `clock_settime` (https://www.gnu.org/software/libc/manual/html_node/Setting-and-Adjusting-the-Time.html) and the proposed change will avoid broken timestamps if clock is adjusted.

RossBencina:
Using CLOCK_MONOTONIC makes sense to me.

I would like to see some discussion of how this change impacts the current PortAudio ALSA implementation.

`PaUtil_GetTime` seems mostly used for debug timestamps, however it appears that in `pa_linux_alsa.c` lines 3279 and  3299, the timestamp `now = PaUtil_GetTime` is compared directly to the result of `alsa_snd_pcm_status_get_trigger_tstamp()`. What kind of timestamp does `alsa_snd_pcm_status_get_trigger_tstamp()` return? I assume CLOCK_MONOTONIC. Does this mean that the PA/ALSA code is currently buggy because it is comparing CLOCK_REALTIME with CLOCK_MONOTONIC?

xyzzy42:
I found in bug due to my change elsewhere, when `PaUtil_GetTime()` is used with `pthread_cond_timedwait()`.  I have a fix.

For the code in `PaAlsaStream_HandleXrun()`, I think the answer is yes, it's wrong now.  Here is an example, with this PR, after triggering a overrun when I added some debug statements:
```
overrun: now is 1288573, trigger ts is 1288571
```
This is monotonic time (about two weeks uptime in seconds), not the UNIX epoch time that CLOCK_REALTIME would return.  If we used that, as in the current code without this PR, we get something like:
```
overrun: now is 1620263316, trigger ts is 1288744
```
Clearly wrong.

The timestamp in the audio callback, while not compared to PaUtil_GetTime() in the PA codebase, likely could be in user code and has this same issue.

ALSA at first used CLOCK_REALTIME for timestamps, but this was not a good idea as I think we all know, and they switched to being configurable with a CLOCK_MONOTONIC default back in 2007.  I identified the commit when this happened in my commit message.  Maybe the PA code predates this change being in common use?

RossBencina:
I have a few comments:
 * For the record, I am 100% ok with changing `PaUtil_GetTime` to use CLOCK_MONOTONIC
 * The requirement for this PR is to not break things or make things worse, it doesn't have to fix additional issues that I mention below.
 * Personal opinion: I don't think that `pa_linux_alsa.c` should mix `PaUtil_GetTime` with ALSA APIs such as the aforementioned comparison with `alsa_snd_pcm_status_get_trigger_tstamp()`. For something like that, the code should directly retrieve the clock by calling the appropriate native API directly.
 * As a rule, values returned by `PaUtil_GetTime` should only be compared with other values returned by `PaUtil_GetTime`, otherwise the code is too brittle with the potential for breakage as we have seen here. At the very least, it should not be mixed with APIs at a lower level of abstraction.
 * `PaUtil_GetTime` should never be called directly by clients, it is a PortAudio private function, not part of the public API. So there should be no reason to worry about this change breaking clients.

RossBencina:
This link provides additional details including reasons not to use CLOCK_MONOTONIC_RAW (a comment claims that it is not in the VDSO so has additional syscall overhead).

https://stackoverflow.com/questions/3523442/difference-between-clock-realtime-and-clock-monotonic

xyzzy42:
> This link provides additional details including reasons not to use CLOCK_MONOTONIC_RAW (a comment claims that it is not in the VDSO so has additional syscall overhead).
> 
> https://stackoverflow.com/questions/3523442/difference-between-clock-realtime-and-clock-monotonic

I think this is out of date.  The current VDSO appears to support MONOTONIC_RAW.

https://elixir.bootlin.com/linux/latest/source/lib/vdso/gettimeofday.c#L243

But CLOCK_MONOTONIC is intended to be more accurate than CLOCK_MONOTONIC_RAW.

xyzzy42:
> * `PaUtil_GetTime` should never be called directly by clients, it is a PortAudio private function, not part of the public API. So there should be no reason to worry about this change breaking clients.

I was linking against a static library build of PortAudio, to test my changes for this, HR timestamps, and some other things, so calling PaUtil_GetTime() worked.  It is not part of the dynamic symbol table of the shared library and can't be called when using that.

xyzzy42:
The break on MacOS is because it can't set a pthread condattr clock, and so they are stuck with CLOCK_REALTIME on MacOS.  But since MacOS already uses the equivalent of CLOCK_MONOTONIC for PaUtil_GetTime(), how does it work now?

The answer is it does not work now.

The code using PaUtil_GetTime() to set a condition variable timeout,  `PaUnixThread_New()`, is in pa_unix_util.c, but the only users of this function are two Linux host APIs, ALSA and asihpi.  It's not used on MacOS.

Would it be better to not build this function on MacOS, since it is unused and does not work?  Or to fix it so that it would work if it were used?  "[PaUtil_GetTime] should not be mixed with APIs at a lower level of abstraction," seems an apt observation, as every instance identified so far appears broken on at least one platform.

src/os/unix/pa_unix_util.c
```diff
@@ -270,11 +274,16 @@ PaError PaUnixThread_New( PaUnixThread* self, void* (*threadFunc)( void* ), void
...
     PaError result = paNoError;
     pthread_attr_t attr;
+    pthread_condattr_t cattr;
     int started = 0;
 
     memset( self, 0, sizeof (PaUnixThread) );
     PaUnixMutex_Initialize( &self->mtx );
-    PA_ASSERT_CALL( pthread_cond_init( &self->cond, NULL ), 0 );
+    PA_ASSERT_CALL( pthread_condattr_init( &cattr ), 0 );
+#if defined(CLOCK_MONOTONIC)
```
philburk, line 283: This allows it to compile on MacOS. But it is probably not the best solution.

```suggestion
#if defined(CLOCK_MONOTONIC) && !defined(__APPLE__)
```

Maybe for Mac, if we cannot implement this function properly then we should assert at runtime.

philburk (changes requested):

RossBencina:
> The code using PaUtil_GetTime() to set a condition variable timeout

It would be better if `PaUtil_GetTime()` was not used for that. I think we need to scale back `PaUtil_GetTime()` usage, not weave it more tightly into more code.

xyzzy42:
> > The code using PaUtil_GetTime() to set a condition variable timeout
> 
> It would be better if `PaUtil_GetTime()` was not used for that. I think we need to scale back `PaUtil_GetTime()` usage, not weave it more tightly into more code.

It's using that now, so it's not more tightly, it's equally tightly.

src/os/unix/pa_unix_util.c
xyzzy42: There is some code in chromium's os library for this.  On POSIX, they use a monotonic clock + pthread_cond_timedwait, like I've changed this to use.  On MacOS, which doesn't have that ability, they use `pthread_cond_reltimedwait_np()`.  So I could change to do that on macos.  But this code isn't used on macos.  It seems a waste to add code that's not used.

philburk (approved):

philburk:
I fixed the build for Mac.
We should solve the Mac issue by separate out a pa_mac_util.c file.  #596

PR #560, opened 2021-04-29
Alsa: Use hires timestamps

xyzzy42:
ALSA added nanosecond timestamps in library version 0.9.1.  Use these
instead of microsecond timestamps for reporting audio timestamps as well
as underrun/overrun times.

alsa-lib version 0.9.1 was released on Mar 11 2003.  I had fallback code
to use non-hires timestamps, but after 18 years it didn't seem
necessary, and I removed it from this commit.

### DIFF
--- a/src/hostapi/alsa/pa_linux_alsa.c
+++ b/src/hostapi/alsa/pa_linux_alsa.c
@@ -220,8 +220,10 @@ _PA_DEFINE_FUNC(snd_config_update_free_global);
 _PA_DEFINE_FUNC(snd_pcm_status);
 _PA_DEFINE_FUNC(snd_pcm_status_sizeof);
 _PA_DEFINE_FUNC(snd_pcm_status_get_tstamp);
+_PA_DEFINE_FUNC(snd_pcm_status_get_htstamp);
 _PA_DEFINE_FUNC(snd_pcm_status_get_state);
 _PA_DEFINE_FUNC(snd_pcm_status_get_trigger_tstamp);
+_PA_DEFINE_FUNC(snd_pcm_status_get_trigger_htstamp);
 _PA_DEFINE_FUNC(snd_pcm_status_get_delay);
 #define alsa_snd_pcm_status_alloca(ptr) __alsa_snd_alloca(ptr, snd_pcm_status)
 
@@ -499,8 +501,10 @@ static int PaAlsa_LoadLibrary()
     _PA_LOAD_FUNC(snd_pcm_status);
     _PA_LOAD_FUNC(snd_pcm_status_sizeof);
     _PA_LOAD_FUNC(snd_pcm_status_get_tstamp);
+    _PA_LOAD_FUNC(snd_pcm_status_get_htstamp);
     _PA_LOAD_FUNC(snd_pcm_status_get_state);
     _PA_LOAD_FUNC(snd_pcm_status_get_trigger_tstamp);
+    _PA_LOAD_FUNC(snd_pcm_status_get_trigger_htstamp);
     _PA_LOAD_FUNC(snd_pcm_status_get_delay);
 
     _PA_LOAD_FUNC(snd_card_next);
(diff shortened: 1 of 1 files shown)


src/hostapi/alsa/pa_linux_alsa.c
```diff
@@ -3156,6 +3160,28 @@ static PaError IsStreamActive( PaStream *s )
...
+    }
+    else
+    {
+        alsa_snd_pcm_status_get_htstamp( status, &timestamp );
+    }
+    if ( delay )
+    {
+        *delay = alsa_snd_pcm_status_get_delay( status );
+    }
+    return timestamp.tv_sec + ( (PaTime)timestamp.tv_nsec / 1000000000.0 );
```
RossBencina, line 3182: Better to multiply my reciprocal here.

philburk: Also please avoid long strings of zeros. Prefer:

```suggestion
    return timestamp.tv_sec + ( (PaTime)timestamp.tv_nsec * 1.0E-9 );
```

xyzzy42: Ok, changed.  It's not exactly the same math, since 1e-9 isn't an exact floating point number, while 1e9 is.  But I don't think any possible values of tv_nsec result in a different result in the end.

RossBencina: You raise a valid point about division potentially preserving more accuracy.  Using reciprocal is desirable for performance reasons, but I would prefer not to throw away usable accuracy.  'PaStreamCallbackTimeInfo' should retain all available accuracy within the limits of what is representable by PaTime (64-bit float).

philburk: I ran a test that measured the difference between (n/1e9) and (n*1e-9) for all possible n.
The maximum error was  1.11022e-16 when n =  500000001
That seems like an insignificant numeric error. I think the performance hit is more significant.

```
void findMaxDelta() {
    double maxDelta = 0.0;
    uint xAtMax = -1;
    for (uint x = 1; x < 1e9; x++) {
        double y1 = x / 1.0e9;
        double y2 = x * 1.0e-9;
        double yd = fabs(y1 - y2);
        if (yd > maxDelta) {
            maxDelta = yd;
            xAtMax = x;
            std::cout << x << ", " << y1 << ", " << y2 << ", " << yd << "\n";
        }
    }
    std::cout << "xAtMax  = " << xAtMax << ", maxDelta = " << maxDelta << "\n";
}
```

xyzzy42: And since the timestamp is only precise to the ns to begin with, once you round that result with max error to nearest nanosecond, it becomes the exact same value as it would have had with no error.  That's what I meant when I said, "I don't think any possible values of tv_nsec result in a different result in the end."

philburk (approved):

src/hostapi/alsa/pa_linux_alsa.c
RossBencina: 1.11022e-16 seems insignificant to me. It is a small fraction of a nanosecond, and float64 only maintains about 13 decimal digits of precision, which by the time `timestamp.tv_sec` is added will cause quite a few digits to be lost. There are 31536000 seconds in a year so after a year the final result of `timestamp.tv_sec + ( (PaTime)timestamp.tv_nsec * 1e-9 )` will only include about 6-7 significant digits of the nanosecond part.

philburk:
I will test this.

RossBencina (approved):

philburk:
I tried to test this. I tried the old code first and it seems Pa_GetStreamTime() is broken and untested!

See #586

xyzzy42:
It works if you use an ALSA device and not the pulseaudio alsa plugin.

philburk:
> It works if you use an ALSA device and not the pulseaudio alsa plugin.

Right. I hacked the deviceId in patest_sine_time and it worked fine.

Issue #548, opened 2021-04-15
Move C++ bindings to a new repo


RossBencina:
We're planning to move the C++ bindings to their own repo.

Be-ing:
I started working on this in #434. I rebased those commits on master on [this branch](https://github.com/Be-ing/portaudio/commits/cpp_cmake), which could be used to resume work on this after new repository is created.

RossBencina:
Making P2 since other work appears to be depending on it.

RossBencina:
Status is that it's P2 so it is likely to get done. There is no assignee. Would you like to work on it?

RossBencina:
Phil and I are considering whether we should let someone take over maintenance of this binding outside the official PortAudio project structure. There's also the idea that a better/more modern-C++ binding would be nice. But we're not going to do that.

Febbe:
> take over maintenance of this binding outside the official PortAudio project structure.

That's going to be hard. I can't take it over, but I can help.

> There's also the idea that a better/more modern-C++ binding would be nice.

What do you think, what has to be modernized?

> But we're not going to do that.

Because you don't have time / resources, or you don't want to break the current API?

PR #561, opened 2021-04-29
Remove redundant build systems in favor of CMake + update build documentation

Be-ing:
This branch is based on #461. I will rebase it after that is merged.

Maintaining redundant build systems is a pointless waste of effort that almost inevitably leads to inconsistencies and frustrated users. CMake can generate MSVC project files and Unix Makefiles. There is no need to keep separate build systems for those.


xyzzy42:
There is a something that autotools can do that cmake doesn't: work with other build systems.

To use PortAudio in another app, autotools make pkgconfig .pc files that describe library and include paths, lib names, etc.  Autotools, cmake, meson, anything else, can and does use these.

cmake has a system to do this too, but it uses complex full cmake language files that must be included.  Nothing besides cmake can use it, since one would need to re-implement all of cmake to interpret them.

Be-ing:
CMake makes a pkgconfig file already.

dmitrykos:
The idea to stick to CMake only looks fine to me. For example currently building with MSVC from '/build/msvc' fails for me due to hard linked ASIO dependency, removing ASIO define leads to another failure due to missing symbols. It is what CMake can fix easily and make code portable and easily maintainable.

xyzzy42:
> CMake makes a pkgconfig file already.

Ok, I see the cmake build does that now, so it can work with other build systems.

Note that .pc file should have "Requires: alsa" when built with ALSA support.  This is something the Fedora build of portaudio patches.

RossBencina:
One obvious argument against CMake being the only build system is that it introduces CMake as a dependency. We have working `./configure'. I think we should retain it.

RossBencina:
@dmitrykos: MSVC is building fine in CI, it uses a script to remove the ASIO stuff. But ultimately it is legacy, and intended to build *with* ASIO.

Be-ing:
> One obvious argument against CMake being the only build system is that it introduces CMake as a dependency

I don't understand how that's a problem.

dmitrykos:
@RossBencina taking into account that Visual Studio has many versions it does not make sense to maintain a default project configuration for some older version when CMake is able to target any version. So basically maintaining a single CMake source we can support multiple versions of different IDEs without going in-depth how those IDEs operate and it will be a very robust and portable approach.

CMake currently is de-facto a major build system which is supported natively by Linux distributions, recent Visual Studio versions  has built-in optional support for CMake, so such dependency is not weak and ok to my view and it simplifies many things related to project configuration.

dechamps:
As someone who [tried to do something similar](https://app.assembla.com/spaces/portaudio/git/merge_requests/7040161) in December 2018 but was overruled, it goes without saying that I strongly support the spirit of this pull request.

Be-ing:
@dmitrykos @dechamps I would appreciate your review on #461 which this branch builds on.

Be-ing:
C++ bindings should be moved to a new repository (#548) before merging this because the C++ bindings depend on the autotools or SCons build system of the C library.

RossBencina:
Closing as obsolete. Please feel free to submit a PR to remove SCons build support. The other build systems should be retained for now.

Be-ing:
At the very least the files in `build` should be moved to another directory. It's really annoying that fish's autocompleted commands for cmake don't work for PortAudio because there is a `build` directory in the repo.

PR #567, opened 2021-05-08
ALSA: fix uninitialized stream pointer in pa_linux_alsa.c

MonkeybreadSoftware:
Avoid compiler warning about uninitialized pointer being used by assigning NULL explicitly and testing for it.

../PortAudioLib/portaudio/src/hostapi/alsa/pa_linux_alsa.c: In function 'PaAlsa_GetStreamInputCard':
../PortAudioLib/portaudio/src/hostapi/alsa/pa_linux_alsa.c:4629:16: warning: 'stream' may be used uninitialized in this function [-Wmaybe-uninitialized]
     PA_ENSURE( alsa_snd_pcm_info( stream->capture.pcm, pcmInfo ) );
                ^~~~~~~~~~~~~~~~~
../PortAudioLib/portaudio/src/hostapi/alsa/pa_linux_alsa.c: In function 'PaAlsa_GetStreamOutputCard':
../PortAudioLib/portaudio/src/hostapi/alsa/pa_linux_alsa.c:4648:16: warning: 'stream' may be used uninitialized in this function [-Wmaybe-uninitialized]
     PA_ENSURE( alsa_snd_pcm_info( stream->playback.pcm, pcmInfo ) );
                ^~~~~~~~~~~~~~~~~

### DIFF
--- a/src/hostapi/alsa/pa_linux_alsa.c
+++ b/src/hostapi/alsa/pa_linux_alsa.c
@@ -4616,11 +4616,13 @@ error:
 
 PaError PaAlsa_GetStreamInputCard( PaStream* s, int* card )
 {
-    PaAlsaStream *stream;
+    PaAlsaStream *stream = NULL;
     PaError result = paNoError;
     snd_pcm_info_t* pcmInfo;
 
     PA_ENSURE( GetAlsaStreamPointer( s, &stream ) );
+
+    if (stream == NULL) return paBadStreamPtr;
 
     /* XXX: More descriptive error? */
     PA_UNLESS( stream->capture.pcm, paDeviceUnavailable );
@@ -4635,11 +4637,13 @@ error:
 
 PaError PaAlsa_GetStreamOutputCard( PaStream* s, int* card )
 {
-    PaAlsaStream *stream;
+    PaAlsaStream *stream = NULL;
     PaError result = paNoError;
     snd_pcm_info_t* pcmInfo;
 
     PA_ENSURE( GetAlsaStreamPointer( s, &stream ) );
+
+    if (stream == NULL) return paBadStreamPtr;
 
     /* XXX: More descriptive error? */
     PA_UNLESS( stream->playback.pcm, paDeviceUnavailable );


src/hostapi/alsa/pa_linux_alsa.c
```diff
@@ -4616,12 +4616,14 @@ static PaError GetAlsaStreamPointer( PaStream* s, PaAlsaStream** stream )
...
 PaError PaAlsa_GetStreamInputCard( PaStream* s, int* card )
 {
-    PaAlsaStream *stream;
+    PaAlsaStream *stream = NULL;
     PaError result = paNoError;
     snd_pcm_info_t* pcmInfo;
 
     PA_ENSURE( GetAlsaStreamPointer( s, &stream ) );
 
+    if (stream == NULL) return paBadStreamPtr;
```
philburk, line 4625: The paBadStreamPtr error is mostly for when the caller passes in a NULL pointer.
This error is not the caller's fault so maybe use paUnanticipatedHostError instead.

philburk (changes requested):
Thanks for addressing these warnings. That's a big help.

src/hostapi/alsa/pa_linux_alsa.c
MonkeybreadSoftware: Feel free to change the error number.

philburk: The NULL check is redundant.
We should set stream=NULL right before the call because that is the PA pattern.

philburk (changes requested):

src/hostapi/alsa/pa_linux_alsa.c
MonkeybreadSoftware: The NULL check would catch a problem if GetAlsaStreamPointer would report OK, but not set the stream. Not sure whether this could happen.

RossBencina: @MonkeyBreadSoftware: if GetAlsaStreamPointer returns paNoError but does not return a valid stream pointer it is badly broken. That's not the kind of thing that we check at runtime.

src/hostapi/alsa/pa_linux_alsa.c
```diff
@@ -4616,12 +4616,14 @@ static PaError GetAlsaStreamPointer( PaStream* s, PaAlsaStream** stream )
 
 PaError PaAlsa_GetStreamInputCard( PaStream* s, int* card )
 {
-    PaAlsaStream *stream;
+    PaAlsaStream *stream = NULL;
```
RossBencina, line 4619: Please move the NULL initialization to just prior to the call site.

philburk:
Let's use #584 instead.

PR #570, opened 2021-05-11
DirectSound: fix recording in 24 bits

Be-ing:
fixes https://bugzilla.audacityteam.org/show_bug.cgi?id=193

I am upstreaming some old patches from Audacity: https://github.com/audacity/audacity/issues/840

### DIFF
--- a/src/hostapi/dsound/pa_win_ds.c
+++ b/src/hostapi/dsound/pa_win_ds.c
@@ -2039,6 +2039,9 @@ static PaError OpenStream( struct PaUtilHostApiRepresentation *hostApi,
         PaSampleFormat nativeInputFormats = paInt16;
         /* PaSampleFormat nativeFormats = paUInt8 | paInt16 | paInt24 | paInt32 | paFloat32; */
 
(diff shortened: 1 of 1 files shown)


dechamps:
The diff makes no sense to me (casting `userData` to `int` and checking if it contains `24`??). If you go back to the [original Audacity patch](https://github.com/audacity/audacity/commit/18eb2bbc7a6296f4f9c896d742b0c41382736abc) you will realize that there is code on the Audacity side as well, and `userData` is being used to do some kind of "secret handshake" between Audacity and their custom PortAudio code. Obviously this can't be merged into upstream PortAudio, and this PR should be abandoned.

Be-ing:
Yeah that's what @illucis [said](https://github.com/audacity/audacity/issues/840#issuecomment-837755387):

> Not sure this is appropriate (as written) for upstream. It looks to me like the fix made was strictly to correct the reported behavior. I'm not entirely sure it would be satisfactory for all users of portaudio.

Do you have any better ideas? I have not looked into this code really and I am unfamiliar with DirectSound.

Issue #322, opened 2020-10-07
portaudio.def for msvc is missing an entry for Pa_GetVersionInfo()


closetcodebrews:
(Please use the mailing list for support requests and general discussion. This is only for actual bugs.)

**Describe the bug**
The portaudio.def file included as a part of the msvc build is missing an entry for Pa_GetVersionInfo.
Came across this as I started try / use your examples/pa_devs.c file.  It tries to call the above method but fails.  You end up with an error 

Error   LNK2019 unresolved external symbol Pa_GetVersionInfo referenced in function main  

**To Reproduce**
1.  follow instructions to build portaudio on windows.  compiled 64 bit dll.
2.  create a new windows console application.  compile and make sure you get the default "hello world" message.
3.  copy in contents from pa_devs.c example file into the main cpp file of your console app. 
4.  change project properties to include the lib and headers. 
5.  copied the 64bit dll into the console application folder to simplify things. 
6.  Try to build console application. 

**Expected behavior**
System will return version information on the console.

**Actual behavior**
Error   LNK2019 unresolved external symbol Pa_GetVersionInfo referenced in function main  

**Desktop (please complete the following information):**
 - OS: Windows
 - OS Version 10
 - PortAudio version:  grabbed master branch just 2 days ago.
 - If Windows or Linux, which Host API (e.g. WASAPI):

**Additional context**

Patch was simple.  
Just added Pa_GetVersionInfo @69 into that file and it built just fine.
console application started working and listed all devices on my PC.

RossBencina:
We currently have two .def files, one in build/msvc and one in cmake_support. please use the same @ number in both. I recommend using @35, the lowest available number after Pa_Sleep. Insert the entry after Pa_Sleep, not later, where the host api extension symbols are defined.

RossBencina:
Related #323

RossBencina:
Fixed by https://github.com/PortAudio/portaudio/pull/574. Not sure why this ticket wasn't closed already. I have reconfirmed that `Pa_GetVersionInfo` is present in both .def files and that the ordinal @35 in both files matches. There is a comparison of all ordinals in both files here (they are perfectly matching):

https://docs.google.com/spreadsheets/d/1-J8BDLMq4SlUpskLLNh3YVDX4Xu2ry7OCXXS-l5bWXo/edit?usp=sharing

PR #574, opened 2021-05-14
- added Pa_GetVersionInfo to portaudio.def

mirko796:
FIXED: `build/msvc/portaudio.def` was missing Pa_GetVersionInfo so given function was not exported to the resulting DLL

### DIFF
--- a/build/msvc/portaudio.def
+++ b/build/msvc/portaudio.def
@@ -35,6 +35,7 @@ Pa_GetStreamReadAvailable           @31
 Pa_GetStreamWriteAvailable          @32
 Pa_GetSampleSize                    @33
 Pa_Sleep                            @34
+Pa_GetVersionInfo                   @35
 PaAsio_GetAvailableBufferSizes      @50
 PaAsio_ShowControlPanel             @51
 PaUtil_InitializeX86PlainConverters @52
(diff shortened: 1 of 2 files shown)


Be-ing:
We are discussing removing the .def file: #461

RossBencina:
Related:  322 "portaudio.def for msvc is missing an entry for Pa_GetVersionInfo()"

RossBencina:
@mirko796 Please could you also add the same change to cmake_support/template_portaudio.def

mirko796:
@RossBencina Sure, done

@Be-ing : 
> We are discussing removing the .def file: #461

That makes perfect sense to m, removing the def file will remove one thing from the maintenance list and IMHO I don't see any harm in exporting all symbols

philburk:
We might remove the .def files but that is a separate topic. As long as we have the .def files, we should maintain them. So this PR is much appreciated.

> IMHO I don't see any harm in exporting all symbols

That can increase the risk of collisions. Also we want to make clear which symbols are internal vs external. Then we can change the names of internal symbols without breaking any apps.

philburk:
Fixes #322

philburk (approved):

PR #589, opened 2021-05-31
Use GNUInstallDirs module instead of hard-coded path

h3ndrk:
This PR uses [GNUInstallDirs](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html) instead of hard-coded paths (e.g. `${CMAKE_INSTALL_LIBDIR}/...` instead of `lib/...`, etc.). The hard-coded paths do not work e.g. in Yocto because when generating multilib images the library directory is `/usr/lib64`. It is best practice to therefore use [GNUInstallDirs](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html) which detects most common environments and provides the correct paths.

This PR depends on #583.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -440,29 +440,31 @@ ENDIF()
 OPTION(PA_DISABLE_INSTALL "Disable targets install and uninstall (for embedded builds)" OFF)
 
 IF(NOT PA_OUTPUT_OSX_FRAMEWORK AND NOT PA_DISABLE_INSTALL)
+  INCLUDE(GNUInstallDirs)
   INCLUDE(CMakePackageConfigHelpers)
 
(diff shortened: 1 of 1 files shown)


Be-ing:
Already done in #461... please let's not make more merge conflicts.

philburk:
> Already done in #461... please let's not make more merge conflicts.

Thanks for pointing out that this an alternative PR. If we cannot move forward on #461 then this may be a good option. It is smaller in scope and easier to accept.

RossBencina:
@Be-ing please reconfirm that this was done in #461 and that we can now close it.

h3ndrk:
@Be-ing implemented this, I will close my PR. Thanks for the CMake rewrite!

PR #597, opened 2021-06-09
replace LPCTSTR with LPCWSTR

jmelas:
This PR removes the last occurance of TCHAR from the portaudio code by explicitly using wchar_t.

### DIFF
--- a/src/hostapi/wmme/pa_win_wmme.c
+++ b/src/hostapi/wmme/pa_win_wmme.c
@@ -329,13 +329,13 @@ static PaError CreateEventWithPaError( HANDLE *handle,
         LPSECURITY_ATTRIBUTES lpEventAttributes,
         BOOL bManualReset,
         BOOL bInitialState,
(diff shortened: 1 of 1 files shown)


RossBencina (approved):

RossBencina:
What is the logic of using Wide Char here instead of the ASCII version? The parameter is just for internal names of events.

At the moment this change has zero effect since the name parameter is allways passed NULL.

philburk (approved):

jmelas:
> What is the logic of using Wide Char here instead of the ASCII version? The parameter is just for internal names of events.
> 
> At the moment this change has zero effect since the name parameter is allways passed NULL.

From MSDN: "Internally, the ANSI version translates the string to Unicode"
https://docs.microsoft.com/en-us/windows/win32/learnwin32/working-with-strings
So the W functions are always faster.

RossBencina:
Thanks!

PR #603, opened 2021-06-22
Fix CMake failing on newer versions (3.10+)

undnull:
It was spitting a bunch of undefined CMAKE_C_XXX errors like this:
```
CMake Error: Error required internal CMake variable not set, cmake may not be built correctly.
Missing variable is:
CMAKE_C_COMPILE_OBJECT
```


Be-ing:
This is taken care of in #461.

undnull:
Oh I see. I'll use that fork for now then

PR #613, opened 2021-07-12
cmake: Fixed examples and tests could not be built if shared library is enabled with PA_BUILD_SHARED.

dmitrykos:
Currently if shared library build is enabled (`PA_BUILD_SHARED = ON`) and Examples and Tests are enabled they will fail to compile and link.

This PR fixes above situation, so Examples and Tests are built and linked correctly in case of shared library build.

The reason why some Examples and Tests could not be compiled is that they are using some private API of PortAudio from `src/common` file space. Before this PR these functions could be linked only if PortAudio is build as a static library.

To succeed for the shared library PR introduces possibility to export internal PortAudio functions selectively with PA_LIB_API prefix (see src/common/pa_lib.h for more details). In case of shared library the PA_LIB_SHARED and PA_LIB_EXPORTS must be defined in order to export some private PortAudio API which are then used by tests and some examples.

As a result the shared library is participating in tests and examples directly and therefore tests are testing actual binary.

This PR also fixes problem actively discussed in this PR #461 and allows to avoid usage of Object Library feature which would require min CMake 3.12 that would limit compatibility of PortAudio with older systems, for example Ubuntu 16.04 by default comes with CMake 3.5.1.

The implementation is tested (examples and tests link correctly and work) on Ubuntu 16.04, Windows 10 with MinGW32 and Visual Studio 2019.


Be-ing:
This approach was already discussed in #461/#285. I am in doubt that introducing another means to export some functions separate from the .def file is a good idea.

dmitrykos:
> I am in doubt that introducing another means to export some functions separate from the .def file is a good idea.

Such proposed selective export does not deprecate current .def but allows to export only that API which is needed for tests and examples. Tests and examples violate the usage of API anyway, normally they should not access private APIs but to solve this gently the selective export is a good solution. 

It allows to test and run examples against the shared binary that is a direct purpose of tests and examples. Another alternative approach is to move internal API to the public space or export via a wrapping functions with specific public names although it will be too bulky to my view.

RossBencina:
This change appears to export PA internal symbols. We already agreed that this is not a desirable approach.

dmitrykos:
Exporting symbols could be done for example only when tests/examples are built to allow correct linkage. Relying on .def does not look flexible to me, although if it is decided like so the PR is closed.

PR #614, opened 2021-07-21
dsound: Use appropriate wrapper functions for COM objects.

Biswa96:


Be-ing:
What problem does this fix?

Biswa96:
Two problems:

1. If I enable `PA_USE_DIRECTSOUNDFULLDUPLEXCREATE` in CMakeLists.txt file for MINGW these errors are shows

```c
portaudio/src/hostapi/dsound/pa_win_ds.c: In function 'InitFullDuplexInputOutputBuffers':
portaudio/src/hostapi/dsound/pa_win_ds.c:1558:14: warning: implicit declaration of function 'IUnknown_QueryInterface'; did you mean 'IDirectSound_QueryInterface'? [-Wimplicit-function-declaration]
 1558 |         hr = IUnknown_QueryInterface( pCaptureBuffer8, &IID_IDirectSoundCaptureBuffer, (LPVOID *)&stream->pDirectSoundInputBuffer );
      |              ^~~~~~~~~~~~~~~~~~~~~~~
      |              IDirectSound_QueryInterface
portaudio/src/hostapi/dsound/pa_win_ds.c:1564:9: warning: implicit declaration of function 'IUnknown_Release' [-Wimplicit-function-declaration]
 1564 |         IUnknown_Release( pCaptureBuffer8 );
      |         ^~~~~~~~~~~~~~~~
```

2. Take `pCaptureBuffer8` variable. It is of type `LPDIRECTSOUNDCAPTUREBUFFER8` or `IDirectSoundCaptureBuffer8*`. So, it should use the wrapper function of that type, like these `IDirectSoundCaptureBuffer8_QueryInterface`. Though it is same as `IUnknown_QueryInterface` but the type specific functions are proper here.

RossBencina:
This looks like it introduces a dependency on DirectSound 8, which I don't think we had before. 

I would have thought that `IUnknown_QueryInterface` and ` IUnknown_Release` have definitions somewhere. Is it possible that there is a missing include?

Biswa96:
> This looks like it introduces a dependency on DirectSound 8

The code is already using dsound8. This change is just using matching wrapper functions. Like, green bins for biodegradable, blue for glass, red for non-biodegradable waste.

Biswa96:
> This looks like it introduces a dependency on DirectSound 8

The code is already using dsound8 e.g. `LPDIRECTSOUNDBUFFER8 pRenderBuffer8; LPDIRECTSOUNDCAPTUREBUFFER8 pCaptureBuffer8`. This change is just using matching wrapper functions. Like, green bins for biodegradable, blue for glass, red for non-biodegradable waste.

RossBencina:
The code only uses dsound8 if PAWIN_USE_DIRECTSOUNDFULLDUPLEXCREATE is defined. We need to support building without dsound8.

Please confirm that `IUnknown_QueryInterface` and `IUnknown_Release` do not exist.

Biswa96:
OK, understood.

PR #619, opened 2021-08-05
wasapi: Improved reliability of the supported platform check.

### LINKED ISSUE #617: Port audio on winXP fails at WASAPI backend when avrt.dll is present
In windows XP, when avrt.dll is present, portaudio fails on WASAPI backend and fails to consider other backends.
Renaming/deleting avrt.dll solves issue and portaudio correctly uses other available backends.

Expected behaviour is that portaudio discards WASAPI regardless of whether it finds "avrt.dll" or not and picks up other working backends.

[pabad.txt](https://github.com/PortAudio/portaudio/files/6913215/pabad.txt)
[pagood.txt](https://github.com/PortAudio/portaudio/files/6913216/pagood.txt)
portaudio_fail

dmitrykos:
Check for the supported Windows version explicitly in PaWasapi_Initialize() and avoid relying on existence on avrt.dll for detecting the compatible platform. 

User could provide avrt.dll stub that will break initialization of the library if OS version does not support WASAPI backend.

Fixes #617 issue.

### DIFF
--- a/src/hostapi/wasapi/pa_win_wasapi.c
+++ b/src/hostapi/wasapi/pa_win_wasapi.c
@@ -1207,6 +1207,8 @@ static BOOL IsWindowsVersionOrGreater(WORD wMajorVersion, WORD wMinorVersion, WO
 }
 #endif
 // Get Windows version
+// note: We are trying to get Windows version starting from Windows Vista. Earlier OS versions
+//       will fall into WINDOWS_UNKNOWN case.
 static EWindowsVersion GetWindowsVersion()
 {
 #ifndef PA_WINRT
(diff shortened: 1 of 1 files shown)


MichalPetryka:
I feel like the maintainers should discuss whether it's worth it to support old unsupported OSes like Windows XP in modern PortAudio versions. It's not like people still write new applications for them and software that supports such old OSes doesn't really update its dependencies I guess.

Be-ing:
I agree that supporting Windows XP with new versions of PortAudio 7 years after Microsoft dropped support for it is a bit ridiculous. On the other hand, this small change seems fine.

dmitrykos:
This fix does not directly deal with XP but it improves the platform compatibility check which was relying on checking for DLL existence only. WASAPI is not supported on anything lower than Windows Vista, so do not worry much about it ;)

pachuco:
> It's not like people still write new applications for them and software that supports such old OSes doesn't really update its dependencies I guess.

I write software for such old OSes, whenever feasible.

PR #621, opened 2021-08-12
CMake: set shared library version to 2.0.0

### LINKED ISSUE #538: cmake setup does not use shared object version
**Describe the bug**
When building 19.7.0 using cmake on Arch Linux, I noticed, that the setup is not using the correct shared object version.

**To Reproduce**
```
  cmake -DCMAKE_INSTALL_PREFIX=/usr \
        -DCMAKE_BUILD_TYPE='None' \
        -DPA_ALSA_DYNAMIC=ON \
        -DPA_BUILD_TESTS=ON \
        -Wno-dev \
        -B build \
        -S .
  make VERBOSE=1 -C build
  make VERBOSE=1 test -C build
  make DESTDIR="${pkgdir}" install -C build
  install -vDm 644 LICENSE.txt -t "${pkgdir}/usr/share/licenses/${pkgname}"
  install -vDm 644 README.* -t "${pkgdir}/usr/share/doc/${pkgname}"
```

**Expected behavior**
The correct shared object version is built (example data from the 19.6.0 package):

```
lrwxrwxrwx root/root         0 2020-06-03 13:52 usr/lib/libportaudio.so -> libportaudio.so.2.0.0
lrwxrwxrwx root/root         0 2020-06-03 13:52 usr/lib/libportaudio.so.2 -> libportaudio.so.2.0.0
-rwxr-xr-x root/root    195792 2020-06-03 13:52 usr/lib/libportaudio.so.2.0.0
lrwxrwxrwx root/root         0 2020-06-03 13:52 usr/lib/libportaudiocpp.so -> libportaudiocpp.so.0.0.12
lrwxrwxrwx root/root         0 2020-06-03 13:52 usr/lib/libportaudiocpp.so.0 -> libportaudiocpp.so.0.0.12
-rwxr-xr-x root/root     63256 2020-06-03 13:52 usr/lib/libportaudiocpp.so.0.0.12
```

**Actual behavior**
Only the shared object is created, but without versioning:

```
-rwxr-xr-x root/root    187656 2021-04-06 21:38 usr/lib/libportaudio.so
```

 - OS: Arch Linux
 - OS Version: n/a
 - PortAudio version: 19.7.0
 - If Windows or Linux, which Host API (e.g. WASAPI): n/a

**Additional context**
Without the correct shared object version, the versioning is reset (ABI compatibility is void) and all dependants have to be rebuilt against this version.
Builds using cmake can not be considered a working solution because of this at the moment.

Be-ing:
for consistency with autotools
Fixes #538


Be-ing:
@dvzrv can you confirm this is how it should be?

dvzrv:
I can't really comment on that right now, as it requires me to build a development version of portaudio (the patch does not apply cleanly on top of 19.7.0).

Generally though: I guess it makes absolute sense to have the SOVERSION and VERSION value be set by a variable instead of hardcoding it (at best close to the project version at the top of the document).

Be-ing:
Good idea. I don't know how PortAudio got into this state with disjointed .so versions and release version names, but :shrug:

Be-ing:
On second thought, maybe we should bump the .so version to 19.8 for the next release to resolve this confusing situation. What would be the consequences of that? IIUC that would trigger unnecessary rebuilds of downstream applications for this one release, but I think that may be worthwhile to resolve this confusing situation.

dvzrv:
Well, that would be up for the maintainers to decide I guess.
Generally the bump of the soname implies that there are breaking changes. For the record: There are several projects where the project version does not match the soname version (which does not mean that this is desired or pretty).

And yes, this would imply a rebuild for all dependants in downstream distributions.

Be-ing:
@philburk @RossBencina what are your thoughts about this?

philburk:
> Generally the bump of the soname implies that there are breaking changes. 

If I understand correctly, the SOVERSION is the Application Binary Interface version and not the implementation version. If the ABI did not change then we do not need to bump the SOVERSION. And the SOVERSION does not need to match the library version. Is that right?

But it might be nice to have them match. We could update the SOVERSION to match the PA version whenever the PA ABI changed. And then leave it alone until the PortAudio ABI changed. That may be what Be-ing is proposing.

I am open to recommendations. Ross?

Be-ing:
> If I understand correctly, the SOVERSION is the Application Binary Interface version and not the implementation version. If the ABI did not change then we do not need to bump the SOVERSION. And the SOVERSION does not need to match the library version. Is that right?

I think that is correct.

> But it might be nice to have them match. We could update the SOVERSION to match the PA version whenever the PA ABI changed. And then leave it alone until the PortAudio ABI changed. That may be what Be-ing is proposing.

Yes, it is confusing to have the SOVERSION not match the version name of the library.

Be-ing:
Ping @RossBencina what are you thoughts on this?

RossBencina:
We've merged #686. This should be updated or closed. Please confirm.

Issue #259, opened 2016-10-22
SCONS build fails for patest_sine


PortAudio-admin:
*Issue created by @philburk*

SCONS tries to build a bunch of test programs. The script is here:

    src/SConscript:202:testNames = ["patest_sine", .....

Some of the test programs have been converted to examples. For example "test/patest_sine.c" is now "examples/paex_sine.c". So "patest_sine" is not found and the build fails.

The file src/SConscript needs to be updated to use the current names.

PR #623, opened 2021-08-12
Remove scons

Be-ing:


RossBencina:
This should remove all of the scons files. For example, the following files also need to be removed.

```
/SConstruct
/src/SConscript
```

Possibly there are also other `SConstruct` and `SConscript` files elsewhere. I have not done an exhaustive search.

philburk:
There is an Issue #259 that says scons build was not working in 2016.

Be-ing:
> This should remove all of the scons files.

Good catch, fixed.

Be-ing:
ping

philburk:
I see that the MSVC build is failing.  I assume that is unrelated because I doubt that it uses scons.

philburk:
All tests just passed!

philburk (approved):

PR #618, opened 2021-08-05
fix JACK host API build for MSVC and MinGW

Be-ing:
based on #461

PR #626, opened 2021-08-19
JACK: add missing #include <pthread.h>

Be-ing:
This is required for building with MinGW.

### DIFF
--- a/src/hostapi/jack/pa_jack.c
+++ b/src/hostapi/jack/pa_jack.c
@@ -57,6 +57,7 @@
 #include <errno.h>  /* EBUSY */
 #include <signal.h> /* sig_atomic_t */
 #include <math.h>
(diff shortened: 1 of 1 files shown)


src/hostapi/jack/pa_jack.c
```diff
@@ -57,6 +57,7 @@
 #include <errno.h>  /* EBUSY */
 #include <signal.h> /* sig_atomic_t */
 #include <math.h>
+#include <pthread.h>
```
MichalPetryka, line 60: The PR says `regex.h`, yet this adds `pthread.h`, is this correct?

Be-ing: Huh, I guess I messed up the commit message.

philburk (approved):
This was extracted from #618

PR #627, opened 2021-08-19
JACK: remove unnecessary #include <unistd.h>

Be-ing:
This breaks the build with MSVC.


philburk (approved):
This was extracted from #618

PR #628, opened 2021-08-19
JACK: use POSIX pthread_equal instead of == to compare pthread_t

Be-ing:
pthread_t is not guaranteed to be an arithmetic type by the POSIX
standard. In pthread4w it is implemented as a struct, so ==
does not work to compare two pthread_t values.


philburk (approved):
This was extracted from #618

PR #629, opened 2021-08-19
GH Actions: run CMake build on all branches

Be-ing:
This allows contributors to test builds on GitHub Actions on their
own repositories without needing to open a pull request first.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -1,10 +1,6 @@
 name: CMake build
 
-on:
-  push:
-    branches: [ master ]
-  pull_request:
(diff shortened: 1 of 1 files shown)


.github/workflows/cmake.yml
```diff
@@ -2,9 +2,7 @@ name: CMake build
 
 on:
```
MichalPetryka, line 3:
```suggestion
on: [push, pull_request]
```

Be-ing: good idea, done

philburk (approved):

Issue #630, opened 2021-08-21
cmake: options should keep the PA_ prefix to avoid name pollution


amurzeau:
(Please use the mailing list for support requests and general discussion. This is only for actual bugs.)

**Describe the bug**
When using CMake, portaudio options are scattered in the global namespace as they don't have the same prefix. This can cause conflicts with other projects.

Thus, options should have the PA_ prefix like before (like CMAKE options which are almost always prefixed with `CMAKE`, same for many other projects).
Prefix was removed with commit https://github.com/PortAudio/portaudio/commit/242a0241f69bd359b692004f3974ce39ec5137fd.

**To Reproduce**
Steps to reproduce the behavior. Include code if applicable.

Use cmake and configure the project.
Options are named without prefix, like this:
 - DEBUG_OUTPUT
 - JACK
 - WMME
 - ASIO
 - SKELETON

**Expected behavior**

Options available in CMake should have a prefix like PA_ to avoid polluting the "global namespace":
 - PA_DEBUG_OUTPUT
 - PA_JACK
 - PA_WMME
 - PA_ASIO
 - PA_SKELETON

**Desktop (please complete the following information):**
 - PortAudio version: git: 093d7f7fdf2cd1dbed0cb888257b6b39b4bed7db

For example, in Qt Creator or cmake-gui, options are grouped by prefixes :
![image](https://user-images.githubusercontent.com/5435069/130329437-8e369176-28bf-47eb-af61-77ad45993d66.png)

amurzeau:
@Be-ing maybe you have a opinion about using PA_ prefix in CMake options ?

Be-ing:
This is weird and was removed intentionally.

amurzeau:
you meant unintentionally ?

Also, unrelated to variable but cmake stuff, headers are not part of add_library call, so IDEs won't show them anymore, I think its better to keep headers (even if they won't affect the build by itself)
![image](https://user-images.githubusercontent.com/5435069/130329971-053e4490-34d0-4bd0-aa35-e5e9dd78abd6.png)

Be-ing:
No, I meant intentionally. Adding unnecessary prefixes to everything is not conventional.

Be-ing:
> Also, unrelated to variable but cmake stuff, headers are not part of add_library call, so IDEs won't show them anymore, I think its better to keep headers (even if they won't affect the build by itself)

Please open a separate issue for that and specify what IDE you are talking about. It works fine as it is in KDevelop.

[...]

PR #631, opened 2021-08-21
CMake: Keep previous config naming to keep configuration options grouped

amurzeau:
To have a view what's really need to change, here is a PR.

Using the previous naming scheme allows to:
 - Keep all available configuration options grouped in GUIs like cmake-gui
   or Qt Creator.
 - Avoid requiring changes on user side due to change in config names.
 - Avoid name conflicts when vendoring (probably more often seen on Windows
   where binaries are not as available as on Linux)

I've checked that the build on the different platform is exactly the same (by diff-ing output log of the cmake workflow).
Also, I've fixed the examples for MME that where not using right variable (they were using `MME` instead of `WMME`, at line 52)

Closes: #630


dmitrykos (approved):
All is good for me and fixes the intermixing of PortAudio local options with CMake global constants and other top-level constants and variables by introducing developer-friendly grouping by prefix.

Be-ing:
1. I don't think the `PA_` prefix is a good idea.
2. Furthermore adding another superfluous `USE_` prefix is annoying.
3. This breaks cmake/PortAudioConfig.cmake.in

amurzeau:
I've updated the PR to fix cmake/PortAudioConfig.cmake.in.

dmitrykos:
> 2. Furthermore adding more superfluous `USE_` and `ENABLE_` prefixes is annoying.

More verbose name let's developer understand what such variable/constant is doing without digging into the code and checking its actual usage, e.g. makes code much more readable. So such naming pattern is perfectly fine and convenient, especially for developers seeing PortAudio build system for the first time.

.github/workflows/cmake.yml
```diff
@@ -90,9 +90,9 @@ jobs:
            ${{ matrix.cmake_options }}
            -DCMAKE_INSTALL_PREFIX=${{ matrix.install_dir }}
            -DCMAKE_BUILD_TYPE=${{ env.cmake_build_type }}
-           -DSKELETON=ON
+           -DPA_USE_SKELETON=ON
            -DBUILD_TESTING=ON
```
philburk, line 94: I'm curious about why PA_BUILD_EXAMPLES is prefixed and BUILD_TESTING is not. Is BUILT_TESTING a CMake convention?

philburk (commented):
I like the PA_USE_prefixes. I think that will help avoid collisions. We do this through PortAudio.

RossBencina:
I like that the `PA_USE_ASIO`-type symbols for host API selection are are consistent with the source code. I think this change is good. If it reverts to something that was already in use in the previous CMake config, that is another point in its favor -- there's no point breaking things without good reason.

This PR also includes things like `BUILD_EXAMPLES` -> `PA_BUILD_EXAMPLES` but does not change `BUILD_TESTING`. This is inconsistent. Consistency makes things easier to use. If there is no matching source-code symbol, either use `PA_` for both of these, or neither.

.github/workflows/cmake.yml
Be-ing: Yes, BUILD_TESTING is a CMake convention.

Be-ing:
> something that was already in use in the previous CMake config, that is another point in its favor

That is a legitimate reason, not making it easier to vendor a library.

.github/workflows/cmake.yml
dmitrykos: Isn't `BUILD_TESTING` variable/option set by CMake when only CTest is used?
https://cmake.org/cmake/help/v3.1/module/CTest.html

Currently according to CMakeLists.txt `BUILD_TESTING` is used as local option:

**set here**
https://github.com/PortAudio/portaudio/blob/06eab6074b74a3d8bd155bed714105707337c1e8/CMakeLists.txt#L9

**applied here**
https://github.com/PortAudio/portaudio/blob/06eab6074b74a3d8bd155bed714105707337c1e8/CMakeLists.txt#L364-L379

thus to my view here we have ambiguous symbol because CTest is not used by the script and it needs to be fixed by prefixing with PA_ to become `PA_BUILD_TESTING` or better `PA_BUILD_TESTS`.

dmitrykos:
> That is a decent reason, not making it easier to vendor a library.

Assigning clearly distinctive names (by using specific prefix, lile `PA_`) to the library-local variables and options makes code easily readable, maintainable and much more reliable because developer will not need to spend a lot of time parsing documentation of build tools in order to understand whether this or that variable is script-local or some global/tool dependent.

Be-ing:
I considered adapting the tests to use CTest, but it didn't seem appropriate because of the tests' reliance on the presence of real hardware.

dmitrykos:
Changes look fine now. @RossBencina, @philburk could we merge this PR?

philburk (approved):

Issue #632, opened 2021-08-21
cmake: headers are missing from add_library (which prevent them from appearing in some IDEs)


amurzeau:
**Describe the bug**

When using CMake, headers don't always appear in IDEs that use CMake model to show the project tree.
This includes:
 - Qt Creator (left)
 - Visual Studio (right)
(there are maybe other IDEs that also CMake model)

![image](https://user-images.githubusercontent.com/5435069/130334549-fde65594-6f73-4fc4-a039-46ade70ff2e9.png) ![image](https://user-images.githubusercontent.com/5435069/130334557-b3f6e7a5-100c-40d9-abe6-7a47011084f3.png)

**To Reproduce**
 - Qt Creator
   - Open Qt Creator and choose to open a Project
   - Choose CMakeLists.txt at the root directory of portaudio sources
 - Visual Studio
   - Use cmake to generate a visual studio project
   - Open portaudio.sln in Visual Studio

**Expected behavior**
The header files should also be present in the IDEs, which means they should be added to add_library/add_executable like regular .c/.cpp sources.

**Actual behavior**
Headers are not shown in some IDEs like Qt Creator or Visual Studio, see above images.

**Desktop (please complete the following information):**
 - OS: Windows 10 10.0.19041
 - PortAudio version: git master 093d7f7fdf2cd1dbed0cb888257b6b39b4bed7db

Be-ing:
Do your IDEs not have a shortcut to switch to the header associated with a .c or .cpp file?

amurzeau:
Yes it has shortcuts, but mostly for switch between headers and sources of the same name (as often used for C++ classes).
But sometimes you just want to open an header without having to open a .c or .cpp file beforehand, this is the main use case affected here.

dmitrykos:
To show header files in the IDE CMake script needs a correction which is something like this:

```c
file(GLOB LIB_INC "${CMAKE_CURRENT_SOURCE_DIR}/include/*.h")
source_group("Public Header Files" FILES ${LIB_INC})
```

Also grouping of the source code (.c) files can be improved too to avoid a single listing by using `source_group("Group Name" FILES ${SOME_SRC_FILES_AND_HEADERS})`.

PR #633, opened 2021-08-22
CMake: include headers in sources of PortAudio

amurzeau:
Some IDEs like Qt Creator or Visual Studio use CMake model to construct
the project tree.
So for these IDEs, headers need to be included in targets.
As a side note, this is also what is done in CMake's own sources they add all headers (private and public ones) as sources.
This can add a small cost of maintaining CMakeLists.txt when adding new files: both `.h` and `.cpp/.c` files need to be added, not only `.cpp/.c` files.

I found that this incur a change in generated package for `macOS Clang framework PortAudio build` (probably a positive change):
Previously, `lib/portaudio.framework/Headers` was empty, with this commit this is a folder containing public headers.
In fact, `PUBLIC_HEADER` property requires that referenced headers are part of target_sources as stated here: https://cmake.org/pipermail/cmake/2008-August/023625.html

MinGW headers are not added as they are not really part of PortAudio, but just because MinGW is missing them.

Fixes: #632

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,15 +11,29 @@ option(BUILD_EXAMPLES "Include example projects" OFF)
 
 add_library(PortAudio
   src/common/pa_allocation.c
+  src/common/pa_allocation.h
   src/common/pa_converters.c
+  src/common/pa_converters.h
   src/common/pa_cpuload.c
+  src/common/pa_cpuload.h
   src/common/pa_debugprint.c
+  src/common/pa_debugprint.h
   src/common/pa_dither.c
+  src/common/pa_dither.h
+  src/common/pa_endianness.h
   src/common/pa_front.c
+  src/common/pa_hostapi.h
+  src/common/pa_memorybarrier.h
   src/common/pa_process.c
+  src/common/pa_process.h
   src/common/pa_ringbuffer.c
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -289,6 +313,9 @@ elseif(UNIX)
   endif()
 endif()
 
+# Add public headers to sources of PortAudio (used by some IDEs to list them in project tree)
+target_sources(PortAudio PRIVATE "${PORTAUDIO_PUBLIC_HEADERS}")
```
dmitrykos, line 317: How about to try to additionally group public headers into `Public Header Files` group in IDE, will CMake create Public Header Files in your case?:

```cmake
source_group("Public Header Files" FILES ${PORTAUDIO_PUBLIC_HEADERS})
```

Be-ing:
Looks fine to me.

CMakeLists.txt
amurzeau: Yes, it is shown as `Public Header Files` in IDE, I've added that to the PR.
image

dmitrykos (approved):
Looks great, I think we can give it a go?

dmitrykos:
@amurzeau, thank you very much for the improvement.

PR #635, opened 2021-08-25
CMake: export compile definitions for host APIs

Be-ing:
This allows applications to guard usage of host API specific
extensions with #ifdefs without the application's build system
needing to check for the existence of the host API specific header.


Be-ing:
Any comments on this?

dmitrykos (approved):
Looks fine to me.

Issue #213, opened 2012-04-06
ALSA function snd_pcm_sw_params_set_xfer_align() deprecated


PortAudio-admin:
*Issue created by gineera on Assembla*

The semantics of 'xfer_align' in Alsa were changed (essentially, removed) from Alsa version 1.0.16 onwards, and the function marked 'Deprecated'.  This results in a build warning on all more recent Alsa versions, and the function is anyway a 'noop' stub.

To maintain backwards compatibility the call cannot simply be removed from pa_linux_alsa, as Ecasound devels found out, but needs to be called with the older Alsa versions.

See http://permalink.gmane.org/gmane.linux.alsa.devel/67752 for Ecasound notes.

RossBencina:
I am changing this to P2 because https://github.com/PortAudio/portaudio/pull/636 (-Werror CI improvements) depends on it being fixed.

PR #636, opened 2021-08-30
Remove -Werror from autotools/darwin build, enable it on CI for all platforms

veprbl:
The `-Werror` was added very long time ago: c1e1d067d2f6b0fe184eae7ebf1fcb7b2d5ff5c7
This flag is more suitable for running on CI or during development, but encumbers downstream packaging.

cc NixOS/nixpkgs#136060

### DIFF
--- a/.github/workflows/autotools.yml
+++ b/.github/workflows/autotools.yml
@@ -11,6 +11,9 @@ jobs:
 
     runs-on: ubuntu-latest
     name: Ubuntu
+    env:
+      CFLAGS: "-Werror"
+      CXXFLAGS: "-Werror"
 
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -30,6 +30,7 @@ jobs:
               -DPA_USE_ASIO=ON
               -DASIO_SDK_ZIP_PATH=asiosdk.zip
               -DCMAKE_TOOLCHAIN_FILE=cmake/toolchains/i686-w64-mingw32.cmake
+            werror_flags: "-Werror"
           - name: Windows MSVC
             os: windows-latest
             install_dir: C:\portaudio
@@ -43,6 +44,7 @@ jobs:
               -DPA_USE_ASIO=ON
               -DASIO_SDK_ZIP_PATH="asiosdk.zip"
               -DCMAKE_TOOLCHAIN_FILE=D:\a\portaudio\portaudio\vcpkg\scripts\buildsystems\vcpkg.cmake
+            werror_flags: "" # TODO: use "/WX"
           - name: Windows MinGW
             os: windows-latest
             install_dir: C:\portaudio
@@ -56,6 +58,7 @@ jobs:
               -DPA_USE_ASIO=ON
               -DASIO_SDK_ZIP_PATH="asiosdk.zip"
               -DCMAKE_TOOLCHAIN_FILE=D:\a\portaudio\portaudio\vcpkg\scripts\buildsystems\vcpkg.cmake
+            werror_flags: "-Werror"
           - name: macOS Clang
             os: macOS-latest
             install_dir: ~/portaudio
@@ -64,6 +67,7 @@ jobs:
             cmake_options:
               -DCMAKE_FRAMEWORK=OFF
               -DCMAKE_TOOLCHAIN_FILE=vcpkg/scripts/buildsystems/vcpkg.cmake
+            werror_flags: "-Werror"
           - name: macOS Clang framework
             os: macOS-latest
             install_dir: ~/portaudio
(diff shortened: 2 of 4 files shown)


Be-ing:
I don't think there's any good reason to remove this upstream. If packagers are annoyed by it, I think that's their problem and they can remove it if they want.

veprbl:
@Be-ing Would you accept a patch that adds -Werror for autotools on linux and for cmake builds?

Be-ing:
Thank you for noticing that inconsistency in the build system, which is one more argument for removing autotools. I think we should use `-Werror` with CMAKE_BUILD_TYPE=Debug.

philburk:
We definitely want -Werror somewhere in our CI to catch warnings.

@veprbl - Do you have more details on the problems -Werror causes for packagers?

r-burns:
Werror is good to enable for development, but unless your CI checks every possible compiler out there, warnings can always sneak by. For example, Nixpkgs ran into some new warnings with gcc 10. I think the recommended approach is to keep Werror off by default, and enable it on the CI using CPPFLAGS.

veprbl:
> @veprbl - Do you have more details on the problems -Werror causes for packagers?

The compilation errors caused by `-Werror` are not so helpful to the end users. They are often seen as low importance and not reported upstream (One could say that upstream, in theory, has access to all the same compilers, so what is there to report). In the end work is put into silencing those errors without any productive output.

> We definitely want -Werror somewhere in our CI to catch warnings.

That is understandable. How about adding it in the CI workflows?

philburk:
I am in favor of merging this after we add -Werror to the CI.

philburk:
Note that the Werror broke the MSVC CI.

veprbl:
> Note that the Werror broke the MSVC CI.

Correct. After using /WX it is still broken:
```
D:\a\portaudio\portaudio\build\asiosdk_2.3.3_2019-06-14\host\pc\asiolist.cpp(24,41): error C2220: the following warning is treated as an error [D:\a\portaudio\portaudio\build\PortAudio.vcxproj]
43
D:\a\portaudio\portaudio\build\asiosdk_2.3.3_2019-06-14\host\pc\asiolist.cpp(24,41): warning C4267: 'argument': conversion from 'size_t' to 'DWORD', possible loss of data [D:\a\portaudio\portaudio\build\PortAudio.vcxproj]
44
D:\a\portaudio\portaudio\build\asiosdk_2.3.3_2019-06-14\host\pc\asiolist.cpp(31,42): warning C4267: 'argument': conversion from 'size_t' to 'DWORD', possible loss of data [D:\a\portaudio\portaudio\build\PortAudio.vcxproj]
```

Any thoughts?

.github/workflows/autotools.yml
```diff
@@ -11,6 +11,8 @@ jobs:
 
     runs-on: ubuntu-latest
     name: Ubuntu
+    env:
+      CXXFLAGS: "-Werror"
```
RossBencina, line 15: Should this not also set `CFLAGS` or, alternatively, perhaps only `CPPFLAGS`. I am not 100% sure how to address both C and C++, but most of our files are `C`. Some discussion here:

https://stackoverflow.com/questions/5541946/cflags-ccflags-cxxflags-what-exactly-do-these-variables-control

RossBencina:
I think this is heading in the right direction. It's OK to limit `-Werror` to CI. Given that the problem with `/WX` on MSVC is with the ASIO SDK (something we can't fix), I think that we can move ahead with merging this without `/WX`.

.github/workflows/cmake.yml
```diff
@@ -19,6 +19,7 @@ jobs:
             cmake_generator: "Unix Makefiles"
             cmake_options:
               -DOSS=ON
+            werror_flags: "-Werror"
```
Be-ing, line 22: This should be done in CMakeLists.txt for `${CMAKE_BUILD_TYPE} STREQUAL "Debug"` and change the `cmake_build_type` variable in this GH Actions workflow file to `Debug`.

Be-ing:
Please update the title of the pull request to reflect what it is actually doing now.

.github/workflows/autotools.yml
veprbl: @RossBencina Right, if there is a C code, we should enable the flag there as well. `CPPFLAGS` would be the wrong flags, as CPP stands for C Pre Processor. I've implemented your suggestion using `CFLAGS` and `CXXFLAGS`.

.github/workflows/cmake.yml
```diff
@@ -16,6 +16,8 @@ jobs:
             cmake_options:
               -DPA_USE_OSS=ON
               -DCMAKE_TOOLCHAIN_FILE=vcpkg/scripts/buildsystems/vcpkg.cmake
+              -DOSS=ON
```
RossBencina, line 19: What is the purpose of this `-DOSS=ON`? Is this needed. I'm not sure whether this is an unrelated change.

RossBencina:
Regarding the warnings that are breaking CI, we already have tickets for these:
`jack_port_get_latency` deprecated: #641
`snd_pcm_sw_params_set_xfer_align` deprecated: #213

philburk:
Thanks for adding the -Werror to the CI. They caused CI to fail on the deprecated functions, which is good.

We should fix the deprecated functions before we merge this PR.

.github/workflows/cmake.yml
veprbl: Was an incorrect rebase on my side

dechamps:
FYI I am proposing a slightly different approach in #780.

PR #637, opened 2021-08-31
Fix documentation for Pa_OpenDefaultStream

allsopp:
For the function Pa_OpenDefaultStream, passing a value of 0 to numInputChannels
or numOutputChannels opens an output-only or input-only stream, respectively.


philburk (approved):
Thanks!

Issue #281, opened 2018-09-03
Library for Android


PortAudio-admin:
*Issue created by Anonymous on Assembla*

Need an android playback port audio library

philburk:
Closing as duplicate of #154

Issue #154, opened 2011-02-09
PortAudio implementation for Android


PortAudio-admin:
*Issue created by @RossBencina*

[Trac: Originally assigned to somebody]

We could add a PortAudio implementation for Android. This would have to work within the constraints of the NDK. It would need to work with pre-2.3 Anrdroid meaning that it would need to use AudioTrack via JNI interfaces pre-2.3 and OpenSL ES on Android 2.3 and later.

PR #638, opened 2021-09-05
OpenSLES candidate

Gundersanne:
Since PortAudio is on GH now I figured I might as well make a PR. I had some conversations on mailing lists & the old tracker ages ago but that sort of died down.

This is an OpenSLES hostapi for PortAudio and up, for Android 14 and up. And I would love some feedback on it. I added a README.md under `src/hostapi/opensles` which has some extra notes.

Because testing this on Android is a bit of a hassle, there might be some blind spots, but most of the tests included under `test/` work fine.

I'm sure there are also some areas where it could conform better to the PortAudio (hostapi-)api.

There's still loads of potential improvements, like exposing more of the OpenSLES specific features, like [the ability to choose a  performance mode](https://developer.android.com/ndk/guides/audio/opensl/opensl-prog-notes#performancemodes). I think there might be some support for sample rate conversion if low-latency isn't enabled (which is the default). Also https://github.com/Gundersanne/portaudio_opensles/issues/14 needs investigating.

Linking  #154 and  #281

---

Also has contributions from @jwinarske for CMake support and bugfixes, and @darksylinc for more bugfixing.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -278,6 +278,16 @@ elseif(UNIX)
     target_compile_definitions(PortAudio PRIVATE PA_USE_COREAUDIO=1)
     set(PKGCONFIG_LDFLAGS_PRIVATE
       "${PKGCONFIG_LDFLAGS_PRIVATE} -framework CoreAudio -framework AudioToolbox -framework AudioUnit -framework CoreFoundation -framework CoreServices")
+  elseif(ANDROID)
+      find_package(OpenSLES)
+      cmake_dependent_option(OPENSLES "Enable support for OpenSLES" ON OPENSLES_FOUND OFF)
+
+      if(OPENSLES)
+        target_link_libraries(PortAudio PRIVATE "${OPENSLES_LIBRARIES}")
+        set(PORTAUDIO_PUBLIC_HEADERS "${PORTAUDIO_PUBLIC_HEADERS}" include/pa_opensles.h)
+        target_sources(PortAudio PRIVATE src/hostapi/opensles/pa_opensles.c)
+        target_compile_definitions(PortAudio PRIVATE PA_USE_OPENSLES=1)
+      endif()
   else()
     # Some BSDs have a reimplementation of alsalib, so do not explicitly check for Linux.
     find_package(ALSA)
--- a/src/os/unix/pa_unix_hostapis.c
+++ b/src/os/unix/pa_unix_hostapis.c
@@ -52,6 +52,8 @@ PaError PaAsiHpi_Initialize( PaUtilHostApiRepresentation **hostApi, PaHostApiInd
 PaError PaMacCore_Initialize( PaUtilHostApiRepresentation **hostApi, PaHostApiIndex index );
 PaError PaSkeleton_Initialize( PaUtilHostApiRepresentation **hostApi, PaHostApiIndex index );
 
+PaError PaOpenSLES_Initialize( PaUtilHostApiRepresentation **hostApi, PaHostApiIndex index );
+
 /** Note that on Linux, ALSA is placed before OSS so that the former is preferred over the latter.
  */
 
@@ -97,6 +99,10 @@ PaUtilHostApiInitializer *paHostApiInitializers[] =
 
 #if PA_USE_SKELETON
         PaSkeleton_Initialize,
+#endif
+
+#if PA_USE_OPENSLES
+        PaOpenSLES_Initialize,
 #endif
 
         0   /* NULL terminated array */
--- a/Makefile.in
+++ b/Makefile.in
@@ -44,7 +44,7 @@ PALIB = libportaudio.la
 PAINC = include/portaudio.h
 
 PA_LDFLAGS = $(LDFLAGS) $(SHARED_FLAGS) -rpath $(libdir) -no-undefined \
-	     -export-symbols-regex "(Pa|PaMacCore|PaJack|PaAlsa|PaAsio|PaOSS|PaWasapi|PaWasapiWinrt|PaWinMME)_.*" \
+	     -export-symbols-regex "(Pa|PaMacCore|PaJack|PaAlsa|PaAsio|PaOSS|PaWasapi|PaWasapiWinrt|PaWinMME|PaOpenSLES)_.*" \
 	     -version-info $(LT_CURRENT):$(LT_REVISION):$(LT_AGE)
 
 COMMON_OBJS = \
@@ -150,6 +150,7 @@ SRC_DIRS = \
 	src/hostapi/wasapi \
 	src/hostapi/wdmks \
 	src/hostapi/wmme \
+	src/hostapi/opensles \
 	src/os/unix \
 	src/os/win
 
--- a/cmake/modules/FindOpenSLES.cmake
+++ b/cmake/modules/FindOpenSLES.cmake
@@ -1,0 +1,36 @@
+# - Find OpenSLES
+# Find the OpenSLES includes and libraries
+#
+#  OPENSLES_INCLUDE_DIR - where to find dsound.h
+#  OPENSLES_LIBRARIES   - List of libraries when using dsound.
+#  OPENSLES_FOUND       - True if dsound found.
+
+if(OPENSLES_INCLUDE_DIR)
+  # Already in cache, be silent
+  set(OPENSLES_FIND_QUIETLY TRUE)
+else()
+  find_package(PkgConfig)
+  pkg_check_modules(PC_OPENSLES QUIET OpenSLES)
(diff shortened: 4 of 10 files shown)


philburk:
Thank you for this contribution. But please note that OpenSL ES on Android is deprecated in favor of Oboe. Oboe is simpler and has many advantages over OpenSL ES.

https://listserv.cuit.columbia.edu/scripts/wa.exe?A2=PORTAUDIO;e3ed65e8.2102C&S=b

The design of Oboe was inspired by PortAudio. So it should be much easier to implement an Oboe backend.

https://github.com/google/oboe

darksylinc:
> Thank you for this contribution. But please note that OpenSL ES on Android is deprecated in favor of Oboe. Oboe is simpler and has many advantages over OpenSL ES.

That's cool but it is sounding like the PR will be rejected for that reason. To that end:

1. There's currently no Android backend. An OpenSL ES is infinitely better than no Android backend. Currently there is no way to get Android sound via portaudio; while this is a 3-year-old tested backend. It works out of the box and passes both ASAN* and static analysis** and has no compiler warnings
2. OpenSL ES backend supports older versions of Android oboe can't because OSLES predates it.
3. Other platforms also have multiple backends (e.g. jack vs alsa vs oss; dsound vs asio vs wasapi vs wdm-ks vs wmme). An Oboe backend can come up later; where portaudio can dynamically select between the two based on availability and superiority; like it does on other platforms

The OSLES backend may not be the best one; but at least it fulfills the role of getting audio output/input.


\* I'm not a dev of this backend. I am a user of relatively new portaudio that checked out the OpenSL ES fork because it was the only thing available for Android, and I was in charge of running tests on that backend to ensure it met minimum quality. To that end we ran ASAN (that's how we detected a bug and why I am now listed as a contributor), static analysis and overall code review of the backend.

** scan-build could only detect one issue in line 721 of `pa_opensles.c` where a nullptr may be dereferenced if portaudio user decides to use the portaudio API without specifying both output / no speakers and without input / no mic; i.e. an invalid usage of portaudio:
```cpp
    if( framesPerBuffer == paFramesPerBufferUnspecified ) {
        if( outputParameters ) {
            framesPerHostBuffer = (unsigned long) (outputParameters->suggestedLatency * sampleRate);
        }
        else {
            framesPerHostBuffer = (unsigned long) (inputParameters->suggestedLatency * sampleRate);
        }
    }
```
To the analyzer there is no guarantee `inputParameters` won't be nullptr if outputParameters is also a nullptr; however we didn't think there was a point of reporting it because this would be an invalid portaudio usage.


While we can't guarantee Gundersanne's work is a high quality implementation; it appears like it is (enough to convince us) and we were extremely surprised it worked out of the box without problems, which is a rare occurrence whenever an Android project is involved.

philburk:
Thank you for the analysis. I agree that this seems like a very good quality implementation. The problem is not with this code but with OpenSL ES. OpenSL ES has many severe limitations that are addressed with AAudio/Oboe.

> OpenSL ES backend supports older versions of Android oboe can't because OSLES predates it.

Actually Oboe is a supports both AAudio and OpenSL ES. Oboe uses AAudio after Android 8.1 (API 27) and falls back to using OpenSL ES on earlier versions. Oboe runs on devices back to API 16. It probably runs on API 14 but it is hard to find devices that old for testing.

> Other platforms also have multiple backends

There is currently no need for multiple back ends on Android. An Oboe back end is sufficient.

We could add this implementation now and then add an Oboe implementation later. But once we add an OpenSL ES implementation then it would be very difficult to remove it without breaking some app.

OpenSL ES is deprecated and all work on it is frozen. For native Android APIs we are putting our efforts into AAudio/Oboe.

philburk:
Sorry. I closed it accidentally. Discussion is still open.

Gundersanne:
The bulk of this was written a while back, and Android is a bit of a moving target.  But imo if it's deprecated (I didn't realize actually ^_^) then let's leave it at that, no point in maintaining or merging deprecated stuff.

It will live on in my fork(s) regardless. I'll ultimately leave it up to you to decide if this is worth having, but I'll close this for now then.

CMakeLists.txt
```diff
@@ -278,6 +278,16 @@ elseif(UNIX)
     target_compile_definitions(PortAudio PRIVATE PA_USE_COREAUDIO=1)
     set(PKGCONFIG_LDFLAGS_PRIVATE
       "${PKGCONFIG_LDFLAGS_PRIVATE} -framework CoreAudio -framework AudioToolbox -framework AudioUnit -framework CoreFoundation -framework CoreServices")
+  elseif(ANDROID)
```
Be-ing, line 281: Why is this gated on both UNIX and ANDROID? OpenSLES is not exclusive to either. I think this should be treated like JACK and allowed on all operating systems if the required library is found.

CMakeLists.txt
```diff
@@ -278,6 +278,16 @@ elseif(UNIX)
...
+      find_package(OpenSLES)
+      cmake_dependent_option(OPENSLES "Enable support for OpenSLES" ON OPENSLES_FOUND OFF)
+
+      if(OPENSLES)
+        target_link_libraries(PortAudio PRIVATE "${OPENSLES_LIBRARIES}")
```
Be-ing, line 286: Export an INTERFACE target from the find module rather than directly using `${OPENSLES_LIBRARIES}`

src/os/unix/pa_unix_hostapis.c
```diff
@@ -99,5 +101,9 @@ PaUtilHostApiInitializer *paHostApiInitializers[] =
         PaSkeleton_Initialize,
 #endif
 
+#if PA_USE_OPENSLES
+        PaOpenSLES_Initialize,
+#endif
```
Be-ing, line 106: I'm not sure if there are any OpenSL ES implementations for Windows, but if there are, this should be added to pa_win_hostapis.c as well.

Be-ing:
Unless someone is actively working on an Oboe backend or intends to do so imminently, I think it would be good to merge this.

CMakeLists.txt
Gundersanne: The implementation does contain some android-specific code, which could be preprocessored out where necessary of course to make this a universal unix implementation.

src/os/unix/pa_unix_hostapis.c
Gundersanne: I had a look around, there [seems to be 1 other implementation around according to wikipedia](https://en.wikipedia.org/wiki/OpenSL_ES#Implementations), and I'm pretty sure the 2nd implementation is not open source :/

darksylinc: IMO we should apply [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)

Even if OSLES is supported on Windows I can't imagine it being the better alternative to any of the other available backends

And if another platform is truly benefited from OSLES, then that can be evaluated when the time arrives

Be-ing:
Please add a condition to the cmake_build.yml matrix to build this on GitHub Actions. You can use https://github.com/marketplace/actions/setup-android-ndk

CMakeLists.txt
Be-ing: If you want to make this compatible with non-Android systems, great, but otherwise I think it's fine to add a comment here in CMakeLists.txt explaining this has Android specific code even though OpenSL ES is not exclusive to Android.

Be-ing:
If this is adapted to build without Android, it could be useful to merge even if there is also an Oboe backend.

Gundersanne:
@Be-ing Thank you for the review, & all reasonable suggestions. Especially the CMake stuff needs a little work I agree.

I'll reopen again for visibility, to see what some of the other maintainers think. I don't really have plans to implement an Oboe backend. Regarding this PR I'm happy either way, depending on what the consensus is I'm of course happy to finish it up according to review comments.

jwinarske:
@Be-ing happy to make changes to the CMake.  Also would consider implementing Oboe backend.

philburk:
I don't know of any other platforms that support OpenSL ES.

> Even if OSLES is supported on Windows I can't imagine it being the better alternative to any of the
> other available backends

I agree. I think OpenSL ES is a dead end.

Be-ing:
> Also would consider implementing Oboe backend.

That would be great!

> I don't know of any other platforms that support OpenSL ES.

In that case, if @jwinarske (or anyone else) writes an Oboe backend, let's not merge this.

philburk:
I am currently working on an Oboe backend for PortAudio.

calvin2021y:
any update for Open SL ES  ?

CMakeLists.txt
philburk: I don't know of any systems besides Android that support OpenSL ES.

PR #644, opened 2021-09-12
alsa: Fix building on BSD

alarixnia:
If alsalib is installed this causes a compilation failure :x

### DIFF
--- a/src/hostapi/alsa/pa_linux_alsa.c
+++ b/src/hostapi/alsa/pa_linux_alsa.c
@@ -79,6 +79,11 @@
 
 #include "pa_linux_alsa.h"
 
+/* Allow this to build on the BSDs if alsalib is found */
+#ifndef ESTRPIPE
+    #define ESTRPIPE EPIPE
+#endif
+
 /* Add missing define (for compatibility with older ALSA versions) */
 #ifndef SND_PCM_TSTAMP_ENABLE
     #define SND_PCM_TSTAMP_ENABLE SND_PCM_TSTAMP_MMAP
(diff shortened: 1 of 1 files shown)


Be-ing:
Thanks for opening this PR. This looks pretty hacky but for the single specific line of code where PortAudio uses ESTRPIPE it seems like a reasonable workaround. I am unsure if this should be merged upstream; I don't think upstream libraries and applications should have to compensate for ALSA reimplementations' incompleteness. I think it would be better to define this symbol in BSDs' alsalib reimplementations. Would that be feasible?

alarixnia:
The alsalib is not a reimplementation, but the same one used on Linux, outputting to an OSS-compatible device. I think the ESTRPIPE definition is needed because it's usually provided by Linux kernel headers.

Be-ing:
Okay. Please add a comment above the #ifndef explaining that.

Be-ing:
Come to think of it, I have a better idea how to resolve this. Instead of fudging an alias for a missing symbol to define it as another symbol that does exist, let's guard the code that uses the symbol:

```
    if( res == -EPIPE || res == -ESTRPIPE )
    {
        *xrun = 1;
    }
   // ESTRPIPE is provided by the Linux kernel but not available on BSDs which support alsalib
#ifdef ESTRPIPE
   else if (res == -ESTRPIPE)
   {
       *xrun = 1;
   }
#endif
    else
    {
        ENSURE_( res, paUnanticipatedHostError );
    }
```

Be-ing:
Related pull request in libsndfile: https://github.com/libsndfile/libsndfile/pull/775

philburk:
Be-ing's solution is nice. But I notice libsndfile uses

    #if defined ESTRPIPE && ESTRPIPE != EPIPE

Do we need that extra check to avoid compiler warnings when EPIPE==ESTRPIPE.

Be-ing:
@philburk can you approve the GitHub Actions jobs for this new contributor? The code looks good to me now but let's check that CI still passes before merging.

PR #648, opened 2021-09-16
configure.in: Properly disable AudioIO when not found

lanodan:
Not sure if the `./configure` et al. files should be updated in this PR as well… I lean towards removing them in favor of generating them only in tarballs if required by some distros (gentoo already runs it because of some patches related to Audacity).

### DIFF
--- a/.github/workflows/autotools.yml
+++ b/.github/workflows/autotools.yml
@@ -14,6 +14,8 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+    - name: autoreconf
+      run: autoreconf -if
     - name: configure
       run: ./configure
     - name: make
--- a/configure.in
+++ b/configure.in
@@ -139,7 +139,7 @@ if test "x$with_asihpi" != "xno"; then
 fi
 have_audioio=no
 if test "x$with_audioio" != "xno"; then
-    AC_CHECK_HEADERS([sys/audioio.h], [have_audioio=yes])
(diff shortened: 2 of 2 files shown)


.github/workflows/autotools.yml
```diff
@@ -14,6 +14,8 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+    - name: autoreconf
+      run: autoreconf -if
```
RossBencina, line 18: Please remove this. It is our policy to manually regenerate `configure` and commit it to the repo (yes, we forgot this last week).

philburk:
I forgot to generate a configure file for #645 
I will do that. You may want to wait until you can rebase on top of that PR to reduce merge conflicts.

philburk:
Thanks for spotting this error and suggesting a fix.
This PR seems stalled so let's close it and use #654 instead.

lanodan:
> Thanks for spotting this error and suggesting a fix.

You're welcome

> This PR seems stalled so let's close it and use #654 instead.

I'm not really sure what you mean by stalled:

> You may want to wait until you can rebase on top of that PR to reduce merge conflicts.

I did choose to wait, there wasn't any commits or messages after this.

philburk:
I'm sorry. It was not "stalled". You were waiting for me to check in the configure file.
But I could not do that without the fix in configure.in.
So I went ahead and did them together.

Thanks again for pointing this out.

PR #649, opened 2021-09-16
move Visual Studio project files out of 'build' directory

Be-ing:
This allows building CMake conventionally inside a directory
called 'build' without mixing it with files in version control.


Be-ing:
I emphasize again that there is no compelling reason for these to be in the code repository. CMake can easily generate them and CMake is included in Visual Studio now. Since removing these unnecessary files has been repeatedly refused, this PR is a harm reduction measure to at least get them out of the way of using CMake normally.

philburk (approved):

RossBencina:
I'm fine with this change. The MSVC files are being retained for use with older versions of MSVC.

Note that we need to update the docs at `doc/src/tutorial/compile_windows.dox` I'll do that now.

PR #652, opened 2021-10-03
Fix include directories missing in PortAudioTargets.cmake

jcelerier:
It seems that this was refactored recently, leading to nothing being added to PortAudioTargets.cmake, thus causing apps importing PortAudio that way not being able to find the PA includes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,7 +38,7 @@ add_library(PortAudio
 target_include_directories(PortAudio PUBLIC
   $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
   $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src/common>
-  $<INSTALL_INTERFACE:include>
+  $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
 )
 if(UNIX)
   target_compile_options(PortAudio PRIVATE -fPIC)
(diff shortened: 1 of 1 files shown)


jcelerier:
@Be-ing @philburk paging you since I believe that it's this commit which lost that: https://github.com/PortAudio/portaudio/commit/242a0241f69bd359b692004f3974ce39ec5137fd

dechamps:
I was actually in the process of writing that exact same patch… and then I noticed that you beat me to it 😀

I agree with the problem statement and the solution. @Be-ing @philburk This should be merged IMHO.

philburk (approved):

CMakeLists.txt
```diff
@@ -38,6 +38,7 @@ add_library(PortAudio
 target_include_directories(PortAudio PUBLIC
   $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
   $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src/common>
+  $<INSTALL_INTERFACE:include>
```
Be-ing, line 41:
```suggestion
  $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
```
Thanks for catching this. I think it would be better to use the GNUInstallDirs variable than hardcoding `include`.

jcelerier:
Thanks for the review, good point, I rarely use GNUInstallDirs so wasn't aware of the correct way 🙈

Be-ing (approved):

PR #657, opened 2021-10-29
alsa: Fix distorted sound from half-duplex capture.

kleinerm:
If an audio stream is only opened for capture in half-duplex
mode, then the self->playback.canMmap variable is false by
default. This triggers setup of a hostBufferSizeMode of
paUtilBoundedHostBufferSize, something that is only meant to
happen for playback streams under some conditions. This leads
to very distorted captured sound.

Fix this by also checking for self->playback.pcm, ie. that
the playback component is actually used, so this setting
only gets applied for half-duplex playback or full-duplex
playback + capture streams if needed.

Tested to work correctly now for playback, capture, and
full-duplex playback + capture.

Signed-off-by: Mario Kleiner <mario.kleiner.de@gmail.com>

### DIFF
--- a/src/hostapi/alsa/pa_linux_alsa.c
+++ b/src/hostapi/alsa/pa_linux_alsa.c
@@ -2694,7 +2694,7 @@ static PaError PaAlsaStream_DetermineFramesPerBuffer( PaAlsaStream* self, double
     PA_UNLESS( framesPerHostBuffer != 0, paInternalError );
     self->maxFramesPerHostBuffer = framesPerHostBuffer;
 
-    if( (!self->playback.canMmap && self->playback.pcm) || !accurate )
+    if( (self->playback.pcm && !self->playback.canMmap) || !accurate )
     {
         /* Don't know the exact size per host buffer */
         *hostBufferSizeMode = paUtilBoundedHostBufferSize;
--- a/src/hostapi/alsa/pa_linux_alsa.c
+++ b/src/hostapi/alsa/pa_linux_alsa.c
@@ -2694,7 +2694,7 @@ static PaError PaAlsaStream_DetermineFramesPerBuffer( PaAlsaStream* self, double
     PA_UNLESS( framesPerHostBuffer != 0, paInternalError );
     self->maxFramesPerHostBuffer = framesPerHostBuffer;
 
-    if( !self->playback.canMmap || !accurate )
+    if( (!self->playback.canMmap && self->playback.pcm) || !accurate )
     {
         /* Don't know the exact size per host buffer */
         *hostBufferSizeMode = paUtilBoundedHostBufferSize;


philburk:
@kleinerm - this seems like a good fix. But we would like to reproduce the problem and verify the fix.

Can you either create an Issue for this with more detail, OR add some repro steps here?

kleinerm:
Hi, thanks for the review. So in terms of additional explanation, i found this logic bug in the if-conditional about two years ago with my own PsychPortAudio driver, which is a complex wrapper around PortAudio for use by scripting languages like Matlab, GNU/Octave and Python. It is used for neuroscience/medical research applications, as part of Psychtoolbox-3 for Octave/Matlab and PsychoPy for Python. My commit with my workaround -- hacking into portaudio structs "behind the back" of portaudio to undo the effects of the wrong if-conditional in portaudio, can be found here:

https://github.com/Psychtoolbox-3/Psychtoolbox-3/commit/0da6f5f54f1484c42b170ab3acee4d44a657a01c#diff-e4f7227d347c1573c2bb28158abd0389717b155ff2434392d246f727302103fe

In pure half-duplex capture scenarios with user provided ``paFramesPerBufferUnspecified``, it resets the hostBufferSizeMode back from paUtilBoundedHostBufferSize, which apparently causes misconversion, to paUtilFixedHostBufferSize which works. Without this, at least on multiple tested onboard Intel HDA compliant audio chips and a USB soundcard, at typical pure audio-capture settings (ie. no playback, and therefore the whole ``self->playback`` struct does not get initialized, so all variables including ``canMmap`` stay zero / false), one gets distorted sound recording.

According to my commit above, this used to work correctly until portaudio from 24th May 2009, where this logic bug was introduced. I used to build and include my own customized libportaudio from before May 2009, and that worked perfectly fine wrt. audio capture on Linux+ALSA. When i switched to dynamically linking against the November 2018 upstream portaudio in November 2018, i hit this bug. So this logic error introduced a regression into portaudio which went unnoticed for almost 10 years. The problem happens at least for typical PC/Laptop/Macintosh onboard sound chips - i don't have any pro audio sound interfaces available to test at the moment.

I only now got around to turn my hack into a proper fix for you, sorry for the late report. Hosting on GitHub makes it faster to submit fixes - no need to sign up for yet another code hosting site/mailing-list/issue-tracker.

Wrt. reproduction, the only test case i have is my own software, which can be installed on Debian/Ubuntu via ``sudo apt install octave-psychtoolbox-3`` - and old version, or more recent versions from NeuroDebian: http://neuro.debian.net/pkgs/octave-psychtoolbox-3.html

The demo to run after launching octave via ``octave`` would be ``more off; BasicSoundInputDemo;`` - it will capture sound half-duplex until a keypess, then play the captured sound back.

However, as the production driver contains those hacks mentioned above to work around the portaudio logic bug, i'd have to build and attach a custom copy of the PsychPortAudio.mex driver binary here, which removes the workarounds to expose the bug.

Thoughts?

kleinerm:
Found it: The specific commit which introduced the regression in May 2009 is this one:
https://github.com/PortAudio/portaudio/commit/85a522fc4abb130b8e15131b1345bb857505c333

From that change (specifically these lines https://github.com/PortAudio/portaudio/commit/85a522fc4abb130b8e15131b1345bb857505c333#diff-3e7e9f0cc2582a6369526f96cc51af25755aaac51b02202b20f337530d802cbaL1834) it becomes clear that this was an unintended behaviour change for pure half-duplex capture, as the if condition is changed from

``if( !accurate )``

to

``if( !self->playback.canMmap || !accurate )``

overlooking that ``self->playback.canMmap`` can be false either because a playback stream was initialized which can't use memory mapping - the intended case this is supposed to handle, **or** because this is a half-duplex capture setup without a valid playback stream, so everything in ``self->playback`` is zero-initialized and therefore ``canMmap == false`` -- the unintended case it should have ignored. This pull request makes it ignore the unintended case.

kleinerm:
I just learned that the cause for distorted sound from half-duplex capture on ALSA is actually a bug in my own application.
My application is callback driven via the `paCallback()` method, and its recording path does not adapt to varying `framesPerHostbuffer` values across different callback invocations. It assumed a fixed `framesPerHostbuffer` during the whole session - the one from the first callback invocation.

So my app could only deal with `paUtilFixedHostBufferSize`, not with varying buffers from mode `paUtilBoundedHostBufferSize`, and that was the cause of audio artifacts, not some problem in the portaudio buffer processor.

While I still think this pull request is worth merging for fixing a real logic bug introduced over a decade ago, and described in the comments above, the severity of the bug is no longer "causes audio recording artifacts", but only "probably causes higher cpu load than necessary etc. due to the need for buffer adaptation, because the more computationally optimal `paUtilFixedHostBufferSize` setting doesn't get chosen when it could be chosen".

Surprisingly this bug in my code wasn't causing any other symptoms in over 12 years of production use of Portaudio.

philburk (approved):

src/hostapi/alsa/pa_linux_alsa.c
```diff
@@ -2694,7 +2694,7 @@ static PaError PaAlsaStream_DetermineFramesPerBuffer( PaAlsaStream* self, double
     PA_UNLESS( framesPerHostBuffer != 0, paInternalError );
     self->maxFramesPerHostBuffer = framesPerHostBuffer;
 
-    if( !self->playback.canMmap || !accurate )
+    if( (!self->playback.canMmap && self->playback.pcm) || !accurate )
```
RossBencina, line 2697:
```suggestion
    if( self->playback.pcm && (!self->playback.canMmap) || !accurate )
```

Check if the struct is valid before checking the `canMmap` field.

PR #660, opened 2021-11-09
Fixed a case-sensitive include

oleg68:
This PR fixed cross-compiling compiling on linux with mingw


RossBencina:
@dmitrykos is this the correct fix, or should the file be renamed to have mixed case?

philburk:
The file in PortAudio is all lower case.

Be-ing (approved):
Ubuntu build failures look like unrelated intermittent network error

philburk:
The CI failures were fixed by #665
@oleg68 - You may want to rebase on the latest top of tree.

oleg68:
> The CI failures were fixed by #665 @oleg68 - You may want to rebase on the latest top of tree.

Done

RossBencina:
@dmitrykos it's a small change but we'd like you to confirm whether the correct fix should be to change the include statement or the target file's name. I think it's probably important for the file name to match case with the system version, but I do not know for sure since it seems like many of these include files have had their case changed over time.

dmitrykos (approved):
The fix is ok. In the current Windows SDK (my version - 10.0.19041.0) this file has low-case name.

dmitrykos:
@RossBencina the PR is correct because according to my Windows SDK lookup this filename is in low-case now. In the past it was like in our sources but now not anymore.

PR #665, opened 2021-11-13
CI: use apt-get update to fix libasound2-dev

### LINKED ISSUE #664: CMake CI fails to install libasound2-dev
I created a new pull request for Mac. The CI fails for the CMake Ubuntu builds with this error:

E: Failed to fetch http://azure.archive.ubuntu.com/ubuntu/pool/main/a/alsa-lib/libasound2-dev_1.2.2-2.1ubuntu2.4_amd64.deb  404  Not Found [IP: 52.147.219.192 80]

https://github.com/PortAudio/portaudio/runs/4200168025?check_suite_focus=true

I suspect that this is actually a problem with GitHub and not our CI scripts.

philburk:
Fixes #664

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -80,7 +80,9 @@ jobs:
     - name: checkout Git repository
       uses: actions/checkout@v2
     - name: "[Ubuntu] install dependencies"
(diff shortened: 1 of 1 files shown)


Be-ing (approved):
Thanks. I'm used to `dnf` which doesn't have a separate metadata update step like `apt` does.

PR #667, opened 2021-11-18
dsound: Fix warning of redefining _WIN32_WINNT when compiling with GCC/MinGW

### LINKED ISSUE #549: "_WIN32_WINNT" redefined warning in pa_win_ds.c
@mgeier wrote:
There is a warning about another preprocessor definition:

    src/hostapi/dsound/pa_win_ds.c:54:0: warning: "_WIN32_WINNT" redefined
     #define _WIN32_WINNT 0x0400 /* required to get waitable timer APIs */
     ^
    [...]/mxe/usr/i686-w64-mingw32.static/include/_mingw.h:233:0: note: this is the location of the previous definition
     #define _WIN32_WINNT 0x502

@dmitrykos wrote:

May be it makes sense to adjust it to:

    #if !defined(_WIN32_WINNT) || (_WIN32_WINNT < 0x0400)
        #undef _WIN32_WINNT
        #define _WIN32_WINNT 0x0400 /* required to get waitable timer APIs */
    #endif

@RossBencina wrote:

I am not sure about redefining `_WIN32_WINNT ` if it is already defined. I think it is better to move the define further up the file before any include statements.

dmitrykos:
The warning is thrown as follows:
```c
[  5%] Building C object CMakeFiles/PortAudio.dir/src/hostapi/dsound/pa_win_ds.c.obj
pa_win_ds.c:54: warning: "_WIN32_WINNT" redefined
   54 | #define _WIN32_WINNT 0x0400 /* required to get waitable timer APIs */
      |
<command-line>: note: this is the location of the previous definition
```

`_WIN32_WINNT` is defined by GCC/MinGW and therefore it has to be checked before defining to 0x0400. If `_WIN32_WINNT` is defined to a lower value it will be redefined to 0x0400, therefore the expected behavior is preserved.

Closes #549

### DIFF
--- a/src/hostapi/dsound/pa_win_ds.c
+++ b/src/hostapi/dsound/pa_win_ds.c
@@ -47,11 +47,15 @@
 */
 //#define PA_WIN_DS_USE_WMME_TIMER
 
+#if !defined(_WIN32_WINNT) || (_WIN32_WINNT < 0x0400)
+    #undef _WIN32_WINNT
+    #define _WIN32_WINNT 0x0400 /* required to get waitable timer APIs */
+#endif
+
 #include <assert.h>
 #include <stdio.h>
 #include <string.h> /* strlen() */
 
-#define _WIN32_WINNT 0x0400 /* required to get waitable timer APIs */
 #include <initguid.h> /* make sure ds guids get defined */
 #include <windows.h>
 #include <objbase.h>


jmelas (changes requested):
I think this define could be removed completely.
We don't support windows lower than NT4 do we?

RossBencina:
Looks ok, but please move the updated code at the top of the file, as Phil and I already requested  in #549

dmitrykos:
@RossBencina, done!

philburk (approved):

RossBencina (approved):

jmelas:
I am surprised this was merged.
_WIN32_WINNT should not be defined in code. It should be defined in the build settings.
Both in configure and CMake build settings we already define _WIN32_WINNT=0x0501
So I still believe that this define
#define _WIN32_WINNT 0x0400 /* required to get waitable timer APIs */
should have been removed completely.

dmitrykos:
@jmelas, another PR can remove this `ifdef` section. This one was closing #549 as per its discussion. I agree that this guard clause `#define _WIN32_WINNT 0x0400` may not be needed because min. Windows version that supports DS is Windows 95 (0x0400) anyway.

philburk:
@jmelas - Can you please review and comment on #676 ?

PR #671, opened 2021-11-22
[cmake] Add a test that tries to use the generated CMake libs

jcelerier:
This way there should hopefully be less CMake usage requirements regressions


RossBencina:
It looks like this is failing CI. I don't know why. We'd need to fix whatever is wrong with CI before merging this.

.github/workflows/cmake.yml
```diff
@@ -116,6 +116,10 @@ jobs:
       run: cmake --build build --config ${{ env.cmake_build_type }} --parallel 2
     - name: install
       run: cmake --install build --config ${{ env.cmake_build_type }}
+    - name: test
+      run: |
+        cmake -S ./test/test-use-cmake -B build-test -DCMAKE_PREFIX_PATH="${{ matrix.install_dir }}"
```
philburk, line 121: We normally use underscores instead of dashes in file and folder names.

philburk (changes requested):
Thanks for adding the test. It looks like it has already found a problem generated libs.

philburk:
@Being - Do you know this CI is failing?
Is it because jack is not installed on the CI server?
Does CMakeLists.txt handle that correctly?

Be-ing:
This new test setup is setting `CMAKE_PREFIX_PATH`... but that shouldn't interfere with finding system packages :thinking:

.github/workflows/cmake.yml
jcelerier: wops okay, will fix

philburk: Thanks. Underscores now used.

philburk (commented):
We cannot merge until the CI passes.

philburk (approved):

philburk (changes requested):
Need to resolve CI failures.

philburk:
@jcelerier - I do not see an Issue filed for this.

Are the CI failures because there is a problem in cmake and this new test reveals it? If so then this is a good test.

Or is the test itself broken? I can't tell.

RossBencina:
This seems like a reasonable PR, however it needs to pass CI before merging.

@jcelerier could you please rebase it to master and see if the errors go away?

jcelerier:
Sure, rebased. I honestly don't have the time at all to fix every CI failure... but they pretty much indicate bugs as the patch is an example of how PortAudio is supposed to be used when included through CMake

jcelerier:
e.g. for the ubuntu  and macOS Clang ones: 

```
CMake Error at /usr/local/share/cmake-3.23/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find JACK (missing: JACK_INCLUDEDIR)
```

means that the generated config file is likely not recording where PortAudio found the JACK it is built against, same for that Regex issue on MSVC, etc.

RossBencina:
@be-ing: would you mind reviewing this patch and taking a quick look at the CI failures (see discussion above). My naïve interpretation is that there is something very wrong with our CMake configuration if a simple test like this fails on multiple platforms.

Be-ing:
cmake isn't being called with the same arguments as the main build, so it isn't finding the dependencies from vcpkg. Try adding this to the cmake command:

```
           -G "${{ matrix.cmake_generator }}"
           ${{ matrix.cmake_options }}
           -DCMAKE_INSTALL_PREFIX=${{ matrix.install_dir }}
           -DCMAKE_BUILD_TYPE=${{ env.cmake_build_type }}
           -DVCPKG_TARGET_TRIPLET=${{ matrix.vcpkg_triplet }}
```

RossBencina:
@Be-ing thanks. Would that command likely fix the Ubuntu and Mac cmake builds too?

philburk:
@Be-ing - you suggested adding those argument to the cmake command.
Would those be added to line 121 or line 122 in cmake.yml or both?

Be-ing:
I think that will fix all the broken CMake builds. The arguments only need to be passed to the CMake configure step, not the build step.

RossBencina:
@Be-ing looks like a different error on Ubuntu and Mac now:

```
CMake Error: CMake was unable to find a build program corresponding to "Unix Makefiles".  CMAKE_MAKE_PROGRAM is not set.  You probably need to select a different build tool.
CMake Error: CMAKE_C_COMPILER not set, after EnableLanguage
-- Configuring incomplete, errors occurred!
CMake Error: CMAKE_CXX_COMPILER not set, after EnableLanguage
```

Windows errors are different:

```
CMake Error: CMake was unable to find a build program corresponding to "Unix Makefiles".  CMAKE_MAKE_PROGRAM is not set.  You probably need to select a different build tool.
CMake Error: CMAKE_C_COMPILER not set, after EnableLanguage
-- Configuring incomplete, errors occurred!
CMake Error: CMAKE_CXX_COMPILER not set, after EnableLanguage
```

Be-ing:
> CMake Error: CMake was unable to find a build program corresponding to "Unix Makefiles".  CMAKE_MAKE_PROGRAM is not set.  You probably need to select a different build tool.

This is puzzling. Maybe try not specifying the `-G` parameter?

> regex.h POSIX header NOT found and NOT available from TRE

This is due to not finding the required dependency from vcpkg.

jcelerier:
There's a " Could not find toolchain file:
vcpkg/scripts/buildsystems/vcpkg.cmake " in some logs which may explain the
other errors ; wrong toolchain file -> zero CMake variable set ->
absolutely nothing will be found. But it seems to be the exact same path
than in the above "configure" invocation step so I don't really understand
why now it does not find it...


------------------------------
Jean-Michaël Celerier
*cto* ossia.io | *consulting inquiries* celtera.dev | *personal*
jcelerier.name
t: +336 81 31 53 08

Be-ing:
The relative paths to the vcpkg CMake toolchain file does not work because the test CMake command uses `test/test_use_cmake` as the source directory. I suggest trying absolute paths so the same path will work for the main build and this test.

jcelerier:
tried with ${GITHUB_WORKSPACE}, let's see

jcelerier:
hmm, this doesn't help - and for good reason, the test itself does not use vcpkg (and I don't think it has a reason to, apps which use portaudio should be able to build against a portaudio that was built with vcpkg without having to know about portaudio itself being built against vcpkg, right) ?

RossBencina:
Agree vcpkg should not be required by clients/users. I'm not sure it should be required to build portaudio, even in CI. But I'm not clear on the details of why it's being used right now. @Be-ing can you explain please?

Be-ing:
vcpkg is used for the dependency on JACK, plus TRE and pthreads4win on Windows.

RossBencina:
@Be-ing does the cmake build work on windows without vcpkg if JACK is disabled?

Really pthreads should not be used on Windows, but that's a limitation of our PA/JACK implementation I suppose.

.github/workflows/cmake.yml
```diff
@@ -116,6 +116,17 @@ jobs:
       run: cmake --build build --config ${{ env.cmake_build_type }} --parallel 2
     - name: install
       run: cmake --install build --config ${{ env.cmake_build_type }}
+    - name: "test configure"
```
RossBencina, line 119:
```suggestion
    - name: "configure build-test"
```

@jcelerier I am going to rename these CI steps, as much as anything to force re-run CI to get the latest error logs (the old ones have expired). You'll have to pull the changes to your local working copy.

.github/workflows/cmake.yml
```diff
@@ -116,6 +116,17 @@ jobs:
...
+      run: cmake
+          -S ./test/test_use_cmake -B build-test
+          -G "${{ matrix.cmake_generator }}"
+          -DCMAKE_PREFIX_PATH="${{ matrix.install_dir }}"
+          -DCMAKE_INSTALL_PREFIX=${{ matrix.install_dir }}
+          -DCMAKE_BUILD_TYPE=${{ env.cmake_build_type }}
+          -DVCPKG_TARGET_TRIPLET=${{ matrix.vcpkg_triplet }}
+          ${{ matrix.cmake_options }}
+    - name: "test build"
```
RossBencina, line 128:
```suggestion
    - name: "build build-test"
```

RossBencina:
There are now only two different errors. In all cases the errors are in the "configure build-test" step that is added by this PR.
 
 1. For "Ubuntu GCC" and "macOS Clang" the "configure build-test" step fails on JACK detection: 
 
  ```
 -- Checking for module 'jack'
  --   No package 'jack' found
  CMake Error at /usr/local/share/cmake-3.25/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
    Could NOT find JACK (missing: JACK_INCLUDEDIR)
    
    -- Configuring incomplete, errors occurred!
```
  
2. For "macOS Clang framework" the error is different:

```
CMake Error at /Users/runner/work/portaudio/portaudio/vcpkg/scripts/buildsystems/vcpkg.cmake:819 (_find_package):
  By not providing "FindPortAudio.cmake" in CMAKE_MODULE_PATH this project
  has asked CMake to find a package configuration file provided by
  "PortAudio", but CMake did not find one.
```

File: test/test_use_cmake/CMakeLists.txt [hunk 4]
```diff
@@ -0,0 +1,30 @@
...
+#include <portaudio.h>
+int main()
+{
+  Pa_Initialize();
+  Pa_Terminate();
+}
+]=]
+)
+
+find_package(PortAudio REQUIRED)
```
dmitrykos, line 21: `find_package()` will fail because CMake will try to find `FindPortAudio.cmake` in `cmake\modules` which is not provided. If intent of the implementation was to find built PortAudio shared library and link it to the test executable then probably `find_library()` is the best way to go, or `FindPortAudio.cmake` has to be provided with `find_library()` for built PortAudio library.

dmitrykos (changes requested):

test/test_use_cmake/CMakeLists.txt
philburk: @dmitrykos - Do we simply need to replace "find_package" with "find_library"?
Or is it a more complicated change?

If so then we can make that simple change and merge it if it works.

dmitrykos: Previously I commented wrongly about `find_package()` because PA CMake build actually creates `PortAudioConfig.cmake` in the installation path (I did not know that it was done).

My earlier proposer with `find_library()` ignores linkage of 3-rd party deps (ALSA, Regex, Jack, ...) which are pulled by `PortAudioConfig.cmake`, so some workloads

> Do we simply need to replace "find_package" with "find_library"?

The change for a test can as follows starting from the line 21:

```cmake
find_library(PORTAUDIO_LIB portaudio)
find_path(PORTAUDIO_INCLUDE portaudio.h)
include_directories(${PORTAUDIO_INCLUDE})

add_library(test_use_static STATIC "${CMAKE_CURRENT_BINARY_DIR}/main.cpp")
add_library(test_use_shared SHARED "${CMAKE_CURRENT_BINARY_DIR}/main.cpp")
add_executable(test_use_exe "${CMAKE_CURRENT_BINARY_DIR}/main.cpp")

target_link_libraries(test_use_static PRIVATE ${PORTAUDIO_LIB})
target_link_libraries(test_use_shared PRIVATE ${PORTAUDIO_LIB})
target_link_libraries(test_use_exe PRIVATE ${PORTAUDIO_LIB})
```

I will experiment with possible fixes in my fork.

dmitrykos: @philburk after experimenting in my fork I managed to get idea of this PR working, see my post in this thread.

dmitrykos:
After experimenting in my fork I managed to get idea of this PR working, see PR: https://github.com/dmitrykos/portaudio/pull/1

There were multiple issues related to `cmake.yml` and the main reason why `built-test` was not linking well is the vcpkg's specifics - CMake was not using vcpkg and could not locate vcpkg packages if CMake configuration is launched with source path specified which is deeper than root of the source repository, e.g. `-S some/path/to/sb-folder`. To make vcpkg work correctly I had to `cd` to `some/path/to/sb-folder` and only then launch cmake. Then it worked :)

I also cleaned up `PortAudioConfig.cmake` to make it more consistent, as well as `cmake.yml` by removing all hardcoded paths - now it looks less bulky.

Ubuntu MinGW has issues with linker asserting when it is trying to link Jack library, so Jack is excluded from that build type without much affecting the consistency of the test because it is used in other builds.

@jcelerier could you add me to your repository so I could commit changes to this PR or you could copy paste all changes from my PR, so we could finally complete this task.

dmitrykos:
Looks like PR's thread is stalled? @RossBencina it seems you are able to commit into the source of this PR, could you update its implementation to my (https://github.com/dmitrykos/portaudio/pull/1) fixed one? Or if not this PR could be closed and I will reopen my referencing this PR.

RossBencina:
@dmitrykos yes please submit a new PR and we'll close this one.

dmitrykos:
Closed as stalled PR, continuation is in #866.

PR #676, opened 2021-12-09
dsound: Remove _WIN32_WINNT guard clause due to its needlessness

dmitrykos:
DirectSound is supported on min. Windows 95 (`_WIN32_WINNT_NT4 0x0400`) therefore checking for `_WIN32_WINNT < 0x0400` is ambiguous.

It is a continuation of discussion in PR #667 raised by @jmelas.

### DIFF
--- a/src/hostapi/dsound/pa_win_ds.c
+++ b/src/hostapi/dsound/pa_win_ds.c
@@ -47,11 +47,6 @@
 */
 //#define PA_WIN_DS_USE_WMME_TIMER
 
-#if !defined(_WIN32_WINNT) || (_WIN32_WINNT < 0x0400)
-    #undef _WIN32_WINNT
-    #define _WIN32_WINNT 0x0400 /* required to get waitable timer APIs */
-#endif
-
 #include <assert.h>
 #include <stdio.h>
 #include <string.h> /* strlen() */


philburk:
In PR #667, a #define was moved into a section guarded by a #if.
The PR removes it entirely.
Let's see if we can Ross and @jmelas to agree.
Dmitry submitted the PR so he must already agree.

jmelas:
Yes I am quite happy with the define being removed completely!
Since we do not support Windows versions less than Windows 95/NT4
it does not make sense to have this define in code to explicitly request support
for waitable timers!

dmitrykos:
@philburk when @jmelas wrote his critical comment I double checked what is the minimal Windows version for Direct Sound and it appeared that it is Windows 95, e.g. `0x0400`. I was just automatically fixing warning without realizing that. 

Therefore that check is actually excessive and can be safely removed because DS backend can not operate on older OS version anyway, not just waitable timers.

RossBencina:
I am fairly sure that I'm the one that originally implemented waitable timers for PA/DirectSound. The _WIN32_WINNT_NT check was definitely needed (possible in Borland C++ Builder).

For a Windows 95 target  _WIN32_WINNT_NT will _NOT_ be set. See the old version from MSDN quoted under "The following table indicates the macros you must define to target each system using the SDK headers." here: 

https://groups.google.com/g/microsoft.public.vc.mfc/c/2CVWj9m1ky0?pli=1


There is a general discussion here: https://devblogs.microsoft.com/oldnewthing/20070411-00/?p=27283

The problem that is being solved is to make sure that the SDK headers provide the Waitable Timer  API.  If  _WIN32_WINNT_NT  < 4 then they won't be  available and the code will not compile.

philburk:
Ross pointed out a situation where the guard is necessary.
And there is no evidence of the guard being harmful in any situation.
So let's leave the guard in place.
Closing this PR.

jmelas:
Still I am not happy with the code that was merged.
If we really want to support Windows 95 (I mean it's 2021)
it should be written as
#ifndef _WIN32_WINNT
    #define _WIN32_WINNT 0x0400 /* required to get waitable timer APIs on Windows 95*/
#endif
otherwise the define can be removed completely.

PR #678, opened 2021-12-11
Fix WASAPI stream stopped check

AnErrupTion:
This fixes a bug in the WASAPI implementation, specifically the ``IsStreamStopped`` function. It would essentially return ``false`` when the stream was actually stopped and where it should've returned ``true``.

This is a patch from Audacity but that isn't really correlated to it.

Patch source: https://github.com/SartoxSoftware/audacium/blob/9e763b8f420c1569a9024087859349b28b888035/lib-src/portaudio-v19/wasapi-fix.patch

### DIFF
--- a/src/hostapi/wasapi/pa_win_wasapi.c
+++ b/src/hostapi/wasapi/pa_win_wasapi.c
@@ -601,9 +601,9 @@ typedef struct PaWasapiStream
     // must be volatile to avoid race condition on user query while
     // thread is being started
     volatile BOOL running;
-    BOOL stopped;
 
     // stream has not or is no longer started
+    BOOL stopped;
 
     PA_THREAD_ID dwThreadId;
     HANDLE hThread;
--- a/src/hostapi/wasapi/pa_win_wasapi.c
+++ b/src/hostapi/wasapi/pa_win_wasapi.c
@@ -601,6 +601,9 @@ typedef struct PaWasapiStream
     // must be volatile to avoid race condition on user query while
     // thread is being started
     volatile BOOL running;
+    BOOL stopped;
+
+    // stream has not or is no longer started
 
     PA_THREAD_ID dwThreadId;
     HANDLE hThread;
@@ -3797,6 +3800,9 @@ static PaError OpenStream( struct PaUtilHostApiRepresentation *hostApi,
     if (framesPerBuffer == 0)
         framesPerBuffer = ((UINT32)sampleRate / 100) * 2;
 
+    stream->stopped = TRUE;
+    stream->running = FALSE;
+
     // Try create device: Input
     if (inputParameters != NULL)
     {
@@ -4522,6 +4528,7 @@ static PaError StartStream( PaStream *s )
         // Signal: stream running.
         stream->running = TRUE;
     }
+    stream->stopped = FALSE;
 
     return result;
 
@@ -4563,6 +4570,7 @@ void _StreamFinish(PaWasapiStream *stream)
     _StreamCleanup(stream);
 
     stream->running = FALSE;
+    stream->stopped = TRUE;
 }
 
 // ------------------------------------------------------------------------------------------
@@ -4596,7 +4604,7 @@ static PaError AbortStream( PaStream *s )
 // ------------------------------------------------------------------------------------------
 static PaError IsStreamStopped( PaStream *s )
 {
-    return !((PaWasapiStream *)s)->running;
+    return ((PaWasapiStream *)s)->stopped;
 }
 
 // ------------------------------------------------------------------------------------------


src/hostapi/wasapi/pa_win_wasapi.c
```diff
@@ -601,6 +601,9 @@ typedef struct PaWasapiStream
     // must be volatile to avoid race condition on user query while
     // thread is being started
     volatile BOOL running;
+    BOOL stopped;
+
+    // stream has not or is no longer started
```
philburk, line 606: We normally put comments before the code.

philburk (changes requested):
Thanks for this contribution.
In most of the code, stopped is just set to the opposite of running.
Please describe the circumstance when stopped is different than !running.
Is there a window of time when a stream is being started?
It might help to add a bug that describes the problem that this fixes.

src/hostapi/wasapi/pa_win_wasapi.c
AnErrupTion: Ah, yes, that is a mistake! I'll fix that.

AnErrupTion:
I found this bug in Audacity (but I maintain my own fork of it), so tracing it will be rather complicated. What happens is that, at the end of a track (at playback), it will never stop, and trying to stop it manually doesn't work. I think in that case, a separate variable (like ``stopped``) might be necessary. (Just making some assumptions there, it's not even my patch to begin with after all!)

It is also most likely a matter of time as you asked.

AnErrupTion:
Also, ``stopped`` isn't always the opposite of ``running``. By that, I mean that, for example, ``running`` could be set to ``true``, but stopped could not (or vice versa). See for example at line 4531, ``stopped`` is set to ``false``, but you can also see that ``running`` is only set to ``true`` if the stream is blocking.

philburk:
Thanks for the additional background.

In order to evaluate a fix, we need to understand the bug.
Ideally we should be able to reproduce it and then see if the patch solves the problem.
We could also experiment to see if a simpler patch also solves the problem.
Adding an additional stream state variable does not feel quite right.

>  What happens is that, at the end of a track (at playback), it will never stop, and trying to stop it manually doesn't work.

We need to reproduce this in a test program using WASAPI. 
What happens if we do this?
1. Play a stream with a callback for a few seconds.
2. Call Pa_StopStream() but do NOT call Pa_CloseStream().
3. sleep(4)

I would expect the sound to stop.
If it does not stop then that would be a bug,

Or maybe if we did something similar but kept writing to a blocking stream.

> but you can also see that running is only set to true if the stream is blocking.

Maybe that is part of the bug.

dmitrykos:
I agree with @philburk, the bug with `IsStreamStopped()` must be reproduced first. According to my inspection of the code there should not be any issue with stalled `running` state, at least I can't spot it both polled and non-polled implementations.

At a time when that fix  was implemented probably there was a bug in the non-polled implementation of WASAPI (I suspect) with `running` not cleared correctly but it is unrelated to the current state of WASAPI implementation. Moreover proposed PR is incomplete and does not cover all cases when stream is stopped/started, so it introduces misbehavior.

My proposal is - if there is no test case that confirms a bug with `IsStreamStopped()` then this PR shall be closed as unrelated.

dmitrykos (changes requested):
`stopped` state does not cover all stop/start situations, implementation is incomplete. In general, it is an opposite to `running` state and therefore does not make any sense.

AnErrupTion:
> At a time when that fix was implemented probably there was a bug in the non-polled implementation of WASAPI (I suspect, as it was applied to some old v19 version)

The bug still exists since I was able to reproduce it.

I'll try to make a simpler patch and see if it works.

dmitrykos:
@AnErrupTion it would be great if you could write a small test case, based on one of the PortAudio's examples which hits this bug. There is probability that you misuse PA's API somehow. The patch itself is generally incorrect, you can trace why by checking all places where `running` variable is modified.

AnErrupTion:
I'm also starting to think the API is misused, but I have yet to find where.

dmitrykos:
@AnErrupTion when you call `StopStream()` the `running` will be set to `FALSE` when this function returns. So in this case `IsStreamStopped()` shall report correct state. If that does not happen and you call these functions then a memory corruption can be inside the process memory of your application which breaks correct operation.

AnErrupTion:
I cannot seem to be able to make a simpler patch, it just doesn't work (which is weird to me).
Do you people think it's okay if I keep the ``stopped`` variable but complete its integration? I know it doesn't make much sense to have 2 separate variables, but I've really been trying to find out the source of the bug and can't seem to, which is even more weird...

AnErrupTion:
Ah, interesting. So when completing the integration of the ``stopped`` variable, I actually notice that the bug shows up again. So it seems like it's an issue on a very specific part of the code (or when that code is executed). That means a ``stopped`` variable wouldn't make sense, but a "temporary" variable would make more sense in that case. Though I'm not really a fan of that.

dmitrykos:
@AnErrupTion keeping opposite state in a separate variable does not make much sense. My advise is to use one of the existing PA's examples and try to reproduce the problem you are facing using it. Then if you manage to do so you could post the source code here for the discussion.

philburk:
According to the docs, there are two stream queries for the running state that are slightly different:
http://files.portaudio.com/docs/v19-doxydocs/portaudio_8h.html#a1f8709c4971932643681a6f374c4bb5a

Generally active is the opposite of stopped.
But if a callback returns a value other than paContinue then both active and stopped are false.
Until Pa_StopStream() is called and then stopped is true.

The implementation of WASAPI IsStreamStopped() always returns the opposite of IsStreamActive() so it is likely that this behavior is incorrect with WASAPI. There are many Perhaps Audacity is relying on the correct behavior.

I added a test to patesst_stop_callbackstop.c at line 231 before the call to Pa_StopStream(). See #679

dmitrykos:
Phil thank you for the test #679 it managed to reveal 2 issues with stream active/stopped handling:

1.  A callback could be called again if `paContinue` wasn't returned previously
2.  The `IsStreamStopped()` and `IsStreamActive()` were not working as per referenced documentation (so I take my words back regarding - "does not make much sense"

The amount of corrections is larger and it is more complex than a proposed PR which is still missing necessary logic. Could we close this PR and I would open a new referencing this one?

philburk:
@AnErrupTion - thanks for bringing this problem to our attention.

@dmitrykos wrote:
> The amount of corrections is larger and it is more complex than a proposed PR which is still
> missing necessary logic. Could we close this PR and I would open a new referencing this one?

Sounds good. Closing.

PR #679, opened 2021-12-12
test: check active and stopped states

philburk:
Query Pa_IsStreamActive() and Pa_IsStreamStopped()
and verify correct response.

### DIFF
--- a/test/patest_callbackstop.c
+++ b/test/patest_callbackstop.c
@@ -150,7 +150,7 @@ int main(void);
 int main(void)
 {
     PaStreamParameters outputParameters;
-    PaStream *stream;
+    PaStream *stream = NULL;
     PaError err;
     TestData data;
     int i, j;
--- a/test/patest_callbackstop.c
+++ b/test/patest_callbackstop.c
@@ -124,6 +124,27 @@ static void StreamFinished( void* userData )
 }
 
 
+/**
+ * @return paNoError if successful, otherwise a negative error.
+ */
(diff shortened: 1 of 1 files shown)


philburk:
This test is to help verify #678

dmitrykos:
@philburk thank you for the test, I tried to run WASAPI against it and it failed due to a callback call after paComplete. The problem is probably outside #678 because `running` state variable is not participating in the scenario. 

There is possibility that buffer can be processed multiple times before hCloseRequest event is checked again. I am going to fix it and make test pass, then will comment in #678 regarding how we proceed further.

dmitrykos (commented):
The `stream` variable has to be initialized to NULL to avoid failing on bad pointer access if test fails (faced this problem):

https://github.com/PortAudio/portaudio/blob/18f1d9d9ada6d02d508357b23a5b999fb3e8ee9f/test/patest_callbackstop.c#L153

philburk:
Thanks Dmitry. I initialized stream to NULL.

philburk:
__LINE__ used. Thanks.

dmitrykos:
Phil, great! I pushed new #681 PR related to our discussion in #678. It passes this test successfully now.

RossBencina:
For reference, there is a state machine diagram here that indicates the correct active/stopped values for each stream state:

http://www.portaudio.com/docs/proposals/010-ClarifyStreamStateMachine.html

RossBencina (approved):

PR #681, opened 2021-12-13
wasapi: Fixed active and stopped stream states management and unexpected invocation of PaStreamCallback

dmitrykos:
This PR is a continuation of the work done in #678  and #679.

It introduces two important fixes in WASAPI implementation:

1. Active and stopped stream states management and reporting with `Pa_IsStreamActive()` and `Pa_IsStreamStopped()`
2. `PaStreamCallback` could be called again if `paContinue` was not returned during its previous call e.g. user code was intending to stop rendering by returning `paComplete` or `paAbort`

### DIFF
--- a/src/hostapi/wasapi/pa_win_wasapi.c
+++ b/src/hostapi/wasapi/pa_win_wasapi.c
@@ -569,7 +569,6 @@ PaWasapiHostProcessor;
 // ------------------------------------------------------------------------------------------
 typedef struct PaWasapiStream
 {
-    /* IMPLEMENT ME: rename this */
     PaUtilStreamRepresentation streamRepresentation;
     PaUtilCpuLoadMeasurer      cpuLoadMeasurer;
     PaUtilBufferProcessor      bufferProcessor;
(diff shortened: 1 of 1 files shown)


dmitrykos:
@AnErrupTion would you please try this version in your application and check if active/stopped states are working for you as expected now.

AnErrupTion:
I will soon!

RossBencina (approved):
Looks ok to me. Maybe wait until we see whether it fixes the Audacity issue though.

philburk (approved):
Looks good. Thanks for the fix.

BTW, it was a little hard to evaluate the fix because there was also lots of general cleanup.

dmitrykos:
@philburk, I agree, but some were necessary to stay inline with naming convention, such as `isSomeVar`, so I decided to update existing `bSomeVar` to look similar and make code more clear. There were minor additional corrections in the functions I crossed by and they could go into a different patch but I was worried not to recall what functions need to be updated. I makes sense to split a fix and refactoring though, will try to do it next time like this. Merging this so far.

PR #684, opened 2022-01-03
cmake: fix invalid config file being generated in some cases

jcelerier:
The problem being, that e.g. if(@PA_USE_ALSA@) expands to if() which is not valid CMake code

### DIFF
--- a/cmake/PortAudioConfig.cmake.in
+++ b/cmake/PortAudioConfig.cmake.in
@@ -8,11 +8,11 @@ include(CMakeFindDependencyMacro)
 
 find_dependency(Threads)
 
-if(@PA_USE_JACK@)
+if("@PA_USE_JACK@")
   find_dependency(Regex)
   find_dependency(JACK)
 endif()
 
-if(@PA_USE_ALSA@)
+if("@PA_USE_ALSA@")
   find_dependency(ALSA)
 endif()


jcelerier:
@philburk @Be-ing  (i don't know if there's a CMake team that I could use to ping you folks directly ;p)

also: happy new year !

Be-ing:
>  if(@PA_USE_ALSA@) expands to if()

I presume this occurs when PA_USE_ALSA is not defined? If so, maybe we should explicitly define it as `OFF` instead of this unintuitive trick?

jcelerier:
well, I don't know if it's unintuitive - I checked on my hard drive with a very generic `rg 'if.*@'' -g '*.cmake.in'`and the vast majority of the matches seemed to quote variables in this case (as according to cmake rules, the only strings that evaluate to a true value are "1", "ON", "TRUE" and the ilk. So maybe having that can be in addition ? 
(or, even better: 

     set(PA_USE_ALSA "@PA_USE_ALSA@")

which will render as 

     set(PA_USE_ALSA "ON/OFF")

and be much more clear if e.g. someone wants to quickly check how a linux distro built the portaudio package

Be-ing:
I don't especially care how the problem is solved. What matters is that it is solved. So don't wait on me to merge this.

PR #686, opened 2022-01-16
cmake: Assign library .so version for Linux statically as 2.

kleinerm:
I am no library versioning expert, but using ${PROJECT_VERSION} as
SOVERSION in the CMakeLists.txt file builds a dynamic libportaudio.so
with a SOVERSION (cfe. `objdump -p libportaudio.so | grep SONAME`) of
libportaudio.so.19.8.

This means that any executable compiled/linked against libportaudio.so
will encode a runtime dependency on libportaudio.so.19.8, whereas for
older v19.7 portaudio and earlier, the dependency was on libportaudio.so.2.

That means that any future application binary will have a dependency on
the specific portaudio version it was built against, even though it would
be api and abi compatible to other or older versions of libportaudio, so
upgrades of libportaudio.so to other versions become impossible without
rebuilding all applications that use libportaudio on that machine, or
playing tedious hacky games with symbolic links.

Hardcode the SOVERSION to 2, so the future libportaudio v19.8 will stay
compatible with applications built against v19.7 and earlier. I am not
sure if the old autotools/automake style build scripts also need a similar
update, as i only use the new cmake system.

My understanding is that the soversion should only be incremented if
libportaudio receives backwards incompatible abi changes, not on each
minor version release.

Signed-off-by: Mario Kleiner <mario.kleiner.de@gmail.com>


Be-ing:
duplicate of #621

philburk:
This is not the same as #621.
This one sets SOVERSION to 2

#621 set SOVERSION and VERSION to 2.0.0

It does seem something is broken here. Which PR is the best?

Be-ing:
I think `2`, but I'd prefer to keep discussion on one PR (#621).

kleinerm:
If i run `objdump -p /usr/lib/x86_64-linux-gnu/libportaudio.so | grep SONAME` against the libportaudio.so v19.6 shipping on my Ubuntu 20.04.3-LTS system, the output is:

  SONAME               libportaudio.so.2

Same against v19.7. That's why my pull uses SOVERSION 2, so a new libportaudio will also have exactly the same SONAME as past ones, and so software compiled/linked against such a portaudio from master will encode exactly the same dependency.

Iow. new libportaudio can be used as drop-in replacement without need for rebuilding all dependent applications, which is imho the right thing to do, given there aren't any abi incompatible changes in upcoming 19.8 vs. 19.7/19.6/... so far.

This is btw. how i realized the breakage. I built my sound driver against the wip Pulseaudio host api branch from illuso, to prepare for it already and then my new driver stopped working on any system without illuso's Pulseaudio enabled Portaudio.

Using the SONAME from this pull request fixed the compatibility breakage.

RossBencina (approved):
Based on the discussion this looks like the right solution to me. But my expertise on SO versioning is nil.

philburk (approved):
It seems safest to set this back to "2" for better compatibility.

PR #672, opened 2021-11-25
Add WASAPI loopback implementation

WeekendWarrior1:
Re implements this patch: https://github.com/audacity/audacity/commit/be7467c0e32c397d6223f2dc844a585d079b9875#diff-6ea9c8ba5ecef609fc4cfa9fbbb1bc11ad66acb997565df06ad6ce1562a10144 to add loopback for WASAPI devices.

I'm very new to C so I welcome any feedback or improvements to this PR.

Closes this:
PortAudio/portaudio#668

And would really help python-sounddevice users:
spatialaudio/portaudio-binaries#6
spatialaudio/python-sounddevice#287
https://github.com/tlecomte/friture/issues/64
https://github.com/spatialaudio/python-sounddevice/issues/281
spatialaudio/python-sounddevice#14 

Also see this project which maintains a fork of portaudio with loopback support specifically for PyAudio: https://github.com/intxcc/pyaudio_portaudio

PR #688, opened 2022-02-08
Fixed wrong default WASAPI device selection due to newly added loopback code.

dmitrykos:
Fixes problem noticed in #672 by @bear101.


bear101:
This fixes the issue, thanks!

PR #691, opened 2022-02-15
Mac: change some internal errors to host errors.

### LINKED ISSUE #620: Mac Core returns paInternalError instead of paUnanticipatedHostError
**Describe the bug**
If some error occures in a coreaudio function, it always returns `paInternalError` that make troubleshooting too difficult.

But the actual error is also stored with with `PaUtil_SetLastHostErrorInfo`. It contradicts the specification of `Pa_GetLastHostErrorInfo` in https://github.com/PortAudio/portaudio/blob/master/include/portaudio.h

> The values in this structure will only be valid if a PortAudio function has previously returned the paUnanticipatedHostError  error code.

**To Reproduce**
Every time when some error occures in portaudio fuctions using CoreAudio.

**Expected behavior**
Return `paUnanticipatedHostError` and store the acttual information to `LastHostErrorInfo`.

**Actual behavior**
Returning `paInternalError`. `Pa_GetLastHostErrorInfo` returns the actual information.

 - OS: Mac OS
 - OS Version: 10.15
 - PortAudio version: any

**Additional context**
Seems the problem is in https://github.com/PortAudio/portaudio/blob/master/src/hostapi/coreaudio/pa_mac_core_utilities.c, PaMacCore_SetError:

I'd suggest to replace ``result = paInternalError;`` with ``result = paUnanticipatedHostError;`` all times in this function.

philburk:
Fixes #620

### DIFF
--- a/src/hostapi/coreaudio/pa_mac_core_utilities.c
+++ b/src/hostapi/coreaudio/pa_mac_core_utilities.c
@@ -222,7 +222,7 @@ PaError PaMacCore_SetError(OSStatus error, int line, int isError)
         break;
     case kAudioHardwareBadStreamError:
         errorText = "Audio Hardware: BadStream";
-        result = paBadStreamPtr;
+        result = paUnanticipatedHostError;
         break;
     case kAudioHardwareUnsupportedOperationError:
         errorText = "Audio Hardware: Unsupported Operation";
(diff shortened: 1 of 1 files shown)


RossBencina (approved):

philburk:
This was merged despite the CI errors because they are on Windows and unrelated to this change.

PR #694, opened 2022-02-22
ASIO Link Pro driver doesn't like the debugger

jjYBdx4IL:
Blacklist "ASIO Link Pro" ASIO driver.


RossBencina:
Who manufactures this "ASIO Link Pro" driver?

RossBencina:
We blacklist devices when they crash the application in release mode. We don't blacklist drivers just because they are not compatible with debuggers. Please provide further information about the problem this is trying to solve.

philburk:
@jjYBdx4IL - Thanks for the code. If this is a fix for a problem then it would help us if you filed an Issue.
It should explain what bad things happen if the driver is used.
Then we can understand why this fix is necessary.

Issue #595, opened 2021-06-07
Error -9998 on all tests MacOS


mkhaa830:
(Please use the mailing list for support requests and general discussion. This is only for actual bugs.)

**Describe the bug**
All tests in bin fail with the output 

Ring modulate input for 20 seconds.
An error occured while using portaudio
 Error number: -9998
 Error text: Invalid number of channels
bye

**To Reproduce**
Steps to reproduce the behavior. Include code if applicable.
1. cd bin
2. ./patest1

**Expected behavior**
The test should run and produce some sort of output, alongside all other tests in bin.

**Actual behavior**
The test exits with error code -9998

**Desktop (please complete the following information):**
 - OS: MacOS
 - OS Version 11.3.1
 - PortAudio version: stable

RossBencina:
There are two types of programs to fix here: (1) examples, (2) tests. Examples could be kept simple (e.g. assume mono input). Tests could be more adaptive and support the available channel count -- ie work with both mono or stereo input.

philburk:
I will review the tests and examples to see which are broken and make appropriate changes.

RossBencina:
@philburk  is this issue resolved now?

janstary:
I would like to run these tests on a macOS 14.5, but the test/README.txt does not say how to compile and run them.
The ./configure script does not create test/Makefile and README.txt points to http://www.portaudio.com/trac/wiki/TestPlan
which does not exist.

[Fixed: README.txt changed.]

janstary:
(However, all the tests "should compile without any warnings on GCC 3.3".)

janstary:
Ah, they end up in bin/ - sorry for missing that.
On this MacOS 14.5 (on Macbook Air M1), all the test in bin/ seem to run.

./patest_wire still reports
```
Error number: -9998
Error message: Invalid number of channels
```
in Configuration #4.

The final blip in patest_stop_playout cannot be heard.

philburk:
Remaining work is to document what are "all MacOS tests" and update the test doc.
And then make sure they all work.
(I think they do.)

philburk:
Doxygen can be used to make a list.
Eg. https://files.portaudio.com/docs/v19-doxydocs-dev/group__test__src.html

Tag is: "@ingroup test_src"

philburk:
I believe this original bug is now fixed by #1000 and other PRs.

We are still seeing some Mac problems but they are documented in #742 and other Issues.

PR #706, opened 2022-05-03
patest1: use mono input

philburk:
Avoid error when requesting stereo input.

Also removed confusing stop after 20 seconds
that conflicted with stopping by key press.

For #595

### DIFF
--- a/test/patest1.c
+++ b/test/patest1.c
@@ -49,13 +49,15 @@
 #define M_PI (3.14159265)
 #endif
 
-#define SAMPLE_RATE (44100)
+#define SAMPLE_RATE        (44100)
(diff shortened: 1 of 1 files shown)


RossBencina:
@being: could you please take a look at the cmake CI build failures on this PR. I don't really understand what's going on. We really can't ignore CI failures.

Be-ing:
> @Be-ing: could you please take a look at the cmake CI build failures on this PR. I don't really understand what's going on. We really can't ignore CI failures.

Fixed in #707. CI was checking out an old commit of vcpkg which led to it trying to download an old version of winpthreads, but it seems winpthreads doesn't have that old version on their server anymore.

RossBencina (approved):

PR #707, opened 2022-05-03
GH Actions: update vcpkg

Be-ing:
This fixes the HTTP 404 errors from vcpkg trying to download an old version of winpthread.


RossBencina:
@Be-ing I'm going to merge this, but is there any way to always use the latest version so that we don't have this problem in future?

Be-ing:
With the lukka/run-vcpkg action, I'm not sure. It would be possible to clone the master branch of https://github.com/microsoft/vcpkg.git to whatever its latest state is. That would have the risk of the opposite problem though; CI could fail because something in vcpkg broke.

RossBencina:
Thanks for the quick turnaround. #706 is passing now.

I think failing because the latest vcpkg is broken is preferable to failing because an old version broke. I think we can rely on the latest version being fixed if it's broken. But if an old version breaks, it will stay broken and we'll have to update our repo again.

Issue #708, opened 2022-05-03
Compiler errors for unused but set variables on MacOS Monterey


TobiasGrothmann:
**Describe the bug**
I'm using a fairly new M1-Pro on MacOS Monterey (12.3.1)

I downloaded portaudio (`pa_stable_v190700_20210406.tgz`) from `http://files.portaudio.com/download.html`.
Running `./configure` worked fine but `make` gave me the following (let's call it) error:
```
> make
/bin/sh ./libtool --mode=compile gcc -c -std=c99 -O2 -Wall -pedantic -pipe -fPIC -DNDEBUG -DPA_LITTLE_ENDIAN -I./include -I./src/common -I./src/os/unix -Wno-deprecated -Werror -arch x86_64 -arch arm64 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.3.sdk -mmacosx-version-min=10.6  -DPACKAGE_NAME=\"\" -DPACKAGE_TARNAME=\"\" -DPACKAGE_VERSION=\"\" -DPACKAGE_STRING=\"\" -DPACKAGE_BUGREPORT=\"\" -DPACKAGE_URL=\"\" -DSTDC_HEADERS=1 -DHAVE_SYS_TYPES_H=1 -DHAVE_SYS_STAT_H=1 -DHAVE_STDLIB_H=1 -DHAVE_STRING_H=1 -DHAVE_MEMORY_H=1 -DHAVE_STRINGS_H=1 -DHAVE_INTTYPES_H=1 -DHAVE_STDINT_H=1 -DHAVE_UNISTD_H=1 -DHAVE_DLFCN_H=1 -DLT_OBJDIR=\".libs/\" -DSIZEOF_SHORT=2 -DSIZEOF_INT=4 -DSIZEOF_LONG=8 -DHAVE_CLOCK_GETTIME=1 -DHAVE_NANOSLEEP=1 -DPA_USE_COREAUDIO=1 -Wno-unused-variable src/common/pa_converters.c -o src/common/pa_converters.lo
[...]

PR #712, opened 2022-05-25
fix errors for unused but set variables on mac

TobiasGrothmann:
see issue: #708

### DIFF
--- a/src/hostapi/coreaudio/pa_mac_core_blocking.c
+++ b/src/hostapi/coreaudio/pa_mac_core_blocking.c
@@ -195,6 +195,7 @@ PaError initializeBlioRingBuffers(
                 data );
         assert( !err );
     }
+    (void) err; /* Prevent "unused variable" warnings. */
 
     result = resetBlioRingBuffers( blio );
     if( result )
@@ -372,6 +373,7 @@ int BlioCallback( const void *input, void *output, unsigned long frameCount,
         /*printf( "reading %d\n", toRead );*/
         framesTransferred = PaUtil_WriteRingBuffer( &blio->inputRingBuffer, input, framesToTransfer );
         assert( framesToTransfer == framesTransferred );
+        (void) framesTransferred; /* Prevent "unused variable" warnings. */
 #ifdef PA_MAC__BLIO_MUTEX
         /* Priority inversion. See notes below. */
         blioSetIsInputEmpty( blio, false );
--- a/aclocal.m4
+++ b/aclocal.m4
@@ -1202,7 +1202,7 @@ fi
 # Invoke $ECHO with all args, space-separated.
 func_echo_all ()
 {
-    $ECHO "$*" 
+    $ECHO "$*"
 }
 
 case "$ECHO" in
@@ -8542,7 +8542,7 @@ _LT_DECL(, macro_revision, 0)
 # included after everything else.  This provides aclocal with the
 # AC_DEFUNs it wants, but when m4 processes it, it doesn't do anything
 # because those macros already exist, or will be overwritten later.
-# We use AC_DEFUN over AU_DEFUN for compatibility with aclocal-1.6. 
+# We use AC_DEFUN over AU_DEFUN for compatibility with aclocal-1.6.
 #
 # Anytime we withdraw an AC_DEFUN or AU_DEFUN, remember to add it here.
 # Yes, that means every name once taken will need to remain here until
@@ -8616,7 +8616,7 @@ m4_ifndef([_LT_PROG_CXX],		[AC_DEFUN([_LT_PROG_CXX])])
 
 # pkg.m4 - Macros to locate and utilise pkg-config.            -*- Autoconf -*-
 # serial 1 (pkg-config-0.24)
-# 
+#
 # Copyright © 2004 Scott James Remnant <scott@netsplit.com>.
 #
 # This program is free software; you can redistribute it and/or modify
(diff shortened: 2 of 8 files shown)


src/hostapi/coreaudio/pa_mac_core_blocking.c
```diff
@@ -62,7 +62,14 @@
 
 #include "pa_mac_core_blocking.h"
 #include "pa_mac_core_internal.h"
-#include <assert.h>
+
+#ifdef NDEBUG
+# undef assert
+# define assert(VAL) do { (void)sizeof(VAL); } while(0)
```
RossBencina, line 68: Thanks for this. The use of `sizeof` here risks not evaluating side effects of VAL. (Of course well written code would have no side effects in an assert or debug print, but we should not assume that.) Did you try the following instead:

` # define assert(VAL) do { (void)(VAL); } while(0)`

If not, please try it, it's preferable if it works. Same for the other DBUG macros.

TobiasGrothmann: I have tried `(void)(VAL)`, but I went for sizeof because I got `expression result unused` errors with `VDBUG` and `VVDBUG`:

```
src/hostapi/coreaudio/pa_mac_core.c:357:14: error: expression result unused [-Werror,-Wunused-value]
    VDBUG( ( "Found %ld device(s).\n", auhalHostApi->devCount ) );
             ^~~~~~~~~~~~~~~~~~~~~~~~
src/hostapi/coreaudio/pa_mac_core_utilities.h:101:33: note: expanded from macro 'VDBUG'
# define VDBUG(MSG) do { (void)(MSG); } while(0)
                                ^~~
```
```
src/hostapi/coreaudio/pa_mac_core.c:1193:13: error: expression result unused [-Werror,-Wunused-value]
    VVDBUG(("OpenAndSetupOneAudioUnit(): in chan=%d, in fmt=%ld, out chan=%d, out fmt=%ld, requestedFramesPerBuffer=%ld\n",
            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/hostapi/coreaudio/pa_mac_core_utilities.h:111:34: note: expanded from macro 'VVDBUG'
# define VVDBUG(MSG) do { (void)(MSG); } while(0)
                                 ^~~

```
```
src/hostapi/coreaudio/pa_mac_core.c:1194:63: error: expression result unused [-Werror,-Wunused-value]
            inStreamParams  ? inStreamParams->channelCount  : -1,
                              ~~~~~~~~~~~~~~                  ^~
src/hostapi/coreaudio/pa_mac_core_utilities.h:111:34: note: expanded from macro 'VVDBUG'
# define VVDBUG(MSG) do { (void)(MSG); } while(0)
                                 ^~~
```

RossBencina: Alright. In addition to the fact that you've provided evidence that `sizeof` is needed to provide a valid fix for the compiler warnings, I thought about this some more and realised that I was wrong to argue for removing `sizeof`: The old code did not evaluate `VAL` in the no-debug case, It used an empty define, e.g,

`# define DBUG(MSG)`

So your patch that uses `sizeof` is entirely equivalent to the current behavior of those macros. I approve.

RossBencina (approved):

RossBencina:
@philburk could you review this please.

RossBencina:
Sorry for the delay. @philburk has his hands full right now. I'd still like to wait and get his review before merging this.

A note: I know that I suggested it, but it now occurs to me that redefining the `assert` macro is bad style. The fix moves things forward, so we'll likely accept it, but I'm not sure we want to keep the `assert` part of the fix in the long term. I think it would be better style for the code that uses `assert` to perform any assert-specific declarations/assignments/computations guarded by `#ifndef NDEBUG` -- however that will require reviewing each `assert` call site in detail.

TobiasGrothmann:
@RossBencina, could you please have another look at the changes? I force pushed and replaced the commit with a (imho) cleaner one.

I just defined a macro `#define IGNORE_UNUSED(VAR) (void) VAR` that can be used in all of these cases.

RossBencina:
Two requests:

1. Please replace `IGNORE_UNUSED` macro with explicit casts to `void`. Cast to `void` is a standard idiom and is clearer than a macro that you'd need to go and look up to see how it is defined.

2. Please move the casts to `void` to directly below each place that the variable is used (e.g. after the `assert` or debug print statement). This ensures that we're telling the compiler that the *value* that was assigned to the variable is used, not just the variable name.

TobiasGrothmann:
> 1. Please replace `IGNORE_UNUSED` macro with explicit casts to `void`. Cast to `void` is a standard idiom and is clearer than a macro that you'd need to go and look up to see how it is defined.

That makes absolute sense. I personally just didn't know that was a thing. Also added a comment like everywhere else.
 
> 2. Please move the casts to `void` to directly below each place that the variable is used (e.g. after the `assert` or debug print statement). This ensures that we're telling the compiler that the _value_ that was assigned to the variable is used, not just the variable name.

Also did that. Sometimes the variables get used multiple times, though.

philburk:
This PR had some unrelated commits in it. So I cherry-picked Tobias' commit to a new branch and am merging it in #722.

Thanks Tobias!

PR #718, opened 2022-06-21
Update compile_windows_mingw-w64.dox

ad8e:
The sawtooth example is absurdly loud. If it's left as part of the default build steps, it will probably cause some hearing damage. This moves it into a separate section with a warning.

An alternative fix would be to lower the sawtooth example to a reasonable volume; after all, if someone compiles it and runs it for another reason, it will have the same problem. Though, many of the examples are probably no different.


philburk (approved):

RossBencina (approved):
I like this improvement to the docs. I wonder whether a similar change can be made for the other build instructions?

PR #722, opened 2022-07-12
fix errors for unused but set variables on mac

philburk:


RossBencina (approved):
Approved assuming it passes CI.

PR #729, opened 2022-09-05
alsa: remove asserts based on maxChans

### LINKED ISSUE #594: Assertion `maxChans <= INT_MAX' failed.
Problem with portaudio:
    src/hostapi/alsa/pa_linux_alsa.c:899: GropeDevice: Assertion `maxChans <= INT_MAX' failed.

   how to solve？

philburk:
Return an error if outside reasonable range.

Fixes #594

### DIFF
--- a/src/hostapi/alsa/pa_linux_alsa.c
+++ b/src/hostapi/alsa/pa_linux_alsa.c
@@ -909,7 +909,7 @@ static PaError GropeDevice( snd_pcm_t* pcm, int isPlug, StreamDirection mode, in
         result = paUnanticipatedHostError;
         goto error;
     }
-    if( maxChans == 0 || minChans == 0)
+    else if( maxChans == 0 )
     {
         /* Weird linking issue could cause wrong version of ALSA symbols to be called,
                                    resulting in zeroed values */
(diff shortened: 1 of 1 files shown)


src/hostapi/alsa/pa_linux_alsa.c
```diff
@@ -901,9 +902,22 @@ static PaError GropeDevice( snd_pcm_t* pcm, int isPlug, StreamDirection mode, in
...
-    assert( maxChans <= INT_MAX );
-    assert( maxChans > 0 );    /* Weird linking issue could cause wrong version of ALSA symbols to be called,
+    const unsigned int kReasonableMaxChannels = 1024;
+    if( maxChans > kReasonableMaxChannels )
+    {
+        PA_DEBUG(( "%s: maxChans = %u, which is unreasonably high\n", __FUNCTION__, maxChans ));
+        result = paUnanticipatedHostError;
+        goto error;
+    }
+    if( maxChans == 0 || minChans == 0)
```
philburk, line 912: Remove check for minChans.

RossBencina (approved):
This is a good stepwise improvement. I'm ok with merging (Squashing) this. I do wonder whether the message should also be printed to the console in non-debug cases. But I guess we don't do that anywhere else. And there are always complaints about PortAudio being noisy with console output.

PR #752, opened 2022-11-16
Fix for WDM KS WaveRT low-latency microphone capture quality

feoff3:
In my tests with latency < 40 ms on Win10 with embeded Realtek sound card, WDM KS capture resulted in sound with lots of cracks. 
I digged into the code and found out 1) there is a wrong assumption that host buffer position should be either 0 or 1/2 of buffer size when event is handled, 2) ensured that host ring buffer is read both from the end and the beginning if necessary (the exisiting code might resulted in reading non-initialized data beyond the end of the host buffer)

### DIFF
--- a/src/hostapi/wdmks/pa_win_wdmks.c
+++ b/src/hostapi/wdmks/pa_win_wdmks.c
@@ -6587,12 +6587,11 @@ static PaError PaPinCaptureEventHandler_WaveRTEvent(PaProcessThreadInfo* pInfo, 
     PaWinWdmIOInfo* pCapture = &pInfo->stream->capture;
     const unsigned halfInputBuffer = pCapture->hostBufferSize >> 1;
     PaWinWdmPin* pin = pCapture->pPin;
-    DATAPACKET* packet = 0;
 
     /* Get hold of current ADC position */
     pin->fnAudioPosition(pin, &pos);
     pos %= pCapture->hostBufferSize;
-    pos &= ~(pCapture->bytesPerFrame - 1);
+    pos -= pos % pCapture->bytesPerFrame;
 
     /* Call barrier (or dummy) */
     pin->fnMemBarrier();
--- a/src/hostapi/wdmks/pa_win_wdmks.c
+++ b/src/hostapi/wdmks/pa_win_wdmks.c
@@ -6582,8 +6582,8 @@ static PaError PaPinRenderSubmitHandler_WaveCyclic(PaProcessThreadInfo* pInfo, u
 static PaError PaPinCaptureEventHandler_WaveRTEvent(PaProcessThreadInfo* pInfo, unsigned eventIndex)
 {
     unsigned long pos;
-    unsigned realInBuf;
-    unsigned frameCount;
+    unsigned frameCount = 0;
+    unsigned bytesToRead;
     PaWinWdmIOInfo* pCapture = &pInfo->stream->capture;
     const unsigned halfInputBuffer = pCapture->hostBufferSize >> 1;
     PaWinWdmPin* pin = pCapture->pPin;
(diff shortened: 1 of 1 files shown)


RossBencina:
Thank you for your contribution, please fix the trailing whitespace issue on line 6620 flagged by CI:
> error: src/hostapi/wdmks/pa_win_wdmks.c(6620) trailing whitespace:
> b'    '

RossBencina:
We notice that you've cut-and-pasted the same buffer copying code to two different places: `PaPinRenderSubmitHandler_WaveCyclic`, `PaPinCaptureEventHandler_WaveRTPolled` -- can you please confirm that both of these code paths have actually been tested on your machine? Copy-pasta is always a risk and I'd like to be sure that both codes are correct and tested.

feoff3:
Yeap, I was considering having just one function instead of two because
they are almost identical right now.
Let me double check if the code works fine on both path, and if it is
possible to reduce the code duplication there.

вт, 22 нояб. 2022 г. в 05:15, Ross Bencina ***@***.***>:

feoff3:
@RossBencina I have tested the SubType_kPolled type of capture code, and it produces distorted sound with ~10ms latency both with my updated code and original code. 
On the other hand, my code improves SubType_kNotification type of capture significantly. So for now, I return the original code for SubType_kPolled, and push my code for SubType_kNotification.

I leave the SubType_kPolled type as is until the further investigation.

RossBencina:
@feoff3 thank you. please fix the whitespace issue as previously requested.

RossBencina:
@feoff3 do you think this should be merged as-is or were you planning to do more work on it? If merge, we'll fix the whitespace.

feoff3:
Oh sorry, I just forgot about the whitespace - xmas and NY has swamped me
into other stuff.

Please feel free to fix the whitespace and merge ;)

dechamps:
Currently this PR only touches the `PaPinCaptureEventHandler_WaveRTEvent()` function. @feoff3 Is this because these issues don't apply to the render side, or because you didn't look at the render side at all? I'm wondering if similar issues on the render side could be causing #763.

feoff3:
Hey there.
Yes I didn't touch render side as I experienced the issue on capture side
only.

I can take a quick look on render side a little bit later... I guess the
original developer used a similar approach to the render as was used for
capture (which is prone to data loss and thus choppy sound)

src/hostapi/wdmks/pa_win_wdmks.c
```diff
@@ -6582,31 +6582,41 @@ static PaError PaPinRenderSubmitHandler_WaveCyclic(PaProcessThreadInfo* pInfo, u
...
     pin->fnAudioPosition(pin, &pos);
-    /* Wrap it (robi: why not use hw latency compensation here ?? because pos then gets _way_ off from
-    where it should be, i.e. at beginning or half buffer position. Why? No idea.)  */
-
     pos %= pCapture->hostBufferSize;
-    /* Then realInBuf will point to "other" half of double buffer */
-    realInBuf = pos < halfInputBuffer ? 1U : 0U;
-
-    packet = pInfo->stream->capture.packets + realInBuf;
+    pos &= ~(pCapture->bytesPerFrame - 1);
```
RossBencina, line 6595: This only works if `bytesPerFrame` is a power of two. But we can have packed 24 bit data or 5 channels, so this won't work as-is. Do you think the following change is good?

  `pos -= pos % pCapture->bytesPerFrame;`

src/hostapi/wdmks/pa_win_wdmks.c
```diff
@@ -6582,31 +6582,41 @@ static PaError PaPinRenderSubmitHandler_WaveCyclic(PaProcessThreadInfo* pInfo, u
...
 {
     unsigned long pos;
-    unsigned realInBuf;
-    unsigned frameCount;
+    unsigned frameCount = 0;
+    unsigned bytesToRead;
     PaWinWdmIOInfo* pCapture = &pInfo->stream->capture;
     const unsigned halfInputBuffer = pCapture->hostBufferSize >> 1;
     PaWinWdmPin* pin = pCapture->pPin;
     DATAPACKET* packet = 0;
```
RossBencina, line 6590: Is this an unused variable? If so please remove it.

RossBencina:
@feoff3 could you review our comments and commits please? It would be great if you could test the changes on your system too.

src/hostapi/wdmks/pa_win_wdmks.c
philburk: The variable is "packet".

RossBencina (changes requested):
Awaiting response to our requests for change, and review of our commits.

feoff3:
@RossBencina  @philburk  @dechamps Guys, thanks for the feedback. Give me several days to check your comments and run a couple of tests on my side.

feoff3:
sorry guys, still stuck with other tasks. gonna do it next week

src/hostapi/wdmks/pa_win_wdmks.c
feoff3: well, 24-bit is packed into 32-bit containers by Windows. Regarding 5 channels, I am not sure if it is possible.

feoff3 (commented):
yeap, makes sense

RossBencina (approved):

PR #762, opened 2023-01-21
WDM-KS: don't request a zero buffer size

### LINKED ISSUE #761: WDM-KS fails to open WaveRT devices with suggestedLatency=0
**Describe the bug**
If the following conditions are true:

- `PaStreamParameters::suggestedLatency` is set to zero;
- The device being opened is backed by the WDM-KS Host API;
- The Windows driver for the device being opened uses [WaveRT](https://learn.microsoft.com/en-us/windows-hardware/drivers/audio/wave-filters) (most modern drivers do);

Then `Pa_OpenStream()` will fail.

**To Reproduce**
1. Change `examples/paex_sine.c`:
   - Set `suggestedLatency` to zero.
   - Set `device` to the index of a WDM-KS device that happens to use WaveRT.
     - [Virtual Audio Cable](https://vac.muzychenko.net/en/) can be used to emulate a WaveRT device, as well as WavePci and WaveCyclic for comparison.
3. Run the example.

**Expected behavior**
The example runs and plays sound for 5 seconds.

**Actual behavior**
`Pa_OpenStream()` fails with the following debug output:

```
 0.072: OpenStream:sampleRate = 44100.000000
 0.072: OpenStream:framesPerBuffer = 64
 0.072: Pin create result = 0x00000000
 0.072: Render pin frames: 0
 0.072: Output frames chosen:0
 0.072: Failed to get buffer with notification
 0.073: Failed to get buffer without notification
 0.073: Buffer size on 128 byte boundary, still fails :(
 0.073: Failed to get output buffer (with notification)
TerminateHostApis in
TerminateHostApis out
An error occurred while using the portaudio stream
Error number: -9999
Error message: Unanticipated host error
```

**Desktop:**
 - OS Version: Windows 11 21H2 22000.1455, but this doesn't seem to matter (it has been [observed](https://github.com/dechamps/FlexASIO/issues/178#issuecomment-1398994952) on Windows 10 too)
 - PortAudio version: latest master (bbe2b5ad55229c13c4d7cd54b5dcd78511c70a4b), and going back to at least 80ef9ac9c4ed1a4b9f3f4b11ad31ed9f69c4dc57
 - Host API: WDM-KS

**Additional context**
- Upstream bug: dechamps/FlexASIO#159

dechamps:
This fixes an issue where Pa_OpenStream() would fail on WDM-KS WaveRT devices if suggestedLatency is zero.

Fixes #761

### DIFF
--- a/src/hostapi/wdmks/pa_win_wdmks.c
+++ b/src/hostapi/wdmks/pa_win_wdmks.c
@@ -2336,6 +2336,8 @@ static PaError PinGetBufferWithNotification(PaWinWdmPin* pPin, void** pBuffer, D
 
     propIn.BaseAddress = 0;
     propIn.NotificationCount = 2;
+    /* Note RequestedBufferSize must be >0, otherwise the Ioctl will fail.
+     * See https://github.com/PortAudio/portaudio/issues/761 */
     propIn.RequestedBufferSize = *pRequestedBufSize;
     propIn.Property.Set = KSPROPSETID_RtAudio;
     propIn.Property.Id = KSPROPERTY_RTAUDIO_BUFFER_WITH_NOTIFICATION;
@@ -2372,6 +2374,8 @@ static PaError PinGetBufferWithoutNotification(PaWinWdmPin* pPin, void** pBuffer
     PA_LOGE_;
 
     propIn.BaseAddress = NULL;
+    /* Note RequestedBufferSize must be >0, otherwise the Ioctl will fail.
+     * See https://github.com/PortAudio/portaudio/issues/761 */
     propIn.RequestedBufferSize = *pRequestedBufSize;
     propIn.Property.Set = KSPROPSETID_RtAudio;
     propIn.Property.Id = KSPROPERTY_RTAUDIO_BUFFER;
@@ -4734,6 +4738,13 @@ static PaError OpenStream( struct PaUtilHostApiRepresentation *hostApi,
         {
             stream->capture.framesPerBuffer = stream->capture.pPin->frameSize;
         }
+
+        if(stream->capture.framesPerBuffer == 0) {
+            /* WaveRT devices will reject a zero RequestedBufferSize.
+             * See https://github.com/PortAudio/portaudio/issues/761 */
+            stream->capture.framesPerBuffer = 1;
+        }
+
         PA_DEBUG(("Input frames chosen:%ld\n",stream->capture.framesPerBuffer));
 
         /* Setup number of packets to use */
@@ -4763,6 +4774,13 @@ static PaError OpenStream( struct PaUtilHostApiRepresentation *hostApi,
         {
             stream->render.framesPerBuffer = stream->render.pPin->frameSize;
         }
+
+        if(stream->render.framesPerBuffer == 0) {
+            /* WaveRT devices will reject a zero RequestedBufferSize.
+             * See https://github.com/PortAudio/portaudio/issues/761 */
+            stream->render.framesPerBuffer = 1;
+        }
+
         PA_DEBUG(("Output frames chosen:%ld\n",stream->render.framesPerBuffer));
 
         /* Setup number of packets to use */


src/hostapi/wdmks/pa_win_wdmks.c
```diff
@@ -4763,6 +4767,10 @@ static PaError OpenStream( struct PaUtilHostApiRepresentation *hostApi,
         {
             stream->render.framesPerBuffer = stream->render.pPin->frameSize;
         }
+        else if(stream->render.framesPerBuffer == 0) {
```
philburk, line 4770: If framesPerBuffer==0 AND pPin->frameSize==0 then this clamping logic will get skipped.
It may be safer to not use an else and just do:

    if(stream->render.framesPerBuffer == 0) {

src/hostapi/wdmks/pa_win_wdmks.c
```diff
@@ -4734,6 +4734,10 @@ static PaError OpenStream( struct PaUtilHostApiRepresentation *hostApi,
         {
             stream->capture.framesPerBuffer = stream->capture.pPin->frameSize;
         }
+        else if(stream->capture.framesPerBuffer == 0) {
```
philburk, line 4737: See comment on line 4771.

philburk (changes requested):

RossBencina:
If this is a workaround for `KSRTAUDIO_BUFFER_PROPERTY/RequestedBufferSize` not accepting zero, then I think that the the clamp to 1 should be performed just prior to the KSRTAUDIO_BUFFER_PROPERTY command rather than the current patch which reads to me like some kind of long-range action at a distance. Just a suggestion/opinion though. I'm open to any alternate reasoning that would reject my perspective.

I've made other comments at #761, we should try to keep the bulk of the discussion there.

src/hostapi/wdmks/pa_win_wdmks.c
```diff
@@ -4734,6 +4734,12 @@ static PaError OpenStream( struct PaUtilHostApiRepresentation *hostApi,
         {
             stream->capture.framesPerBuffer = stream->capture.pPin->frameSize;
         }
+
+        if(stream->capture.framesPerBuffer == 0) {
+            /* WinRT devices will reject a zero RequestedBufferSize. See https://github.com/PortAudio/portaudio/issues/761 */
```
philburk, line 4739: Split long line.
```suggestion
            /* WinRT devices will reject a zero RequestedBufferSize. 
             * See https://github.com/PortAudio/portaudio/issues/761 */
```

src/hostapi/wdmks/pa_win_wdmks.c
```diff
@@ -4763,6 +4769,12 @@ static PaError OpenStream( struct PaUtilHostApiRepresentation *hostApi,
         {
             stream->render.framesPerBuffer = stream->render.pPin->frameSize;
         }
+
+        if(stream->render.framesPerBuffer == 0) {
+            /* WinRT devices will reject a zero RequestedBufferSize. See https://github.com/PortAudio/portaudio/issues/761 */
```
philburk, line 4774:
```suggestion
            /* WinRT devices will reject a zero RequestedBufferSize.
             * See https://github.com/PortAudio/portaudio/issues/761 */
```

philburk (changes requested):
Looks fine except for long lines.

RossBencina:
@dechamps I'm not completely persuaded by your reasoning because I think my solution guarantees that 0 will not ever be passed to KSRTAUDIO_BUFFER_PROPERTY/RequestedBufferSize whereas your solution relies on maintaining an undocumented invariant "at a distance".

Could we at least dd a comment in the code directly prior to where KSRTAUDIO_BUFFER_PROPERTY/RequestedBufferSize is set noting that passing 0 will fail with WinRT devices (do you mean WAVERT? not WinRT?) and linking to #761

Other solutions would be:
- add an assert !=0 at RequestedBufferSize site (don't do this!)
- put the clamp at both places.

dechamps:
@RossBencina 

> Could we at least add a comment in the code directly prior to where KSRTAUDIO_BUFFER_PROPERTY/RequestedBufferSize is set noting that passing 0 will fail with WinRT devices (do you mean WAVERT? not WinRT?) and linking to https://github.com/PortAudio/portaudio/issues/761

Done.

> do you mean WAVERT? not WinRT?

Oh wow, good catch. I don't know why I wrote WinRT literally everywhere including in my original bug report, what a brain glitch… fixed in all places.

dechamps:
The two failing checks are not caused by this PR. Filed #767 to fix the checks.

philburk (approved):

RossBencina (approved):
Thankyou!

dechamps:
I guess this can be merged now?

philburk (approved):

PR #765, opened 2023-01-31
Remove usage of global BUILD_SHARED_LIBS variable

bheisch:
Its not recommended to use a global settings in an cmake library, because this variable forces (when portaudio is used as an insource dependency)  to build every library as an shared or static variabe, which is not preffered (at least thats my guess, because it limits the usage from the library) In my proposal this can be easily fixed, because an local varirable instead of an globalone is used.)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,11 +5,18 @@ project(PortAudio VERSION 19.8)
 # General PortAudio stuff
 #
 
-option(BUILD_SHARED_LIBS "Build dynamic library" ON)
+option(PA_BUILD_SHARED_LIBS "Build dynamic library" ON)
 option(PA_BUILD_TESTS "Include test projects" OFF)
 option(PA_BUILD_EXAMPLES "Include example projects" OFF)
 
+if(PA_BUILD_SHARED_LIBS)
+  set(LIBRARY_BUILD_TYPE SHARED)
+else()
+  set(LIBRARY_BUILD_TYPE STATIC)
+endif()
+
 add_library(PortAudio
+  ${LIBRARY_BUILD_TYPE}
   src/common/pa_allocation.c
   src/common/pa_allocation.h
   src/common/pa_converters.c
@@ -65,7 +72,7 @@ else()
   target_compile_definitions(PortAudio PRIVATE PA_LITTLE_ENDIAN)
 endif()
 
-if(WIN32 AND MSVC AND BUILD_SHARED_LIBS
+if(WIN32 AND MSVC AND PA_BUILD_SHARED_LIBS
   # Check if the user is building PortAudio stand-alone or as part of a larger
   # project. If this is part of a larger project (i.e. the CMakeLists.txt has
   # been imported by some other CMakeLists.txt), we don't want to override
@@ -243,7 +250,7 @@ if(WIN32)
     target_compile_definitions(PortAudio PRIVATE PAWIN_USE_WDMKS_DEVICE_INFO)
   endif()
 
-  if(BUILD_SHARED_LIBS)
+  if(PA_BUILD_SHARED_LIBS)
     configure_file(cmake/portaudio.def.in "${CMAKE_CURRENT_BINARY_DIR}/portaudio.def" @ONLY)
     target_sources(PortAudio PRIVATE "${CMAKE_CURRENT_BINARY_DIR}/portaudio.def")
   endif()
@@ -406,7 +413,7 @@ install(TARGETS PortAudio
 
 # Some of the tests and examples use private symbols which are not
 # exposed by the .def file on Windows.
-if(WIN32 AND BUILD_SHARED_LIBS)
+if(WIN32 AND PA_BUILD_SHARED_LIBS)
   set(LINK_PRIVATE_SYMBOLS OFF)
 else()
   set(LINK_PRIVATE_SYMBOLS ON)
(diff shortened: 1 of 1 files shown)


RossBencina:
Thanks for this. Would be great to hear from some other cmake users on this. I don't have any idea about this.

Be-ing:
I disagree with this. If this is merged, users will need to look up a project specific variable instead of using the standard way.

Be-ing:
> this variable forces (when portaudio is used as an insource dependency) to build every library as an shared or static variabe

If you really want some libraries used this way linked statically and others linked dynamically, you could `set(BUILD_SHARED_LIBS OFF)` before `include` and `set(BUILD_SHARED_LIBS ON)` after.

bheisch:
mmh, i think i get your point, because it may be handier if you build from commandline. But on the otherhand, you already have multiple PortAudio specific settings and its also common practice to give a meaningfull prefix to projectspecific Variables. Like PA_BUILD_TESTS (just an example) because you mostly want to control the specific behaviour of a project. 
Also if you include PortAudio this way, PortAudio will change the settings of the executable, which needs PortAudio. This makes debugging hard(because you would normaly not suspect an library to change global cmake settings. (you would also not change CMAKE_PROJECT_SOURCE_DIR as an example))
But maybe we can make an compromise, when we use the BUILD_SHARED_LIBS variable as the default initial parameter. 
This means you can still set that parameter and PortAudio would behave the same, but you also can overwrite the project specific PortAudio setting.

daschuer:
I think the cmake documentation gives some hints: https://cmake.org/cmake/help/latest/command/add_library.html
> If no type is given explicitly the type is STATIC or SHARED based on whether the current value of the variable [BUILD_SHARED_LIBS](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html#variable:BUILD_SHARED_LIBS) is ON.

https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html
> This variable is often added to projects as an [option()](https://cmake.org/cmake/help/latest/command/option.html#command:option) so that each user of a project can decide if they want to build the project using shared or static libraries.

VCPKG for instance makes use of it and configures the BUILD_SHARED_LIBS https://github.com/microsoft/vcpkg/blob/master/scripts/cmake/vcpkg_configure_cmake.cmake#L192-L196 before using the library specific CMakeLists.txt. 
This works for all libraries, but would require a special case handling after merging this PR.

bheisch:
@daschuer Thats statement in the documentation is absolutly true. But i still think, that my points mentionend before are still valid. 
Also since the given compromise would not introduce any problems, since the initial Behaviour should the same. 

I mean in some C code you wouldnt also be so happy if you use an global variable in your main exe and some library just overwrites it. Thats an nasty Problem, which is hard to figure out. 

And of cause this analogy is  only valid if we get this as insource dependency.

bheisch:
to be honest guys, to activate BUILD_SHARED_LIBS did at least solve my Problem. I wanted just to remove cumbersome bugfixing for other users. 
But if you are realy uncomfortable with my solution or with my compromise, we can also close this PR.

PS:.
I realy like PortAudio and thanks for all your good work (doesnt matter to me, what we do in the end :D )

daschuer:
I probably have not fully understand your issue, but how about make your new variable optional? 
If it is not defined, behave like before.

bheisch:
sure this would also go, or the default variable gets the default value from BUILD_SHARED_LIBS, this way port audio with a passed BUILD_SHARED_LIBS variable would build like always, and if you dont want that or want to override only PortAudio behaviour for shared libraries you can just specify PA_BUILD_SHARED_LIBS.
In my understanding you have the best of both worlds then.

RossBencina:
@bheisch I have a question about your latest proposal: do any other libraries take this approach?

philburk:
> or the default variable gets the default value from BUILD_SHARED_LIBS, 

@bheisch - If it is easy, could you modify this PR to reflect this latest proposal?
It would then be easier to evaluate.

bheisch:
@philburk  sure i make an approach .
@RossBencina to be honest, this is the first time, i see an library to set this variable and i tried the workaround, which was presented. But i still run into Problems.
This Variables works only well if you dont care for insource builds, where you include the Source directly. 
So of cause i can look around through the Internet to find opensource libraries which doesnt use this variable, or (im sure) the exactly opposite. 
The Question is, if you want to support insource builds (add portaudio as a submodule and build it along to your application) this variable cause serious issues.

bheisch:
@philburk i added the initial Parameter to BUILD_SHARED_LIBS, which means  if the initial value is given BUILD_SHARED_LIBS it will be treated with respect to this value, but if not provided you can control the Buildstatus of PortAudio via PA_BUILD_SHARED_LIBRARY.
This way portaudio doesnt change something in the application, which uses it

RossBencina:
@Be-ing @daschuer @dechamps any objections to the latest version of this patch? if no objections we're going to merge it.

philburk (approved):

RossBencina (approved):

dechamps:
FYI it looks like this patch caused the default for `PA_BUILD_SHARED_LIBS` to flip from `ON` to `OFF`, which surprised me - I ended up accidentally releasing a version of my project with PortAudio statically linked instead of bundling it as a separate DLL. (See dechamps/FlexASIO@42e683b14b1e6398d4ceb5d4558b0d6da03ad73e.) Thankfully that's benign in my case, but it might not be for others. Was the change of default value intended?

RossBencina:
> FYI it looks like this patch caused the default for PA_BUILD_SHARED_LIBS to flip from ON to OFF

Given that this PR was merged, and no one is monitoring merged and closed issues/PRs I have created a new issue to track this: #918.

PR #769, opened 2023-02-18
Improve paloopback test params handling

dechamps:
This modernizes paloopback code around TestParameters handling by merging PaQa_SetDefaultTestParameters() and PaQa_OverrideTestParameters(), and using proper scoping instead of error-prone "change this param and then restore it later" patterns.

As a nice side effect, this also fixes a bug where user options would be ignored during the "Test Sample Formats using Half Duplex IO" phase because a PaQa_SetDefaultTestParameters() call was made without being followed by PaQa_OverrideTestParameters(), hence demonstrating why it's a good idea to merge these two functions.

### DIFF
--- a/qa/loopback/src/paqa.c
+++ b/qa/loopback/src/paqa.c
@@ -909,7 +909,7 @@ error:
 }
 
 /*******************************************************************/
-static void PaQa_SetDefaultTestParameters( TestParameters *testParamsPtr, PaDeviceIndex inputDevice, PaDeviceIndex outputDevice )
+static void PaQa_MakeTestParameters( TestParameters *testParamsPtr, PaDeviceIndex inputDevice, PaDeviceIndex outputDevice, const UserOptions *userOptions )
 {
     memset( testParamsPtr, 0, sizeof(TestParameters) );
 
@@ -933,11 +933,7 @@ static void PaQa_SetDefaultTestParameters( TestParameters *testParamsPtr, PaDevi
     testParamsPtr->outputParameters.channelCount = testParamsPtr->samplesPerFrame;
     testParamsPtr->outputParameters.suggestedLatency = Pa_GetDeviceInfo( outputDevice )->defaultLowOutputLatency;
     //testParamsPtr->outputParameters.suggestedLatency = Pa_GetDeviceInfo( outputDevice )->defaultHighOutputLatency;
-}
 
-/*******************************************************************/
-static void PaQa_OverrideTestParameters( TestParameters *testParamsPtr,  UserOptions *userOptions )
-{
     // Check to see if a specific value was requested.
     if( userOptions->sampleRate >= 0 )
     {
@@ -996,18 +992,17 @@ static int PaQa_AnalyzeLoopbackConnection( UserOptions *userOptions, PaDeviceInd
     printf( "=============== Analysing Loopback %d to %d =====================\n", outputDevice, inputDevice  );
     printf( "   Devices: %s => %s\n", outputDeviceInfo->name, inputDeviceInfo->name);
 
-    PaQa_SetDefaultTestParameters( &testParams, inputDevice, outputDevice );
-
-    PaQa_OverrideTestParameters( &testParams, userOptions );
+    PaQa_MakeTestParameters( &testParams, inputDevice, outputDevice, userOptions );
 
     // Loop though combinations of audio parameters.
     for( iFlags=0; iFlags<numFlagSettings; iFlags++ )
     {
         int numRuns = 0;
 
-        testParams.flags = flagSettings[iFlags];
+        TestParameters flagTestParams = testParams;
+        flagTestParams.flags = flagSettings[iFlags];
         printf( "\n************ Mode = %s ************\n",
-               (( testParams.flags & 1 ) ? s_FlagOnNames[0] : s_FlagOffNames[0]) );
+               (( flagTestParams.flags & 1 ) ? s_FlagOnNames[0] : s_FlagOffNames[0]) );
 
         printf("|-   requested  -|-  stream info latency  -|- measured ------------------------------\n");
         printf("|-sRate-|-fr/buf-|- in    - out   - total -|- over/under/calls for in, out -|- frm/buf -|-latency-|- channel results -\n");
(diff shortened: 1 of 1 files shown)


qa/loopback/src/paqa.c
```diff
@@ -996,88 +993,85 @@ static int PaQa_AnalyzeLoopbackConnection( UserOptions *userOptions, PaDeviceInd
...
         {
-            int numBadChannels = PaQa_SingleLoopBackTest( userOptions, &testParams );
+            int numBadChannels = PaQa_SingleLoopBackTest( userOptions, &flagTestParams );
             totalBadChannels += numBadChannels;
         }
     }
 
     printf("\nTest Sample Formats using Half Duplex IO -----\n" );
 
-    PaQa_SetDefaultTestParameters( &testParams, inputDevice, outputDevice );
```
philburk, line 1064: I guess the bug was here.

philburk (approved):
This looks good, cleaner.
I'll let Ross have a look.

qa/loopback/src/paqa.c
```diff
@@ -909,56 +909,54 @@ static int PaQa_SingleLoopBackTest( UserOptions *userOptions, TestParameters *te
 }
 
 /*******************************************************************/
-static void PaQa_SetDefaultTestParameters( TestParameters *testParamsPtr, PaDeviceIndex inputDevice, PaDeviceIndex outputDevice )
+static TestParameters PaQa_GetTestParameters( PaDeviceIndex inputDevice, PaDeviceIndex outputDevice, const UserOptions* userOptions )
```
RossBencina, line 912: - `struct TestParameters` is rather large. I do not agree with returning it by-value. Neither does the internet:
https://stackoverflow.com/questions/30980759/when-should-i-pass-or-return-a-struct-by-value

- The verb "Make" is better than "Get" here.

So my suggestion and request is to go with:

```
static void PaQa_MakeTestParameters( TestParameters *testParamsPtr, PaDeviceIndex inputDevice, PaDeviceIndex outputDevice, const UserOptions* userOptions )
```

RossBencina:
I like the idea. Some change to the implementation requested please.

qa/loopback/src/paqa.c
dechamps: > `struct TestParameters` is rather large. I do not agree with returning it by-value. Neither does the internet:       https://stackoverflow.com/questions/30980759/when-should-i-pass-or-return-a-struct-by-value

The only *potential* advantage of passing that struct as a pointer argument is performance. This is very obviously not a performance-sensitive code path so that argument doesn't apply here. (Even if it did apply, it would not make any difference in this case, because when returning a larger struct, the compiler passes it to the function as hidden pointer behind the scenes, so it will all compile to the same thing anyway.)

There is, however, a significant *readability* argument in favor of returning the struct: it makes it clear what the inputs and outputs of the function are. When passing pointers, it's not necessarily obvious at the callsite which pointees are used as input parameters and which ones are used as output parameters.

Returning a struct also makes it possible to use the return value to initialize a `const` variable, to make it clear that the variable is not changed after initialization. This is another readability advantage.

More arguments in favour of preferring return values to output parameters are laid out in [Abseil TotW #176](https://abseil.io/tips/176) and [C++ Core Guidelines F.20](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-out) (both of which are for C++ but most of the points apply to C too).

> The verb "Make" is better than "Get" here.

Agreed. Done.

RossBencina: PortAudio is a C89 project. C++ guidelines have no bearing here.

RossBencina:
I have requested that the return of the large structure by value be altered to passing in a pointer to the structure. That is standard C practice. The answers to this stack overflow question agree:

https://stackoverflow.com/questions/30980759/when-should-i-pass-or-return-a-struct-by-value

It is also consistent with the usual approach in our code.

Although I am usually open to discussion, in this case I will not approve this PR until the change is made.

qa/loopback/src/paqa.c
dechamps: > PortAudio is a C89 project. C++ guidelines have no bearing here.

The arguments I brought forward in my response, as well as most of the points raised in the links I posted (e.g. literally all the bullet points in the introduction of [Abseil TotW #176](https://abseil.io/tips/176)) apply to C just as well as they apply to C++. Just because it says "C++" on the title doesn't mean the arguments presented lose all merit. The reason why I used C++ sources is because I usually work in C++ so I already had these references at hand.

I am disappointed that you chose to dismiss the points I raised without even attempting to engage in constructive discussion. So far you have not presented specific, concrete reasons as to why this approach is problematic in this particular case.

It is demotivating for contributors to get pushback of their PRs for seemingly arbitrary reasons, and for them to be forced to downgrade the quality of their patches for no other reason than "the maintainer told me to". I don't think this is conducive to attracting contributors to PortAudio in the long term.

dechamps:
> That is standard C practice.

Note that "standard practice" is not the same as "best practice". Otherwise most code out there would be perfect, and I think both you and I know this is very far from being the case.

> The answers to this stack overflow question agree: https://stackoverflow.com/questions/30980759/when-should-i-pass-or-return-a-struct-by-value

I did read through this SO question and the top answers. I see no reason listed there that would reasonably apply to the code we are discussing. There is literally no discussion of readability in the top answers, despite the fact that returning by value vs. passing a pointer has significant readability implications both for caller and callee. Most of the discussions focus on performance, which is irrelevant here (this is not performance sensitive code) and in any case the points raised about performance are dubious at best (as the compiled code would end up being the same, especially so with an optimizing compiler).

Scrolling through the answers, the first answer that actually includes a sensible discussion of the tradeoffs (including readability) is [this one](https://stackoverflow.com/a/45894149/172594) (emphasis mine):

> On a typical PC, performance should not be an issue even for fairly large structures (many dozens of bytes). Consequently other criteria are important, especially semantics: Do you indeed want to work on a copy? Or on the same object, e.g. when manipulating linked lists? The guideline should be to **express the desired semantics with the most appropriate language construct in order to make the code readable and maintainable.**

Well said.

> It is also consistent with the usual approach in the PortAudio code base.

Consistency should not be used as a pretence to perpetuate suboptimal practices, because that makes it impossible to improve the quality of the codebase over time. Rejecting improvements on the grounds that they are not consistent with the rest of the code *de facto* freezes the codebase in its current state, quality-wise, because it is unreasonable to expect people to apply a new pattern to the entire codebase all at once.

In other codebases I work on, consistency arguments are only used as a tie-breaker when there is otherwise no sensible objective argument one way or the other (i.e. it is a purely subjective/cosmetic issue). They are never used to reject improvements, because that's detrimental to the overall quality of the codebase in the long term.

> Although I am usually open to discussion, in this case I will not approve this PR until the change is made.

Sigh.

I have better things to do than to die on this hill, so I have applied your requested change under protest. I believe this can be merged now.

RossBencina (approved):

RossBencina:
@dechamps Thanks for taking the time to explain your position. I think we disagree on a number of points, but like you, I'd like to get this patch merged, and I will do so presently.

I don't think readability is harmed by using an in/out parameter to initialize the struct. It's a standard C idiom. As a general principle I don't believe in pessimising performance for the sake of localised cosmetic/syntactic cleanliness, even in cases where performance is non-critical. (Needless to say, "readability" is a much larger topic than the kind of syntactic minutia that we are quibbling about here.)

I think we disagree on what's optimal. To my mind, in this instance, pass-pointer-to-return-struct is the optimal choice. It objectively has the lowest risk of performing poorly and (subjectively) it is the most idiomatic in C i.e. it is the least surprising to the reader. To me, being "least surprising" enhances readability.

You assert that I am rejecting an *improvement* but I rejected the code precisely because I don't see it as an improvement. I am not making a de facto dismissal, I am making a considered cost-benefit analysis, where consistency with our code and with standard idiom are contributing factors in that analysis.

Finally, I don't think it's fair to extrapolate my reasoning as if the calculus would be the same for any other change to the code base.

dechamps:
Thanks for providing a more detailed response Ross, and I'm glad this finally got merged, even if not in the state I would have liked it. At least we are making some progress.

PR #772, opened 2023-02-25
Stop using paUtilVariableHostBufferSizePartialUsageAllowed in DirectSound

### LINKED ISSUE #770: DirectSound full duplex is glitchy
The DirectSound Host API is glitchy when used in full duplex mode (i.e. when both input and output devices are used in a single PortAudio stream) when using reasonable parameters (default frames per buffer, default high suggested latency).

Example output from `paloopback -r48000 -s1024 --inputLatency 240 --outputLatency 240 -w`:

```
************ Mode = Two Streams (Half Duplex) ************
|-   requested  -|-  stream info latency  -|- measured ------------------------------
|-sRate-|-fr/buf-|- in    - out   - total -|- over/under/calls for in, out -|- frm/buf -|-latency-|- channel results -
| 48000 |   1024 |   21.33  240.00  261.33 |    0/   0/ 469,    0/   0/ 470 | 1024-1024 |  297.29 | OK - #errs = 0
| 48000 |   1024 |   21.33  240.00  261.33 |    0/   0/ 469,    0/   0/ 470 | 1024-1024 |  298.04 | OK - #errs = 0
| 48000 |   1024 |   21.33  240.00  261.33 |    0/   0/ 469,    0/   0/ 470 | 1024-1024 |  298.52 | OK - #errs = 0
| 48000 |   1024 |   21.33  240.00  261.33 |    0/   0/ 469,    0/   0/ 470 | 1024-1024 |  295.52 | OK - #errs = 0
| 48000 |   1024 |   21.33  240.00  261.33 |    0/   0/ 469,    0/   0/ 470 | 1024-1024 |  296.77 | OK - #errs = 0

************ Mode = One Stream (Full Duplex) ************
|-   requested  -|-  stream info latency  -|- measured ------------------------------
|-sRate-|-fr/buf-|- in    - out   - total -|- over/under/calls for in, out -|- frm/buf -|-latency-|- channel results -
| 48000 |   1024 |   21.33  261.33  282.67 |    0/   0/ 469,    0/   0/ 469 | 1024-1024 |  295.40 | channel 0 POP 0.460 at 15220, ".\paloopback_0.wav", channel 1 POP 0.453 at 15220, ".\paloopback_1.wav",  - #errs = 0
| 48000 |   1024 |   21.33  261.33  282.67 |    0/   0/ 469,    0/   0/ 469 | 1024-1024 |  299.40 | channel 0 POP 0.460 at 15410, ".\paloopback_2.wav", channel 1 POP 0.453 at 15410, ".\paloopback_3.wav",  - #errs = 1
| 48000 |   1024 |   21.33  261.33  282.67 |    0/   0/ 469,    0/   0/ 469 | 1024-1024 |  299.44 | channel 0 POP 0.460 at 15412, ".\paloopback_4.wav", channel 1 POP 0.453 at 15412, ".\paloopback_5.wav",  - #errs = 2
| 48000 |   1024 |   21.33  261.33  282.67 |    0/   0/ 469,    0/   0/ 469 | 1024-1024 |  298.60 | channel 0 POP 0.460 at 15373, ".\paloopback_6.wav", channel 1 POP 0.453 at 15373, ".\paloopback_7.wav",  - #errs = 3
| 48000 |   1024 |   21.33  261.33  282.67 |    0/   0/ 469,    0/   0/ 469 | 1024-1024 |  296.06 | channel 0 POP 0.460 at 15252, ".\paloopback_8.wav", channel 1 POP 0.453 at 15252, ".\paloopback_9.wav",  - #errs = 4
```

A quick look at the recorded test signal confirms the issue:

image

MME passes the same test with flying colors.

- PortAudio version: latest master (4bccdfbb68a509e2a96d916046d911959f7d935c)
- Windows 11 22H2 22621.1265
- Loopback audio device: [Virtual Audio Cable](https://vac.muzychenko.net/en/) 4.67.0.12554

dechamps:
This pull request does the following:

- Changes the DirectSound host API code to use `paUtilBoundedHostBufferSize` instead of `paUtilVariableHostBufferSizePartialUsageAllowed`.
- Removes `paUtilVariableHostBufferSizePartialUsageAllowed` as a buffer processor mode as it is now dead - DS was the only code using it.

The rationale behind the above is that `paUtilVariableHostBufferSizePartialUsageAllowed` appears to have no benefit over `paUtilBoundedHostBufferSize`, only downsides, and it causes problems with initialization/priming (see #770).

See individual commit messages for details.

This PR fixes #770.


RossBencina:
To begin: this is a significant change to a core part of the code base and it's going to require a large investment of time to review it properly.

Ideally we need a test case that isolates and demonstrates the bug in the buffer processor.

> paUtilVariableHostBufferSizePartialUsageAllowed appears to have no benefit over paUtilBoundedHostBufferSize

It may appear to have no benefit, and I will have to study this, but I doubt that I would have implemented it that way if it had no benefit whatsoever. Most likely that mode exists to avoid unnecessary data movement, e.g. in the case where the client requests a fixed size buffer.

At a certain time in the project history, when we had less strict review policies, there were a number of poorly reviewed changes to the buffer processor that were let through. So there is some chance that a bug was introduced that did not exist in the original stabel version of the buffer processor.

dechamps:
> To begin: this is a significant change to a core part of the code base and it's going to require a large investment of time to review it properly.

Well to be fair, the change only affects DirectSound. Other Host APIs are not impacted because no other Host API is using this mode.

By the way, Ross, if you don't think you (or anyone else) would have the time to review this with a fine-tooth comb, then please consider [my plea to be less thorough](https://github.com/PortAudio/portaudio/pull/774#issuecomment-1449046691) and approve it by default, instead of rejecting it by default. Your contributors will thank you.

> Ideally we need a test case that isolates and demonstrates the bug in the buffer processor.

I did look for a test suite for the buffer processor that I could add tests to, but I didn't find one. It would be nice to have such a test suite, as the buffer processor is a nice self-contained piece of code with a clear interface boundary, making it ideal for unit testing. It would be a nice project, but I don't think it's reasonable to require this as part of this PR.

For what it's worth, the issue can be trivially reproduced through `paloopback` as shown in #770, albeit with the DirectSound Host API code in the middle.

> It may appear to have no benefit, and I will have to study this, but I doubt that I would have implemented it that way if it had no benefit whatsoever.

One thing to note is that DS is not the only Host API that uses circular buffers, yet it's the only one that is using that mode. That might be an indication that this mode is not that useful.

> Most likely that mode exists to avoid unnecessary data movement, e.g. in the case where the client requests a fixed size buffer.

I agree that this would make sense (it would avoid unnecessary copies if there is no need for sample conversion), but that's not what's happening in practice. `paUtilBoundedHostBufferSize` vs. `paUtilVariableHostBufferSizePartialUsageAllowed` only affects the behaviour of `AdaptingProcess()`, but `AdaptingProcess()` *never* passes host buffers directly to the stream callback - it always copies data into temp buffers first.

https://github.com/PortAudio/portaudio/blob/5b5feb3a331b674cfc71009414514000685bf547/src/common/pa_process.c#L1423-L1459

Maybe it's worth exploring how `AdaptingProcess()` could be improved to directly feed host buffers to the stream callback under certain conditions, but right now, I think it's more urgent to fix the glitches. We can discuss such performance improvements in a separate issue.

dechamps:
> I did look for a test suite for the buffer processor that I could add tests to, but I didn't find one. It would be nice to have such a test suite, as the buffer processor is a nice self-contained piece of code with a clear interface boundary, making it ideal for unit testing. It would be a nice project, but I don't think it's reasonable to require this as part of this PR.

Filed #779 to track this separately.

RossBencina:
I'm assigning this to me for now. Hopefully I'll have time to look at it next week.

PR #773, opened 2023-02-25
wasapi: Refactored audio client instance creation.

dmitrykos:
1) Refactored audio client instance creation by joining separate versions for input/output.
2) Removed unused code related to `IAudioEndpointVolume`.
3) Left comment about double buffering on PA WASAPI side inside `_CalculateFramesPerHostBuffer`.

Change 1. makes implementation more compact.

Change 2. removed `IAudioEndpointVolume` related code which was always unused by PA implementation.

Change 3. does not change PA WASAPI behavior, it just adds the comment about enforced double buffering.

### DIFF
--- a/src/hostapi/wasapi/pa_win_wasapi.c
+++ b/src/hostapi/wasapi/pa_win_wasapi.c
@@ -1,7 +1,7 @@
 /*
  * Portable Audio I/O Library WASAPI implementation
  * Copyright (c) 2006-2010 David Viens
- * Copyright (c) 2010-2022 Dmitry Kostjuchenko
+ * Copyright (c) 2010-2023 Dmitry Kostjuchenko
  *
  * Based on the Open Source API proposed by Ross Bencina
  * Copyright (c) 1999-2019 Ross Bencina, Phil Burk
@@ -584,7 +584,6 @@ typedef struct PaWasapiStream
     IStream                   *captureClientStream;
 #endif
     IAudioCaptureClient       *captureClient;
-    IAudioEndpointVolume      *inVol;
 
     // output
     PaWasapiSubStream          out;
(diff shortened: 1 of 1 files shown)


RossBencina:
Hi Dmitry,

Is there also a third change:
3. remove `IAudioEndpointVolume` control?

If so could you add that to the description please?

Assuming that the changes are unlinked, I think it would be better to split this into 3 separate PRs. I have not reviewed it in depth, but `ActivateAudioClient` refactoring seems like an uncontroversial improvement that we could merge separately. Maybe the `IAudioEndpointVolume` change too. On the other hand, the double buffering change seems like something that needs further feedback from users and would be better handled as a separate PR. I am also concerned about it being "buried" with these other changes.

What do you think?

dmitrykos:
Hi Ross,

Thank you for your comments. 

I considered changes again and propose to leave double buffering switched on for backwards compatibility, unless anyone reports that Shared mode can really reach less than 22 ms latency (none of my devices can for input/output). I left comment about double buffering though for other developers inside `_CalculateFramesPerHostBuffer`.

I modified comments regarding `IAudioEndpointVolume`. `IAudioEndpointVolume` was always unused and not allocated by PA WASAPI (commented out), so it can be safely removed without affecting anything. `IAudioEndpointVolume` was a part of main work - audio client instance creation refactoring/simplification (`inVol`, `outVol` were used inside `ActivateAudioClientOutput` and `ActivateAudioClientInput` which were joined into a single `ActivateAudioClient`), so I do not think that dead code removal deserves separate PR?

dmitrykos:
@RossBencina if proposed changes are ok now it would be great to merge them before the #778 as they both overlap.

philburk (approved):

RossBencina (approved):

PR #776, opened 2023-02-26
use PROJECT_SOURCE_DIR to set the CMAKE_MODULE_PATH

robUx4:
This allows using the source tree inside another project with add_subdirectory().

See https://cmake.org/cmake/help/v3.26/variable/PROJECT_SOURCE_DIR.html which is available in 3.1, the minimum required by portaudio.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -97,7 +97,7 @@ if(PA_USE_SKELETON)
 endif()
 
 include(CMakeDependentOption)
-set(CMAKE_MODULE_PATH "${CMAKE_MODULE_PATH}" "${CMAKE_CURRENT_SOURCE_DIR}/cmake/modules")
+set(CMAKE_MODULE_PATH "${CMAKE_MODULE_PATH}" "${PROJECT_SOURCE_DIR}/cmake/modules")
 
 # JACK is most commonly used on Linux, but it is cross platform, so allow building it on any OS
 # if the FindJACK.cmake module finds POSIX headers.


RossBencina:
I don't know enough about CMake to review this change for possible negative side effects. Could others please advise? @Be-ing @daschuer @dechamps et al

daschuer:
For my understanding we have 
/src
/cmake (1)
/lib/portaudio/cmake (2)

This patch adds (1) instead of (2). Is (1) not already part of the CMAKE_MODULE_PATH?

Why we are certain that (1) contains all modules if (2)?

Febbe:
Actually, this change has the opposite effect. Portaudio's CMakeLists.txt requires those self defined modules, relative to it's top level CMakeLists.txt. But this change requires them to be everywhere, where the most top level project lives.

Imaging `/home/<me>/my_proj/CMakeLists.txt` and `/home/poppler/CMakeLists.txt`
And I add poppler with 

```
add_subdirectory(/home/poppler/CMakeLists.txt ${CMAKE_BUILD_DIRECTORY}/poppler)
```

Then the search path of those modules changes to  `/home/<me>/my_proj/cmake/modules` but in fact those files are still located in `/home/poppler/cmake/modules`

RossBencina:
@robUx4 doesn't sound like others are very happy with this. Please close the PR or defend it.

RossBencina:
Closing as abandoned. If you think there's still a problem with PortAudio please open an issue to discuss further.

PR #778, opened 2023-03-01
wasapi: Add passthrough support (AC3, E-AC3, ...)

gabys999:
add AC3 and E-AC3 passthrough support for wasapi. The example is in the test folder.


dmitrykos:
@gabys999 thank you for PR. Please correct failing tests first and code style according to my previous comments. Note spaces in all your code and align accordingly to make code looking nice and uniform with other PA WASAPI code.

dmitrykos:
@gabys999 thank you for the changes, although you forgot moving 192000 constant into separate define.

Your implementation looks suboptimal to me and confusing in `GetAlternativeSampleFormatExclusive` where the use of `WAVEFORMATEXTENSIBLE_IEC61937` is not really needed. The same is for `IsFormatSupported`. Please see my comments.

Please make these corrections first.

gabys999:
Thank you for pointing out these problems. Previously, we usually only used upper level functions, such as Pa_ StartStream, Pa_ WriteStream, lack of understanding of wasapi. We modified the GetAlternativeSampleFormatExclusive and IsFormatSupported to worry that if the E-AC3 passthrough needs to call these functions, the unallocated memory will cause program crash. Our purpose of submitting the PR is also to collaborate with you so that portaudio can stably support AC3 and E-AC3 audio passthrough for our project. I have reverted the code you mentioned.

dmitrykos:
@gabys999 thank you for your corrections! 

> Our purpose of submitting the PR is also to collaborate with you so that portaudio can stably support AC3 and E-AC3 audio passthrough for our project.

Your kind input is indeed greatly appreciated. The goal of my requests for corrections is to keep PA WASAPI implementation uniform and optimal once your contribution is merged into the repository.

dmitrykos:
The reason of a crash is that `WAVEFORMATEXTENSIBLE_IEC61937` is larger than `WAVEFORMATEXTENSIBLE` therefore writing anything to the additional members of `WAVEFORMATEXTENSIBLE_IEC61937` inside `MakeWaveFormatFromParams` will corrupt the stack memory because currently stack memory is allocated for `WAVEFORMATEXTENSIBLE`.

Now when you cleaned up those functions we need to fix implementation and allocate enough stack memory for both structures in all places where WAVEFORMATEXTENSIBLE is used. For that I propose to define such type (after definition of `WAVEFORMATEXTENSIBLE_IEC61937`):

```c
typedef union _WAVEFORMATEXTENSIBLE_UNION
{
    WAVEFORMATEXTENSIBLE          ext;
    WAVEFORMATEXTENSIBLE_IEC61937 iec61937;
} WAVEFORMATEXTENSIBLE_UNION;
```

and then replace `WAVEFORMATEXTENSIBLE` allocation inside functions or as member variable with `WAVEFORMATEXTENSIBLE_UNION` everywhere in PA WASAPI implementation to guarantee absence of memory corruption. I hope you got my idea.

In case you wish to join forces, please give me write access to your repository and I can implement proposed changes. Then we will check and discuss if all works as expected. Let me know what you think.

dmitrykos:
I quickly went through the code and applied `WAVEFORMATEXTENSIBLE_UNION` to:

```c
typedef struct PaWasapiSubStream:
 -> change     WAVEFORMATEXTENSIBLE wavex;
 -> to         WAVEFORMATEXTENSIBLE_UNION wavexu;

static PaError MakeWaveFormatFromParams(WAVEFORMATEXTENSIBLE_UNION *wavexu, const PaStreamParameters *params,
    double sampleRate, BOOL packedOnly)
    
static HRESULT GetAlternativeSampleFormatExclusive(IAudioClient *client, double sampleRate,
    const PaStreamParameters *params, WAVEFORMATEXTENSIBLE_UNION *outWavexU, BOOL packedSampleFormatOnly)
    
static PaError GetClosestFormat(IAudioClient *client, double sampleRate, const PaStreamParameters *_params,
    AUDCLNT_SHAREMODE shareMode, WAVEFORMATEXTENSIBLE_UNION *outWavexU, BOOL output)
```

Inside `GetClosestFormat` I used it like that:

```c
WAVEFORMATEXTENSIBLE *outWavex   = (WAVEFORMATEXTENSIBLE *)outWavexU;
```

`WAVEFORMATEXTENSIBLE_UNION` must be used as storage type in `GetAlternativeSampleFormatExclusive`, `IsFormatSupported`.

When compiling compiler will throw a lot of warnings about incompatible types used, as well as PaWasapiSubStream::wavex will be absent, so warnings and compile errors need to be corrected accordingly that will result in a working solution.

gabys999:
You are right. The code I modified before is mainly for ac3 and eac3. It does not consider supporting more formats uniformly, so it limits the code implementation.

According to the definition of `WAVEFORMATEXTENSIBLE_IEC61937`:
```
typedef struct {
    WAVEFORMATEXTENSIBLE  FormatExt;    /* Format of encoded data as it is */
                                        /* intended to be streamed over the link */
    DWORD   dwEncodedSamplesPerSec;     /* Sampling rate of the post-decode audio. */
    DWORD   dwEncodedChannelCount;      /* Channel count of the post-decode audio. */
    DWORD   dwAverageBytesPerSec;       /* Byte rate of the content, can be 0. */
} WAVEFORMATEXTENSIBLE_IEC61937, *PWAVEFORMATEXTENSIBLE_IEC61937;
```
My opinion is that `WAVEFORMATEXTENSIBLE_IEC61937` is compatible with `WAVEFORMATEXTENSIBLE`,  wasapi interface can determine whether the type of the input parameter is `WAVEFORMATEXTENSIBLE_IEC61937` or `WAVEFORMATEXTENSIBLE` through `FormatExt.Format.cbsize` ,  the interface can correctly use the allocated stack memory. Is it necessary to define `WAVEFORMATEXTENSIBLE_UNION`, or use `WAVEFORMATEXTENSIBLE_IEC61937` to replace `WAVEFORMATEXTENSIBLE` completely?

I added you to my repository and look forward to your implementation.

dmitrykos:
> Is it necessary to define `WAVEFORMATEXTENSIBLE_UNION`, or use `WAVEFORMATEXTENSIBLE_IEC61937` to replace `WAVEFORMATEXTENSIBLE` completely?

My idea is to differentiate them. Even though `WAVEFORMATEXTENSIBLE_IEC61937` is compatible with `WAVEFORMATEXTENSIBLE` it is declared by MS for use with SPDIF (IEC61937) format specifically and therefore it may be a bit confusing for other developers who consider only PCM case. 

Therefore, to my opinion, union in this case solves ambiguity by differentiating to `ext` and `iec61937` members which clearly denote target use.

I propose to wait for #773 merging because it overlaps with this PR and then I will apply my described approach in your repository.

gabys999:
> I propose to wait for #773 merging because it overlaps with this PR and then I will apply my described approach in your repository.

No problem, I agree with you.

RossBencina:
#773 is merged now.

dmitrykos:
@RossBencina thank you very much. @gabys999 committed changes and PCM mode works fine for me. Would you please check SPDIF functionality on your side with these new changes.

dmitrykos (approved):

dmitrykos (approved):

gabys999:
Okay, I'll check it as soon as possible

gabys999:
@dmitrykos 
I ran the test program in AC3 mode and E-AC3 mode on the Dobly CP850, and their output was correct.

**AC3**
ac3

**E-AC3**
eac3

dmitrykos:
That's great, thank you very much for testing! Let's wait for @RossBencina review, from my side final implementation looks fine.

test/patest_wasapi_ac3.c
```diff
@@ -0,0 +1,215 @@
...
+    @ingroup examples_src
+    @brief Use WMME-specific interface to send raw AC3 data to a S/PDIF output.
+    @author Ross Bencina <rossb@audiomulch.com>
+*/
+/*
+ * $Id: $
+ * Portable Audio I/O Library
+ * Windows MME ac3 sound output test
+ *
+ * Copyright (c) 2009 Ross Bencina
```
RossBencina, line 11: Please update:

```suggestion
 * Copyright (c) 2009-2023 Ross Bencina, <Your Name Here>
```

test/patest_wasapi_ac3.c
```diff
@@ -0,0 +1,215 @@
+/** @file paex_wmme_ac3.c
+    @ingroup examples_src
+    @brief Use WMME-specific interface to send raw AC3 data to a S/PDIF output.
```
RossBencina, line 3: Should this not say "WASAPI" not "WMME"?

test/patest_wasapi_eac3.c
```diff
@@ -0,0 +1,214 @@
+/** @file paex_wmme_ac3.c
+    @ingroup examples_src
+    @brief Use WMME-specific interface to send raw AC3 data to a S/PDIF output.
```
RossBencina, line 3: "WASAPI"

test/patest_wasapi_eac3.c
```diff
@@ -0,0 +1,214 @@
...
+    @author Ross Bencina <rossb@audiomulch.com>
```
RossBencina, line 4: Please fix these comments to include something other than just my name.

test/patest_wasapi_eac3.c
```diff
@@ -0,0 +1,214 @@
...
+int main(int argc, char* argv[])
+{
+    PaStreamParameters outputParameters;
+    PaWasapiStreamInfo wasapiStreamInfo;
+    PaStream *stream;
+    PaError err;
+    paTestData data;
+    int deviceIndex;
+    FILE *fp;
+    const char *fileName = "d:\\test_48k.eac3.spdif";
```
RossBencina, line 118: Better to make the string a `#define` at the top of the file.

test/patest_wasapi_eac3.c
```diff
@@ -0,0 +1,214 @@
...
+    outputParameters.channelCount = CHANNEL_COUNT;
+    outputParameters.sampleFormat = paInt16;
+    outputParameters.suggestedLatency = Pa_GetDeviceInfo( outputParameters.device )->defaultLowOutputLatency;
+    outputParameters.hostApiSpecificStreamInfo = NULL;
+
+    wasapiStreamInfo.size = sizeof(PaWasapiStreamInfo);
+    wasapiStreamInfo.hostApiType = paWASAPI;
+    wasapiStreamInfo.version = 1;
+    wasapiStreamInfo.flags = paWinWasapiExclusive | paWinWasapiUseChannelMask | paWinWasapiMatchEac3;
+    wasapiStreamInfo.channelMask = PAWIN_SPEAKER_5POINT1;
```
RossBencina, line 169: Something is fishy here. `CHANNEL_COUNT` is 2 but you're using a 5.1 channel mask?

src/hostapi/wasapi/pa_win_wasapi.c
```diff
@@ -604,7 +623,7 @@ typedef struct PaWasapiStream
     volatile BOOL isActive;
 
     // Stream state: stopped (triggered by user only via external API)
-    BOOL isStopped;
+    volatile BOOL isStopped;
```
RossBencina, line 626: Is this change required for AC3 / E-AC3? or is it a fix for a bug that is unrealated to the purpose of this PR? If it is an independent fix please move it to a separate PR so that it is easily seen in the merge history.

src/hostapi/wasapi/pa_win_wasapi.c
```diff
@@ -2943,12 +2985,27 @@ static PaError MakeWaveFormatFromParams(WAVEFORMATEXTENSIBLE *wavex, const PaStr
     else
     {
         old->wFormatTag = WAVE_FORMAT_EXTENSIBLE;
-        old->cbSize        = sizeof(WAVEFORMATEXTENSIBLE) - sizeof(WAVEFORMATEX);
+        old->cbSize     = matchEac3 ? sizeof(WAVEFORMATEXTENSIBLE_IEC61937) - sizeof(WAVEFORMATEX) : sizeof(WAVEFORMATEXTENSIBLE) - sizeof(WAVEFORMATEX);
```
RossBencina, line 2988: Consider splitting long line, e.g.

```
a = cond
    ? trueValue
    : falseValue
```

include/pa_win_wasapi.h
```diff
@@ -82,7 +82,13 @@ typedef enum PaWasapiFlags
        playback formats that do not match the current configured system settings.
        this is in particular required for streams not matching the system mixer sample rate.
        only applies in Shared mode. */
-    paWinWasapiAutoConvert              = (1 << 6)
+    paWinWasapiAutoConvert              = (1 << 6),
+
+    /* force data output to match AC3 passthrough mode */
+    paWinWasapiMatchAc3                 = (1 << 7),
```
RossBencina, line 88: I think it is worthwhile to make the naming used in the public API to this functionality as nice as possible. Is there a reason to use the verb "match" here? It doesn't seem right to me. Please consider options like:

- `paWinWasapiAc3` and `paWinWasapiEac3`
- `paWinWasapiForceAc3` and `paWinWasapiForceEac3`
- `paWinWasapiFormatAc3` and `paWinWasapiFormatEac3`
- `paWinWasapiAc3Passthrough` and `paWinWasapiEac3Passthrough`

Or maybe something else. You guys are the most knowledgeable and in the best position to decide what captures the meaning best, but I think that I prefer `paWinWasapiAc3` or `paWinWasapiAc3Passthrough`.

RossBencina:
@gabys999 Thanks for contributing this, I think it's a great addition. I have added my review comments above.

test/patest_wasapi_eac3.c
dmitrykos: @RossBencina I think AC3/EAC3 5.1 data is transferred inside 2-channel PCM stream because SPDIF does not have enough throughput for uncompressed 5.1 PCM. So in this sense it looks correct.

include/pa_win_wasapi.h
dmitrykos: I prefer `*Passthrough` variants.

src/hostapi/wasapi/pa_win_wasapi.c
dmitrykos: Although harmless, reverted my change.

gabys999:
I created this test file based on paex_wmme_ac3.c, and it seems that there were some omissions. I will modify them immediately.

test/patest_wasapi_eac3.c
gabys999: @RossBencina @dmitrykos In this section, I refer to the VLC code, and its channel_count and channel_mask are assigned in this way. After testing, although the channel_count is set to 2, the output is still Dolby Atmos.

RossBencina: @dmitrykos @gabys999 my question is really about whether this `wasapiStreamInfo.channelMask` must match the channel configuration that is encoded in the AC3/EAC3-SPDIF data stream or if it is unnecessary to specify a channel mask here (because the channel information is encoded in the SPDIF data stream). If it is working as-is I guess it does not matter, I'm just checking whether this code is redundant or unnecessary.

RossBencina: Further, if it is necessary, it would be good to add a comment like `/* IMPORTANT: you must specify the correct channel mask here. */` or if it is not necessary use the comment: `/* Set channel mask for consistency with the SPDIFF data format, but it seems to work either way. */`

RossBencina:
@gabys999 @dmitrykos I think this is good to merge now (subject to a final check with Phil later in the week), but maybe a clarifying comment for the assignment to `wasapiStreamInfo.channelMask` in the examples would be nice.

gabys999:
image
@dmitrykos @RossBencina This is an unexpected result. I tested AC3 and EAC3 separately, and each test  is assigned five different value to `wasapiStreamInfo.channelMask` . No matter what `wasapiStreamInfo.channelMask` is assigned to, it will not affect the result of the passthrough output. It does seem to work regardless of the assignment.

src/hostapi/wasapi/pa_win_wasapi.c
```diff
@@ -2917,11 +2961,21 @@ static PaError MakeWaveFormatFromParams(WAVEFORMATEXTENSIBLE *wavex, const PaStr
         useExtensible = TRUE;
     }
 
+    if ((streamInfo != NULL) && (streamInfo->flags & paWinWasapiAc3Passthrough))
+    {
+        isAc3Passthrough = TRUE;
```
RossBencina, line 2966: Probably you should check whether `streamInfo->flags & paWinWasapiUseChannelMask` is set (?and return an error if it is not. Unfortunately I don't know what error to return). Otherwise, if `paWinWasapiUseChannelMask` is not set by the caller, channelMask will be zero.

Alternatively you could copy the channelMask here even if `paWinWasapiUseChannelMask` is not set.

src/hostapi/wasapi/pa_win_wasapi.c
```diff
@@ -2917,11 +2961,21 @@ static PaError MakeWaveFormatFromParams(WAVEFORMATEXTENSIBLE *wavex, const PaStr
...
+    }
+
+    if ((streamInfo != NULL) && (streamInfo->flags & paWinWasapiEac3Passthrough))
+    {
+        isEac3Passthrough = TRUE;
```
RossBencina, line 2971: Same channel mask logic here.

RossBencina:
@gabys999 thanks for testing. I think we have all the information we need for a final review. Hopefully Phil and I will have time to address this tomorrow.

Dmitry wrote:
> I propose to stick to our current implementation as it is now and have `dwEncodedChannelCount` assigned as per MS docs just to be on the safe side. Maybe it is also driver-dependent functionality.

I see. That sounds reasonable to me. Just thinking out loud here, but if the MS docs say `dwEncodedChannelCount` needs to be set according to the content of the SPDIF stream then obviously the channel count needs to be communicated via `Pa_OpenStream` somehow, irrespective of how particular drivers behave. Right now you're doing that using `wasapiStreamInfo.channelMask` and `PopCount(channelMask)`. This seems OK to me. The alternatives are:

1. use  `outputParameters.channelCount`. Unfortunately if `outputParameters.channelCount` is used is make for other confusion, because the PA callback will deal with 2 channel 16 bit frames, even though outputParameters.channelCount is set to e.g. 6.
2. add a separate `wasapiStreamInfo.encodedChannelCount` field.

I'm not arguing for either of these alternatives. But I will discuss them with Phil.

src/hostapi/wasapi/pa_win_wasapi.c
gabys999: > Alternatively you could copy the channelMask here even if paWinWasapiUseChannelMask is not set.

You mean I assign a default value to `channelMask`? Such as those preset in the ac3 and eac3 test code.

dmitrykos:
> 2. add a separate `wasapiStreamInfo.encodedChannelCount` field.

@RossBencina I am in favor of adding an additional members to `PaWasapiStreamInfo` and make a direct link to additional parameters of `WAVEFORMATEXTENSIBLE_IEC61937` as such:
```c
typedef enum PaWasapiPassthroughFormat
{
    ePassthroughFormatDolbyDigital     = 0x00920000,
    ePassthroughFormatDolbyDigitalPlus = 0x000a0cea,
    // and so on
}
PaWasapiPassthroughFormat;

typedef struct PaWasapiStreamPassthroughInfo
{
    PaWasapiPassthroughFormat formatId;
    unsigned long encodedSamplesPerSec;
    unsigned long encodedChannelCount;
    unsigned long AverageBytesPerSec;
}
PaWasapiStreamPassthroughInfo;
```

Then we can have just one `paWinWasapiPassthrough = (1 << 7)` flag which denotes presence of filled by the user `PaWasapiStreamPassthroughInfo`.

With such approach our implementation become more generic and will support all passthrough WASAPI formats. With this we will avoid `PopCount` and any speculations about what `encodedChannelCount` shall be by delegating this task to the user. 

What do you think about this proposal? If it makes sense I will update currently achieved implementation to such approach.

RossBencina:
> What do you think about this proposal? If it makes sense I will update currently achieved implementation to such approach.

@dmitrykos sounds good to me. I like the idea of making it more generic.

include/pa_win_wasapi.h
```diff
@@ -82,7 +82,13 @@ typedef enum PaWasapiFlags
        playback formats that do not match the current configured system settings.
        this is in particular required for streams not matching the system mixer sample rate.
        only applies in Shared mode. */
-    paWinWasapiAutoConvert              = (1 << 6)
+    paWinWasapiAutoConvert              = (1 << 6),
+
+    /* force data output to AC3 passthrough mode */
```
philburk, line 87: Please specify whether the app should pass raw AC3 data or AC3 data wrapped in an S/PDIF wrapper, AKA IEC61937. I think this API requires wrapped IEC61937data so that needs to be clarified.

Please link to an AES or IEC doc or give the exact name of the standard for the data.

RossBencina:
With regard to the data format supplied to PortAudio. I am confused about the relationship between the 24-bit 2 channel AES3/SPDIF payload ([e.g. as described here](https://en.wikipedia.org/wiki/AES3) and the 16-bit 2 channel IEC61937 payload required by Microsoft's API ([see code example at the bottom of this MSDN page](https://learn.microsoft.com/en-us/windows/win32/coreaudio/representing-formats-for-iec-61937-transmissions). To me it seems like Microsoft is adding at least some of the IEC61937 into those extra 8 bits. Which casts doubt on exactly what the format of the "SPDIF ac3" stream that must be generated by the PA callback. Can you guys clarify please?

philburk (changes requested):

gabys999:
I read [Representing Formats for IEC 61937 Transmissions](https://learn.microsoft.com/en-us/windows/win32/coreaudio/representing-formats-for-iec-61937-transmissions) and think some details may still need to be addressed.

The latest vlc source code is as follows:
```
case VLC_CODEC_EAC3:
        wf->SubFormat = _KSDATAFORMAT_SUBTYPE_IEC61937_DOLBY_DIGITAL_PLUS;
        wf->Format.nChannels = 2;
        wf->dwChannelMask = KSAUDIO_SPEAKER_5POINT1;
        break;
...
    wf->Format.wFormatTag = WAVE_FORMAT_EXTENSIBLE;
    wf->Format.nSamplesPerSec = 192000;
    wf->Format.wBitsPerSample = 16;
    wf->Format.nBlockAlign = wf->Format.wBitsPerSample / 8 * wf->Format.nChannels;
    wf->Format.nAvgBytesPerSec = wf->Format.nSamplesPerSec * wf->Format.nBlockAlign;
    wf->Format.cbSize = sizeof (*wf_iec61937) - sizeof (wf->Format);

    wf->Samples.wValidBitsPerSample = wf->Format.wBitsPerSample;

    wf_iec61937->dwEncodedSamplesPerSec = audio->i_rate;
    wf_iec61937->dwEncodedChannelCount = audio->i_channels;
    wf_iec61937->dwAverageBytesPerSec = 0;
```
For all IEC61937-based parameter settings, vlc sets `Format.nSamplesPerSec` to 192000. So when I referenced its code, I also defined `PA_WASAPI_EAC3_SAMPLES_PER_SEC_ (192000)` in portaudio.
> When Dolby Digital Plus content is transmitted over IEC 60958, the link-sampling rate must be four times the sampling rate of the content. Dolby Digital Plus supports content sample rates of 32 KHz, 44.1 KHz, and 48 KHz. Interfaces such as HDMI do not support 128 KHz (32 KHz x 4), so only 44.1 and 48 KHz content sample rates can be supported.

But this page says that DOLBY_DIGITAL_PLUS supports 44.1k and 48k content sampling rates, which means that `Format.nSamplesPerSec` should be set to dwEncodedSamplesPerSec * 4. Also, I'm not sure if `FormatExt.dwChannelMask = KSAUDIO_SPEAKER_5POINT1` and `dwEncodedChannelCount = 6` don't need to be modified, the comments on the example don't add always.

gabys999:
The audio part of my new test file is in Dolby Atmos format, I converted it to pure audio format of EAC3 via ffmpeg. When playing EAC3 audio with portaudio_test, the dolby web status shows 5.1 channels, when playing the test file with our application that has been updated to the latest version of portaudio_x64.dll, the dolby web status shows atmos.
image

image

src/hostapi/wasapi/pa_win_wasapi.c
```diff
@@ -2909,12 +2896,7 @@ PaSampleFormat WaveToPaFormat(const WAVEFORMATEXTENSIBLE *fmtext)
             }
         }
         else
-        if (IsEqualGUID(&fmtext->SubFormat, &pa_KSDATAFORMAT_SUBTYPE_IEC61937_DOLBY_DIGITAL))
-        {
-            return paInt16;
-        }
-        else
-        if (IsEqualGUID(&fmtext->SubFormat, &pa_KSDATAFORMAT_SUBTYPE_IEC61937_DOLBY_DIGITAL_PLUS))
+        if (IsEqualGUID(&fmtext->SubFormat, &pa_KSDATAFORMAT_SUBTYPE_IEC61937_PCM))
```
gabys999, line 2899: @dmitrykos The logic of comparing GUIDs here is wrong and will cause all `fmtext->SubFormat` that are not `pa_KSDATAFORMAT_SUBTYPE_IEC61937_PCM` to return paCustomFormat. After this function correctly returns `paInt16`, pa_test is played normally.

gabys999:
I tried to play dolby atmos files in spdif format with different parameters, for example:
```
    wasapiStreamInfo.flags = paWinWasapiExclusive | paWinWasapiUseChannelMask | paWinWasapiPassthrough;
    wasapiStreamInfo.channelMask = PAWIN_SPEAKER_7POINT1;
    wasapiStreamInfo.passthrough.formatId = ePassthroughFormatDolbyDigitalPlusAtmos;
    wasapiStreamInfo.passthrough.encodedSamplesPerSec = SAMPLE_RATE / 4;
    wasapiStreamInfo.passthrough.encodedChannelCount = 8;
```
Always get the wrong result `WASAPI ERROR HRESULT: 0x88890008 : AUDCLNT_E_UNSUPPORTED_FORMAT`

I saw a message in chromium's [historical source code](https://chromium.googlesource.com/chromium/src/+/d026ac796d..103665932a): 
> // Test Dolby Digital+ / Atmos support. For whatever reason Atmos doesn't use
> // the KSDATAFORMAT_SUBTYPE_IEC61937_DOLBY_DIGITAL_PLUS_ATMOS SubFormat.

In the latest chromium code, I don't see any Dolby Atmos-related implementation
>     case AudioParameters::AUDIO_BITSTREAM_AC3:
>     case AudioParameters::AUDIO_BITSTREAM_EAC3:
>     case AudioParameters::AUDIO_BITSTREAM_DTS:
>     case AudioParameters::AUDIO_BITSTREAM_DTS_HD:
>     case AudioParameters::AUDIO_BITSTREAM_DTSX_P2:
>     case AudioParameters::AUDIO_BITSTREAM_IEC61937:

I don't know if `DolbyDigitalPlusAtmos` is buggy in wasapi, but I changed to `DolbyDigitalPlus` and the Dolby CP850 device goes to automatically recognize the 7.1 channels atmos data, which is how we currently use it.

dmitrykos:
> Because I changed `DolbyDigitalPlusAtmos` to `DolbyDigitalPlus` and the Dolby CP850 device are able to recognize the 7.1 channels atmos data, that's the way we are currently using.

It is most likely a driver issue or OS-side WASAPI bug. I double check PA WASAPI and it seems I did not make mistake when copying value of the `KSDATAFORMAT_SUBTYPE_IEC61937_DOLBY_DIGITAL_PLUS_ATMOS` GUID - `0000010a-0cea-0010-8000-00aa00389b71` where we use first 16 bits of Data1 part and all 16 bits of Data2 part in the `ePassthroughFormatDolbyDigitalPlusAtmos` which is - `0x010a0cea`.

gabys999:
I think so too, I checked the parameters and GUID when the error occurred, and I think they are not wrong. In the latest code, the test program plays ac3 and eac3 passthrough correctly, and also plays atmos with the eac3 parameters.

dmitrykos:
@gabys999  thank you very much for testing, I am glad that now all works as expected! Let's wait for @RossBencina and @philburk for their final word. I think we reached quite optimal implementation in the long run.

gabys999:
I'm glad to be involved in the implementation of the passthrough feature and I think portaudio will work even better in our project. If you need me later, please @ me

RossBencina (changes requested):
Looks good, but I've made a few remarks that need to be addressed.

include/pa_win_wasapi.h
RossBencina: We're going to merge without resolving this, I created a ticket to track further progress:

https://github.com/PortAudio/portaudio/issues/816

test/patest_wasapi_ac3.c
```diff
@@ -0,0 +1,206 @@
...
+#define SAMPLE_RATE         (48000)
+#define FRAMES_PER_BUFFER   (64)
+#define CHANNEL_COUNT       (2)
+
+#define AC3_FILEPATH        "./test_48k.ac3.spdif"
+
+typedef struct
+{
+    short *buffer;
+    int bufferSampleCount;
```
RossBencina, line 61:
```suggestion
    int bufferFrameCount;
```
Count whole frames, not 2-byte words.

test/patest_wasapi_ac3.c
```diff
@@ -0,0 +1,206 @@
...
+    (void) inputBuffer;
+
+    /* stream out contents of data->buffer looping at end */
+
+    for( i=0; i<framesPerBuffer; i++ )
+    {
+        for( j = 0; j < CHANNEL_COUNT; ++j ){
+            *out++ = data->buffer[ data->playbackIndex++ ];
+
+            if( data->playbackIndex >= data->bufferSampleCount )
```
RossBencina, line 91:
```suggestion
```

test/patest_wasapi_ac3.c
```diff
@@ -0,0 +1,206 @@
...
+                data->playbackIndex = 0; /* loop at end of buffer */
+        }
+    }
```
RossBencina, line 94:
```suggestion
        if( data->playbackIndex >= data->bufferFrameCount )
            data->playbackIndex = 0; /* loop at end of buffer */
    }
```

Check for source buffer wrap-around outside inner loop.

test/patest_wasapi_ac3.c
```diff
@@ -0,0 +1,206 @@
...
+
+    /* stream out contents of data->buffer looping at end */
+
+    for( i=0; i<framesPerBuffer; i++ )
+    {
+        for( j = 0; j < CHANNEL_COUNT; ++j ){
+            *out++ = data->buffer[ data->playbackIndex++ ];
+
+            if( data->playbackIndex >= data->bufferSampleCount )
+                data->playbackIndex = 0; /* loop at end of buffer */
```
RossBencina, line 92:
```suggestion
```

test/patest_wasapi_ac3.c
```diff
@@ -0,0 +1,206 @@
...
+    PaStreamParameters outputParameters = { 0 };
+    PaWasapiStreamInfo wasapiStreamInfo = { 0 };
+    PaStream *stream;
+    PaError err;
+    paTestData data;
+    int deviceIndex;
+    FILE *fp;
+    const char *fileName = AC3_FILEPATH;
+    data.buffer = NULL;
+
```
RossBencina, line 111:
```suggestion
long fileSizeBytes;
```

test/patest_wasapi_ac3.c
```diff
@@ -0,0 +1,206 @@
...
+    printf( "reading spdif ac3 raw stream file %s\n", fileName );
+
+    fp = fopen( fileName, "rb" );
+    if( !fp ){
+        fprintf( stderr, "error opening spdif ac3 file.\n" );
+        return -1;
+    }
+    /* get file size */
+    fseek( fp, 0, SEEK_END );
+    data.bufferSampleCount = ftell( fp ) / sizeof(short);
```
RossBencina, line 124:
```suggestion
    long fileSizeBytes = ftell(fp);
    if ((fileSizeBytes % (sizeof(short) * CHANNEL_COUNT)) != 0)
        printf( "warning: input file size is not a multiple of 4 bytes, discarding remainder" );
    data.bufferFrameCount = fileSizeBytes  / (sizeof(short) * CHANNEL_COUNT);
```

test/patest_wasapi_ac3.c
```diff
@@ -0,0 +1,206 @@
...
+    fseek( fp, 0, SEEK_SET );
+
+    /* allocate buffer, read the whole file into memory */
+    data.buffer = (short*)malloc( data.bufferSampleCount * sizeof(short) );
```
RossBencina, line 128:
```suggestion
    data.buffer = (short*)malloc( data.bufferFrameCount * sizeof(short) * CHANNEL_COUNT );
```

test/patest_wasapi_ac3.c
```diff
@@ -0,0 +1,206 @@
...
+    if( !data.buffer ){
+        fprintf( stderr, "error allocating buffer.\n" );
+        return -1;
+    }
+
+    fread( data.buffer, sizeof(short), data.bufferSampleCount, fp );
```
RossBencina, line 134:
```suggestion
    fread( data.buffer, sizeof(short) * CHANNEL_COUNT, data.bufferFrameCount, fp );
```

test/patest_wasapi_ac3.c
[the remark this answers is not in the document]
RossBencina: @gabys999 I have updated my suggestion to indicate what I mean by using bufferFrameCount instead of bufferSampleCount. Phil and I think this would be clearer, if it is compatible with the data streams (are they always multiples of 4-byte SPDIF frames?) On the other hand, if they are AC3 byte-streams, then we need to go the other way and allow for an odd number of bytes, not just an odd number of 16-bit (`short`) samples.

If you accept my suggestion please could you test it against your test files and also update the other test in the same way?

src/hostapi/wasapi/pa_win_wasapi.c
```diff
@@ -2900,7 +2900,7 @@ PaSampleFormat WaveToPaFormat(const WAVEFORMATEXTENSIBLE *fmtext)
         // Currently PA WASAPI is using KSDATAFORMAT_SUBTYPE_IEEE_FLOAT and KSDATAFORMAT_SUBTYPE_PCM
         // therefore this check is reliable in this way.
         if (!memcmp(&fmtext->SubFormat.Data3, &pa_KSDATAFORMAT_SUBTYPE_IEC61937_PCM.Data3,
-            (sizeof(GUID) - sizeof(pa_KSDATAFORMAT_SUBTYPE_IEC61937_PCM.Data1) - sizeof(pa_KSDATAFORMAT_SUBTYPE_IEC61937_PCM.Data2))))
+            (sizeof(GUID) - offsetof(GUID, Data3))
```
RossBencina, line 2903:
```suggestion
            (sizeof(GUID) - offsetof(GUID, Data3))))
```
Fix previous suggestion.

dmitrykos:
@RossBencina to my view we are pushing too many changes in this PR. It could be merged as it has optimal initial implementation and then it could be improved in the next PR, for example in relation to `bufferSampleCount` vs `bufferFrameCount`. 

@gabys999's implementation is also working ok although it relies on the fact implicitly that data is aligned to the frame count.

gabys999:
@dmitrykos @RossBencina 
I apologize for only just seeing the notice. For the past while, I was dealing with another difficult issue on the project. For this issue, we changed the output environment and the Dolby device was temporarily moved away in order to solve the new problem. I'm not sure how long the current issue will take to deal with, my plan is to resolve it in one to two weeks. Next, we also need to debug the wasapi, as the testing department reported a bug in the LTC codec after updating to the new version of portaudio, as well as accessing the Dobly device again

dmitrykos:
I think we really need to merge this PR as is because it was tested and is working correctly. Other things, like improvement to the tests, could be improved in the subsequent iteration in a new PR.

RossBencina:
@dmitrykos @gabys999 ok, we're going to merge it, but I'm not happy with leaving the issues that I have flagged unresolved. Without clarification I don't think the tests are satisfactory as they stand, for the reasons that I have stated in my latest comments.

include/pa_win_wasapi.h
RossBencina: If the approach is to defer to Microsoft documentation there should be a comment with a reference to the MS documentation.

philburk (approved):
Approving for merge as a good step forward.
Note that there is still work to be done. See #816

dmitrykos:
@philburk thank you for improving the PR! I hope @gabys999 will apply changes proposed by @RossBencina in the next iteration when there is possibility to test new changes.

gabys999:
> @philburk thank you for improving the PR! I hope @gabys999 will apply changes proposed by @RossBencina in the next iteration when there is possibility to test new changes.

Okay, we'll restore the Dolby device environment next week, and I'll test @RossBencina  suggested code while we test the 5.1 ac3 passthrough output.

PR #780, opened 2023-03-04
Treat warnings as errors on CI

dechamps:
This pull request adds a new CMake option that makes it easy and convenient for PortAudio developers to turn warnings into errors, making it easier to find problems while iterating on code.

The option is then enabled on CI, ensuring that any future PortAudio changes are warning-free.

The option is not enabled by default because that would run the risk of breaking the build for people who don't have the same build setup as the CI (in particular, compiler type and version), and might trigger different warnings as a result. In particular, we don't want to break people who are merely interested in building PortAudio, but not interested in making changes to PortAudio.

Unfortunately the current code is not free of warnings. We should clean up the codebase to remove all warnings, but that will require some work. If we block enforcement on this work being done, we would not get immediate benefits and we could end up with even more warnings to clean up. So instead, this PR proposes the following approach:

1. We "grandfather" the specific warning types that are already produced by the existing code, i.e. we explicitly stop them from being turned into errors to avoid breaking CI.
2. We turn on enforcement in CI, thus preventing backsliding.
3. We progressively clean up the codebase, removing warning types from the exception list as we go.
4. Once everything is cleaned up we might want to consider increasing the warning level as much as possible.

This PR handles (1) and (2), with (3) and (4) relegated to future PRs.

This PR supersedes #636, which appears to have stalled.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -114,6 +114,7 @@ jobs:
            -DPA_USE_SKELETON=ON
            -DPA_BUILD_TESTS=ON
            -DPA_BUILD_EXAMPLES=ON
+           -DPA_WARNINGS_ARE_ERRORS=ON
            -S .
            -B build
     - name: build
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,6 +15,43 @@ else()
   set(LIBRARY_BUILD_TYPE STATIC)
 endif()
 
+option(PA_WARNINGS_ARE_ERRORS "Turn compiler warnings into errors" OFF)
+if(PA_WARNINGS_ARE_ERRORS)
+    if(MSVC)
+        add_compile_options(/WX
+            # "Grandfathered" warnings that existed before we started enforcement.
+            # Do *NOT* add warnings to this list. Instead, fix your code so that it doesn't produce the warning.
+            # TODO: fix the offending code so that we don't have to exclude specific warnings anymore.
+            /wd4018 # W3 signed/unsigned mismatch
+            /wd4024 # W1 different types for formal and actual parameter
+            /wd4047 # W1 differs in levels of indirection
+            /wd4101 # W3 unreferenced local variable
+            /wd4133 # W3 incompatible types
+            /wd4146 # W2 unary minus operator applied to unsigned type, result still unsigned
+            /wd4244 # W2 conversion possible loss of data
+            /wd4267 # W3 conversion possible loss of data
+            /wd4305 # W1 truncation
+            /wd4477 # W1 format string requires an argument of type, but variadic argument number has type
+            /wd4996 # W3 unsafe/deprecated
+        )
+    else()
+        add_compile_options(-Werror
+            # "Grandfathered" warnings that existed before we started enforcement.
+            # Do *NOT* add warnings to this list. Instead, fix your code so that it doesn't produce the warning.
+            # TODO: fix the offending code so that we don't have to exclude specific warnings anymore.
+            -Wno-error=deprecated-declarations  # https://github.com/PortAudio/portaudio/issues/213 https://github.com/PortAudio/portaudio/issues/641
+            -Wno-error=incompatible-pointer-types
+            -Wno-error=int-conversion
+            -Wno-error=stringop-overflow
+        )
+        if (CMAKE_C_COMPILER_ID MATCHES "Clang")
+             # Don't fail on older clang versions that don't recognize the latest warnings in the list above.
+             # Note that unrecognized warning options are not a fatal error on GCC, and in fact, GCC will choke on this option. Hence the conditional.
+            add_compile_options(-Wno-error=unknown-warning-option)
+        endif()
+    endif()
+endif()
+
 add_library(PortAudio
   ${LIBRARY_BUILD_TYPE}
   src/common/pa_allocation.c


dechamps:
> This PR handles (1) and (2), with (3) and (4) relegated to future PRs.

Actually I just took a quick look through the warnings and added some commits to this PR to fix some of the most basic/trivial ones I could find.

philburk (commented):
These look good. I will review these with Ross.

qa/paqa_devs.c
```diff
@@ -118,7 +118,7 @@ static int gNumFailed = 0;
 static float NextSineSample( PaQaData *data )
 {
     float phase = data->phase + PHASE_INCREMENT;
-    if( phase > M_PI ) phase -= 2.0 * M_PI;
+    if( phase > M_PI ) phase -= 2.0 * (float) M_PI;
```
philburk, line 121: 2.0 is still a double. Maybe use 2.0f.

RossBencina, line 121: This doesn't look right `2.0` is a double, so you're multiplying a double by `(float) M_PI` and you will get a double.

```suggestion
    if( phase > M_PI ) phase -= (float)(2.0 * M_PI);
```

The double precision multiply should happen at compile time.

philburk (changes requested):

RossBencina:
Review is ongoing... If we go with this `PA_WARNINGS_ARE_ERRORS` mechanism I would prefer to see it also implemented for `configure` rather than just removing `-Werror`.

qa/paqa_devs.c
dechamps: Good catch. I'm a bit confused as to why my change got rid of the warning then. Maybe C4305 is only produced with some kinds of expressions but not others, I don't know.

dechamps:
> If we go with this `PA_WARNINGS_ARE_ERRORS` mechanism I would prefer to see it also implemented for `configure` rather than just removing `-Werror`.

To be clear, removing `-Werror` is a good idea regardless of the mechanism that this PR introduces. I don't think we should enable `-Werror` by default, and the current state is weirdly inconsistent as it's only enabled on darwin but no other platform, which makes no sense. I don't think we should wait for a replacement before removing `-Werror`.

I am not going to implement this mechanism in `configure` because I am not familiar enough with it; besides, I'm not sure we want to have two lists of excluded warnings in two different files that then need to be kept in sync. Regardless of whether anyone is able and willing to add this mechanism to `configure`, I don't think that should block this PR because the main goal here is to ensure that warnings are caught in CI. Providing an easy way for developers to enable this mode locally is more of a "nice to have" and should not block this PR. Anyone should feel free to propose a `configure` equivalent in a separate PR.

RossBencina:
Hi @dechamps, Phil is not around this week, and I need to discuss some details with him prior to merge, but in the interest of moving forward here are my current thoughts:

1. The mechanism that you've added to control `-Werror` for CI/CMake seems fine to me. Perhaps other CMakeistas have thoughts on that?
2. The part of this PR that fixes the source code to remove warnings is uncontroversial (assuming that the fixes are correct). It would make sense to split this into a separate PR that can be merged ASAP (more below). [In general I think it's good to keep the easy to merge pre-requisite stuff in separate PRs and the changes that require discussion in their own PR. I get that there's a strong dependency in this case though.]
3. Regarding the `.configure` change. I think it makes sense to split this to a separate PR. Especially given that you said you will not work on it further, the part that changes the `.configure` side of things could/should be considered separately. Also, I'm not confident to make the final call on merging this change -- I thought we'd already made a decision on this, but I don't remember what it was. If you make a separate PR we can discuss there.
4. I'm not super-excited about having a bunch of flags in the build system to suppress specific warnings. I get that we need to have no warnings so that `-Werror` doesn't break. But global suppression should be a last resort. In order of preference: fix the code to not generate warnings, use local `#pragmas` to supress specific warnings at single line of code, use `#pragma` for whole files, suppress for whole project. If you make a separate PR for (2) above, we can merge it and look at all of the warnings that show up in CI and decide what to do on a case-by-case basis.

dechamps:
> The part of this PR that fixes the source code to remove warnings is uncontroversial (assuming that the fixes are correct). It would make sense to split this into a separate PR that can be merged ASAP

Fair enough, I copied the relevant commits to #799. I also kept the commits in this PR for now and will rebase when the other PR is merged.

> Regarding the `.configure` change. I think it makes sense to split this to a separate PR.

Done in #800 and removed from this PR.

> I'm not super-excited about having a bunch of flags in the build system to suppress specific warnings. I get that we need to have no warnings so that `-Werror` doesn't break the CI builds.

To be clear, this PR does not "suppress" any warnings, it just prevents some pre-existing warnings from causing the CI to fail. I agree that this is not the ideal end state, but it is no worse than the current state (where CI doesn't check warnings at all), and is a step in the right direction otherwise. At the risk of sounding like a broken record: I don't think it makes sense to reject PRs that are a net positive, just because they don't go far enough.

To be clear I do intend on looking at the remaining warnings and fix them, but I can't make any promises because I'm not sure how messy/controversial the other warnings are - I only fixed the most trivial ones. In the mean time, this PR ensures that we cannot introduce new warnings aside from those in the allowlist.

RossBencina:
@dechamps Thanks for splitting out the other PRs, it helps streamline the merge review process.

> To be clear, this PR does not "suppress" any warnings, it just prevents some pre-existing warnings from causing the CI to fail.

By "suppress" what I mean is that I do not see any compiler warnings in the output of the "build" step when following the CI links `All checks have passed > Show all checks > CMake build / Windows MSVC (pull_request)  > Details > Windows MSVC > build`

https://github.com/PortAudio/portaudio/actions/runs/4494193910/jobs/7906371354?pr=780

For comparison, all of the compiler warnings are visible on another recent CI run:

https://github.com/PortAudio/portaudio/actions/runs/4537884938/jobs/7996213560?pr=803

I understand that this PR only performs this suppression when `PA_WARNINGS_ARE_ERRORS` is enabled. I think your changes to `CMakeLists.txt` do a good job of documenting which warnings are suppressed and I like the comments that you have included about "Grandfarthered" warning. If we're not going to fix any more warnings right now we should create a ticket for fixing them later.

> I agree that this is not the ideal end state, but it is no worse than the current state (where CI doesn't check warnings at all), and is a step in the right direction otherwise.
[snip]
> To be clear I do intend on looking at the remaining warnings and fix them, but I can't make any promises because I'm not sure how messy/controversial the other warnings are - I only fixed the most trivial ones. In the mean time, this PR ensures that we cannot introduce new warnings aside from those in the allowlist.

Thanks for the clarification.

> At the risk of sounding like a broken record: I don't think it makes sense to reject PRs that are a net positive, just because they don't go far enough.

With respect to my comments above. I am not rejecting this PR. If my intention was to reject this PR I would try to say so as quickly as possible, I don't want to waste anyone's time. At this stage I am reviewing this PR and providing feedback. And as I noted, it's my feedback without consulting Phil -- and Phil and I almost always discuss things prior to a merge.

Phil and I do frequently and explicitly apply the "if it's an improvement with no downside then merge it, even if more work could be done" principle. That said, I think that there is a trade off between "do it right the first time" and "fix it later" especially when "fix it later" could mean it never gets fixed. Also, "do it right the first time" is an opportunity to communicate our expectations to contributors, in the hope that it is automatically done right next time. There is a further trade off depending on the willingness of a contributor to go the extra mile. And I acknowledge that we need to respect everyone's time.

It is difficult for me to communicate nuance clearly and rapidly by text. But I would say that in general my comments range over (at least) the following categories:

- casual observations
- statements of policy or guiding principle (i.e. attempt to get everyone on the same page)
- suggestions for improvement (just a suggestion, take it or leave it)
- requests for improvement (may be negotiable, maybe not)
- concerns for discussion (meaning, I don't yet feel good about things and I'd like to discuss it)
- objections
- strong objections
- veto

Frequently, the PR contributor will know things about the details and background to the PR that I don't know. Or I may be misinformed or confused. Because of this I fully expect that my concerns and objections can be resolved through discussion and/or improvements to the PR.

dechamps:
> By "suppress" what I mean is that I do not see any compiler warnings in the output of the "build" step when following the CI links `All checks have passed > Show all checks > CMake build / Windows MSVC (pull_request) > Details > Windows MSVC > build`

Right. Unfortunately there is no way around this in MSVC, because (as far as I know) there is no way to tell MSVC to stop certain warnings from being turned into errors without suppressing them altogether. (This is not a problem with gcc/clang, which make this possible through the `-Wno-error=` options. MSVC doesn't seem to have an equivalent mechanism.)

Someone building without `PA_WARNINGS_ARE_ERRORS` (i.e. just using the default build options) will see these warnings, though.

Given that it's unlikely anyone will look at the output of a passing CI run with a fine-toothed comb (especially since any new warnings will be drown in the sea of pre-existing ones), I think this is a reasonable compromise?

> With respect to my comments above. I am not rejecting this PR.

Thanks for making your stance clear. I think this is a case where getting this merged would be beneficial even if we never end up fixing the remaining warnings, because preventing *most* warnings seems better than allowing *all* warnings.

RossBencina:
@dechamps #799 is merged now.

dechamps:
Rebased.

Issue #766, opened 2023-02-02
Crash: recursive Pa_Initialize(), FlexAsio


daschuer:
**Describe the bug**
Mixxxx crashes unconditional with an out of stack memory issue when FlexAsio, itself using Portaudio is installed.
This has been reported and drilled down here:
https://github.com/mixxxdj/mixxx/issues/10081

The issue is a recursive Pa_Initialize() loop. 

**To Reproduce**
Start a Portaudio based Application with FlexAsio installed, like Mixxxx.

**Expected behavior**
There should be no crash, FlexAsio should not be available. 

**Actual behavior**
Crash, seevthf original report. 

**Desktop (please complete the following information):**
 - OS windows 
 - PA: 2020-02-20 and 19.7

**Additional context**

I think there should be a guard that prevents recursive calls a secure solution for all such cases. We may also consider to add  FexAudio to the ignore list

dechamps:
This is not supposed to happen.

In dechamps/FlexASIO#47, a similar issue was encountered and I fixed it in dechamps/FlexASIO@b5a24c80ef949b44492efca1dd8d4b44d69d7c1c. The way the fix works is by changing the behavior of the Windows DLL loader using a [side-by-side assembly](https://learn.microsoft.com/en-gb/windows/win32/sbscs/isolated-applications-and-side-by-side-assemblies-portal). The basic idea is that when the FlexASIO DLL is loaded, the Windows DLL loader will link the FlexASIO DLL *with the PortAudio DLL that is bundled with FlexASIO*, even if another DLL of the same name was already loaded in the app process (in this case, by mixxx). This is precisely supposed to prevent exactly this kind of problem, and as far as I know it is a robust fix.

Because the PortAudio DLL bundled with FlexASIO does *not* have ASIO built in, there should be no infinite recursion, and there is no way for FlexASIO to recurse back into itself.

I am not sure why this solution doesn't seem to work in your particular case.

I would encourage you to [file an issue against FlexASIO](https://github.com/dechamps/FlexASIO/issues/new) (please make sure to [include a FlexASIO log](https://github.com/dechamps/FlexASIO#logging)). Things like "mixxx 2.3.3 on x64 windows is crashing right after FlexASIO 1.9 install" in mixxxdj/mixxx#10081 indicate this is something that should be looked at and investigated on the FlexASIO side.

[...]

Issue #606, opened 2021-06-23
Consider moving to C11


jmelas:
Is there any reason to still use C99?
By moving to C11 we gain access to `<stdatomic.h>` and `<threads.h>`
For example issue #573 could use `_ThreadLocal`  as @philburk suggested.
If we agree, I am happy to create a PR that upgrades to C11 support.

Be-ing:
I think ten years is adequate time to update to a new standard.

RossBencina:
We're using C89 not C99. We need to stick to a common subset that MSVC supports.

In September 2020, Microsoft announced preliminary C11 support:
https://devblogs.microsoft.com/cppblog/c11-and-c17-standard-support-arriving-in-msvc/

I think we need to wait another 10 years.

RossBencina:
Phil and I discussed this:

To continue to support even reasonably recent MSVC, we need to continue with C89 for common code and shared header files. 

The only tractable alternative to using C89 everywhere seems to be that we could consider allowing C99 or C11 for host API implementations on Mac and Unix. We already use C++03 for pa_asio.cpp.

Personally I'm not keen to balkanize the code base and use C89 in some places and C99 or C11 in other places. This will be confusing to maintain.

Be-ing:
A reasonable proposal derailed by Microsoft's lackluster standards compliance. :(

> The only tractable alternative to using C89 everywhere seems to be that we could consider allowing C99 or C11 for host API implementations on Mac and Unix. We already use C++03 for pa_asio.cpp.
Personally I'm not keen to balkanize the code base and use C89 in some places and C99 or C11 in other places. This will be confusing to maintain.

It could be maintained automatically by CI if the C standard version is enforced by the build system.

jmelas:
It seems on mac we already use c99?
See file configure
CFLAGS="-std=c99 $CFLAGS $mac_arches $mac_sysroot $mac_version_min"

philburk:
I will try putting a C99 construct into the common code and see if it breaks out GitHub CI.

philburk:
In #663, we use C11 for the CMake build on Mac.

gruenich:
Four years later, I think it is about time to officially move to C11.

> I will try putting a C99 construct into the common code and see if it breaks out GitHub CI.

I did so and it does not break the CI, see #1087 which uses variable declaration in a for loop head (`for (int i = 0; ...)`).

[...]

PR #794, opened 2023-03-14
Pa_Initialize(): Guard against recursive calls.

daschuer:
This fixes a stack overflow when a diver uses portaudio as well like FlexAsio

Fixes: https://github.com/PortAudio/portaudio/issues/766

### DIFF
--- a/src/common/pa_front.c
+++ b/src/common/pa_front.c
@@ -365,7 +365,7 @@ PaError Pa_Initialize( void )
     {
         // a concurrent initialization is already running
         PA_DEBUG(("Attempting to re-enter Pa_Initialize(), aborting!\n"));
-        result = paInternalError;
+        result = paCanNotInitializeRecursively;
     }
     else
     {
@@ -469,6 +469,7 @@ const char *Pa_GetErrorText( PaError errorCode )
     case paCanNotWriteToAnInputOnlyStream:      result = "Can't write to an input only stream"; break;
     case paIncompatibleStreamHostApi: result = "Incompatible stream host API"; break;
     case paBadBufferPtr:             result = "Bad buffer pointer"; break;
+    case paCanNotInitializeRecursively: result = "PortAudio can not initialized recursively"; break;
     default:
         if( errorCode > 0 )
             result = "Invalid error code (value greater than zero)";
--- a/src/common/pa_front.c
+++ b/src/common/pa_front.c
@@ -374,6 +374,7 @@ PaError Pa_Initialize( void )
         // let recursive calls execute the if branch above.
         // This can happen if a driver like FlexAsio itself uses portaudio
         // and avoids a stack overflow in the user application.
+        // https://github.com/PortAudio/portaudio/issues/766
         initializing_ = true;
 
         PA_VALIDATE_TYPE_SIZES;
--- a/src/common/pa_front.c
+++ b/src/common/pa_front.c
@@ -469,7 +469,7 @@ const char *Pa_GetErrorText( PaError errorCode )
     case paCanNotWriteToAnInputOnlyStream:      result = "Can't write to an input only stream"; break;
     case paIncompatibleStreamHostApi: result = "Incompatible stream host API"; break;
     case paBadBufferPtr:             result = "Bad buffer pointer"; break;
-    case paCanNotInitializeRecursively: result = "PortAudio can not initialized recursively"; break;
+    case paCanNotInitializeRecursively: result = "PortAudio can not be initialized recursively"; break;
     default:
         if( errorCode > 0 )
             result = "Invalid error code (value greater than zero)";
--- a/src/common/pa_front.c
+++ b/src/common/pa_front.c
@@ -365,6 +365,7 @@ PaError Pa_Initialize( void )
     else if( initializing_ )
     {
         // a concurrent initialization is already running
+        PA_DEBUG(("Attempting to re-enter Pa_Initialize(), aborting!\n"));
         result = paInternalError;
     }
     else
--- a/src/common/pa_front.c
+++ b/src/common/pa_front.c
@@ -67,7 +67,6 @@
 #include <string.h>
 #include <stdlib.h> /* needed for strtol() */
 #include <assert.h> /* needed by PA_VALIDATE_ENDIANNESS */
-#include <stdbool.h>
 
 #include "portaudio.h"
 #include "pa_util.h"
@@ -155,7 +154,7 @@ static PaUtilHostApiRepresentation **hostApis_ = 0;
 static int hostApisCount_ = 0;
 static int defaultHostApiIndex_ = 0;
 static int initializationCount_ = 0;
-static bool initializing_ = false;
+static int initializing_ = 0;
 static int deviceCount_ = 0;
 
 PaUtilStreamRepresentation *firstOpenStream_ = NULL;
@@ -375,7 +374,7 @@ PaError Pa_Initialize( void )
         // This can happen if a driver like FlexAsio itself uses portaudio
         // and avoids a stack overflow in the user application.
         // https://github.com/PortAudio/portaudio/issues/766
-        initializing_ = true;
+        initializing_ = 1;
 
         PA_VALIDATE_TYPE_SIZES;
         PA_VALIDATE_ENDIANNESS;
@@ -387,7 +386,7 @@ PaError Pa_Initialize( void )
         if( result == paNoError )
             ++initializationCount_;
 
-        initializing_ = false;
+        initializing_ = 0;
     }
 
     PA_LOGAPI_EXIT_PAERROR( "Pa_Initialize", result );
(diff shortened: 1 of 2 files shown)


dechamps:
The proposed code returns `paNoError` on a recursive call. I don't think that's wise - this situation should never be allowed to occur and it's important to bail out of "undefined behaviour" territory as soon as possible instead of attempting to continue. It would be best to return an error if recursion is detected. (In the case of the FlexASIO-related issue, this will make FlexASIO fail to initialize, which is what we want - FlexASIO should not be allowed to continue with a PortAudio library that is not the one that it came bundled with, as that can cause undefined behaviour and lead to more problems.)

A new error code could be added to the `PaErrorCode` enum for this case, or if that seems excessive I guess `paInternalError` could be returned instead.

daschuer:
Done

src/common/pa_front.c
```diff
@@ -360,8 +362,19 @@ PaError Pa_Initialize( void )
...
+    {
+        // a concurrent initialization is already running
+        result = paInternalError;
+    }
     else
     {
+        // set initializing_ here to
+        // let recursive calls execute the if branch above.
+        // This can happen if a driver like FlexAsio itself uses portaudio
+        // and avoids a stack overflow in the user application.
```
dechamps, line 375: It would be a good idea to link to #766 in this code comment.

src/common/pa_front.c
```diff
@@ -360,8 +362,19 @@ PaError Pa_Initialize( void )
         ++initializationCount_;
         result = paNoError;
     }
+    else if( initializing_ )
+    {
+        // a concurrent initialization is already running
```
dechamps, line 367: It might be a good idea to add a debug statement for this:

```c
PA_DEBUG(("Attempting to re-enter Pa_Initialize(), aborting!\n"));
```

src/common/pa_front.c
```diff
@@ -67,6 +67,7 @@
 #include <string.h>
 #include <stdlib.h> /* needed for strtol() */
 #include <assert.h> /* needed by PA_VALIDATE_ENDIANNESS */
+#include <stdbool.h>
```
MichalPetryka, line 70: AFAIK portaudio is supposed to still support C89, while stdbool is C99. I'd replace the bool with an int to keep compatibility.

daschuer: Portaudio is already using stdbool  here:
https://github.com/PortAudio/portaudio/blob/68e963a990da19bb013133dcbad59c2ed8ea0cf9/src/hostapi/audioio/pa_unix_audioio.c#L54

MichalPetryka: Host APIs are permitted to require newer versions as per #606.

daschuer:
Done.

src/common/pa_front.c
```diff
@@ -154,6 +154,7 @@ static PaUtilHostApiRepresentation **hostApis_ = 0;
 static int hostApisCount_ = 0;
 static int defaultHostApiIndex_ = 0;
 static int initializationCount_ = 0;
+static volatile int initializing_ = 0;
```
RossBencina, line 157: What is the rationale for marking this as `volatile`? If there is a good reason, I think that there should be a comment explaining it. If there is no good reason, remove it.

daschuer: I original wanted to use an atomic here to have the  the feature also save when used from different threads. Unfortunately c98 has no interoperable way for this. volatile should work in practice, but it is implementation defined and there is no guarantee. I will just remove it, because my original use case is solved without it.

daschuer: Done

src/common/pa_front.c
```diff
@@ -360,8 +361,21 @@ PaError Pa_Initialize( void )
         ++initializationCount_;
         result = paNoError;
     }
+    else if( initializing_ )
+    {
+        // a concurrent initialization is already running
+        PA_DEBUG(("Attempting to re-enter Pa_Initialize(), aborting!\n"));
+        result = paInternalError;
```
RossBencina, line 368: Please add new error code as described here: https://github.com/PortAudio/portaudio/pull/794#issuecomment-1477029087

daschuer: Done.

src/common/pa_front.c
```diff
@@ -453,6 +469,7 @@ const char *Pa_GetErrorText( PaError errorCode )
     case paCanNotWriteToAnInputOnlyStream:      result = "Can't write to an input only stream"; break;
     case paIncompatibleStreamHostApi: result = "Incompatible stream host API"; break;
     case paBadBufferPtr:             result = "Bad buffer pointer"; break;
+    case paCanNotInitializeRecursively: result = "PortAudio can not initialized recursively"; break;
```
RossBencina, line 472:
```suggestion
    case paCanNotInitializeRecursively: result = "PortAudio can not be initialized recursively"; break;
```

RossBencina (approved):
I approve. I fixed a typo.

RossBencina:
@dechamps @MichalPetryka I think this is ready to merge. Please could you confirm that it fixes #766 or advise if you have any objections. If you're happy use the GitHub review feature to approve it.

dechamps:
The code looks like by me. I'll let @MichalPetryka confirm that it fixes #766 as they're the reporter.

> If you're happy use the GitHub review feature to approve it.

I don't think I have the project permissions to do that. (It's either that or I can't find the button.)

src/common/pa_front.c
```diff
@@ -154,6 +154,7 @@ static PaUtilHostApiRepresentation **hostApis_ = 0;
 static int hostApisCount_ = 0;
 static int defaultHostApiIndex_ = 0;
 static int initializationCount_ = 0;
+static int initializing_ = 0;
```
MichalPetryka, line 157: Should this be marked volatile to prevent the compiler from doing threading unsafe optimizations here?

daschuer: My original version had this volatile due to the assumption that this will prevent the compiler form reordering. 
However this assumption is not guaranteed and fortunately also not necessary, because the compiler cannot move code across external function calls like `InitializeHostApis()` in our case. 

Another aspect is thread safety. The volatile keyword can be part of the solution but, we need also atomic test_and_set() which is not available in C89. Since thread safety is not a requirement here and also not build in any other call the using code is responsible anyway and it is no point in this PR.

MichalPetryka: I think that initialization and shutdown are rare enough that using a mutex here to ensure thread safety would be okay. @RossBencina What do you say?

daschuer: A mutex will create a deadlock in the recursive case. 
IMHO this PR is good enough, protecting Portaudio in concurrent situation is not required for the original problem.

RossBencina:
> I think that initialization and shutdown are rare enough that using a mutex here to ensure thread safety would be okay.

This is a reasonable point. However, PortAudio is a single threaded API, and this patch is intended to prevent recursion on a single thread. According to our API semantics `Pa_Initialize()` may not be called concurrently on two threads, and even our `initializationCount_` mechanism is not thread-safe. Furthermore, we currently have no portable mutex infrastructure so adding a mutex here is a non-trivial change.

So my vote would be to merge this as-is and worry about whether initialize/terminate, or indeed the whole API, should be made thread-safe as a separate issue.

src/common/pa_front.c
philburk: This is a good discussion of thread safety. But currently PortAudio is not thread safe. This fix prevents a crash caused by a stack overflow. It does not solve the thread safety issue but it is a step in the right direction. We can consider full thread safety in a later PR.

RossBencina: > I think that initialization and shutdown are rare enough that using a mutex here to ensure thread safety would be okay.

This is a reasonable point. However, PortAudio is a single threaded API, and this patch is intended to prevent recursion on a single thread. According to our API semantics `Pa_Initialize()` may not be called concurrently on two threads, and even our `initializationCount_` mechanism is not thread-safe. Furthermore, we currently have no portable mutex infrastructure so adding a mutex here is a non-trivial change.

So my vote would be to merge this as-is and worry about whether initialize/terminate, or indeed the whole API, should be made thread-safe as a separate issue.

philburk (approved):

RossBencina:
@MichalPetryka outside your multi-threaded thread-safety concerns could you please confirm that this patch fixes the originally identified issue #766.

RossBencina:
> I don't think I have the project permissions to do that. (It's either that or I can't find the button.)

@dechamps Go to the "Files Changed" tab and at the top right there is a green "Review Changes" button.

RossBencina:
I've created #808 for discussing multi-thread issues.

dechamps (approved):

MichalPetryka (approved):

MichalPetryka:
> @MichalPetryka outside your multi-threaded thread-safety concerns could you please confirm that this patch fixes the originally identified issue #766.

Yeah, it does.

Issue #775, opened 2023-02-26
DirectSound hangs/glitches when using small input host buffers


dechamps:
When running `paloopback -r48000 -s512` with a DirectSound input device:

- If using `--inputLatency 25`, the DirectSound host API hangs (as in, the stream callback is never called);
- If using `--inputLatency 30`, the stream callbacks follow an irregular cadence and glitches are reported;
- If using `--inputLatency 35`, the stream appears to be fine.

This happens regardless of half duplex or full duplex mode. Output seems unaffected.

I've known about this issue since 2018, see dechamps/FlexASIO#29. (I apologize for waiting this long to report it upstream.) Given it was indirectly reported to me by a variety of users over time, I suspect it affects most PortAudio and most Windows versions (at least the modern ones) on most hardware.

The discussion on dechamps/FlexASIO#29 basically sums up the root cause, but just to confirm, here are some results of printf debugging directly on the DirectSound host API code from `paloopback` with the above parameters:

`--inputLatency 100`:

```
StartStream: DSW_StartInput returned = 0x0.
[1191663.76512] IDirectSoundCaptureBuffer_GetCurrentPosition() = 0 capture 0.00000s read 0.00000s size 0.11067s
[1191663.82027] IDirectSoundCaptureBuffer_GetCurrentPosition() = 0 capture 0.00000s read 0.00000s size 0.11067s
[1191663.87735] IDirectSoundCaptureBuffer_GetCurrentPosition() = 0 capture 0.01000s read 0.00000s size 0.11067s
[1191663.93182] IDirectSoundCaptureBuffer_GetCurrentPosition() = 0 capture 0.01000s read 0.00000s size 0.11067s
[1191663.98797] IDirectSoundCaptureBuffer_GetCurrentPosition() = 0 capture 0.02000s read 0.00000s size 0.11067s
[1191664.05155] IDirectSoundCaptureBuffer_GetCurrentPosition() = 0 capture 0.03000s read 0.00000s size 0.11067s
[1191664.11109] IDirectSoundCaptureBuffer_GetCurrentPosition() = 0 capture 0.04000s read 0.00000s size 0.11067s
[1191664.16542] IDirectSoundCaptureBuffer_GetCurrentPosition() = 0 capture 0.04000s read 0.03125s size 0.11067s
INPUT STREAM CALLBACK
INPUT STREAM CALLBACK
[...]

Issue #763, opened 2023-01-21
WDM-KS audio is choppy with some values of suggestedLatency


dechamps:
**Describe the bug**
When using the WDM-KS host API, audio is choppy with some values of `suggestedLatency`. The distribution of good and bad values seems random. This is not specific to small values.

**To Reproduce**
1. Change `examples/paex_sine.c`:
   - Set `device` to the index of a WDM-KS device.
      - The hardware I'm testing with is an ASUS Xonar U7 using the Microsoft standard `usbaudio2.sys` driver (WaveRT).
   - Try various reasonable values for `suggestedLatency`.
3. Run the example.

**Expected behavior**
Smooth audio. In my testing this happens for some values of suggestedLatency, for example `0.030` or `0.031`, but not `0.029` nor `0.100`.

This is an example debug output with `0.031` which works:

```
 0.067: OpenStream:sampleRate = 44100.000000
 0.067: OpenStream:framesPerBuffer = 64
 0.067: Pin create result = 0x00000491
 0.067: Pin create result = 0x00000491
 0.067: Pin create result = 0x00000491
 0.067: Pin create result = 0x00000491
 0.067: Pin create result = 0x00000491
 0.067: Pin create result = 0x00000491
 0.068: Pin create result = 0x00000491
 0.068: Pin create result = 0x00000491
 0.070: Pin create result = 0x00000000
 0.070: Render pin frames: 0
 0.070: Output frames chosen:1367
 0.070: PinGetBuffer: SubType_kNotification
 0.070: Output buffer start = 000001CF37D60000, size = 16404, membarrier = 0
 0.070: BytesPerInputFrame = 0
 0.071: BytesPerOutputFrame = 6
 0.071: Failed to register position register
 0.071: Failed to register rendering position register, using PinGetAudioPositionViaIOCTLRead
Latency: 0.030998s
 0.072: In  buffer len: 0.000 ms
 0.072: Out buffer len: 61.995 ms
 0.072: Timeout = 496 ms
 0.074: PreparePinsForStart = 0
 0.074: Processing thread started!
Play for 5 seconds.
```

**Actual behavior**
Choppy audio with many discontinuities per second. This only happens with some values of `suggestedLatency` but not others, with no apparent logic.

This is an example debug output with `0.029` which is choppy:

```
 0.072: OpenStream:sampleRate = 44100.000000
 0.073: OpenStream:framesPerBuffer = 64
 0.073: Pin create result = 0x00000491
[...]

PR #796, opened 2023-03-18
mme: don't restrict the host buffer to 16-bit

### LINKED ISSUE #139: enable >16 bit output with WMME api

Currently PA/WMME always uses 16 bit host buffers internally.

InitializeWaveHandles() could be updated to try a preferred format and fall back to 16 bit if necessary (it already tries waveformatex before waveformat. it could try waveformatex(preferred) then waveformatex(16) then waveformat(16).

Not sure exactly what else would need to change to support other buffer formats.

We'd want to support linear integer formats. not sure we need to support float since we probably want to use our own floating point converters rather than windows ones.

### LINKED ISSUE #790: MME only supports 16-bit host I/O
In latest master, the MME Host API will only ever create 16-bit integer host buffers:

https://github.com/PortAudio/portaudio/blob/b72559b9bc6173ac7209f4ca476b54e8afa73ef4/src/hostapi/wmme/pa_win_wmme.c#L2358-L2359

https://github.com/PortAudio/portaudio/blob/b72559b9bc6173ac7209f4ca476b54e8afa73ef4/src/hostapi/wmme/pa_win_wmme.c#L2408-L2409

As explained in #774, this causes unnecessary conversions to take place if the user buffer is not 16-bit, as the Windows Audio Engine will convert everything to 32-bit float anyway internally. It also causes precision loss.

The same problem in the DirectSound Host API is being tracked in #54. Having this problem in MME is much worse because that is the default Host API on Windows, meaning users get suboptimal behaviour by default!

I tried to fix it by doing the exact same thing as #774 but there's something fishy going on somewhere as `paloopback` is unhappy with the change:

```
   Default suggested input latency (msec): low = 90.00, high = 180.00
   Default suggested output latency (msec): low = 90.00, high = 180.00
   Running with suggested latency (msec): input = 90.00, out = 90.00
   Amplitudes: left = 0.773455, right = 0.806735
   Loop gain: left = 3.071770, right = 3.071770
ERROR - loop gain is too high. Should be around than 1.0. Please lower output level and/or input gain.
   Background noise = -nan
```

dechamps:
Currently, the MME Host API code only creates 16-bit integer MME buffers. All audio data provided by the user is therefore converted by PortAudio to and from 16-bit, regardless of the user buffer format.

This basically makes it impossible to pass 24-bit audio through the MME Host API.

If the user buffer format is not 16-bit, this also causes pointless conversions to take place, *even if the hardware is running at 16-bit*, because modern Windows versions (Vista+) convert the data to floating point behind the scenes before it is handed off to the hardware. This can lead to silly situations where 32-bit float samples from the user are (lossily) converted to 16-bit by PortAudio, then ended off to Windows via MME, only to be converted back to 32-bit float again, before finally being converted to the format the hardware device is configured to use. This can easily lead to two layers of 16-bit dithering (one from PortAudio, and one from Windows) being piled on top of each other, resulting in an elevated noise floor.

This commit fixes this problem by configuring the MME buffers to use the same format as the user buffer. This should stop PortAudio from converting samples in all cases except paInt8, which is not supported by WAVEFORMATEX (only paUInt8 is). This is pretty much the same idea as #774.

The new code assumes that MME will accept whatever format we throw at it. I feel confident that this is always true on Vista+ regardless of hardware, because starting from Vista MME does not access hardware directly - it always goes through the Windows Audio Engine which supports all formats in both directions (I verified this using paloopback).

On pre-Vista Windows, this should still work all the way back to Windows 98 SE, because MME [goes through KMixer which supports all formats](https://learn.microsoft.com/en-us/windows-hardware/drivers/audio/background-of-non-pcm-support#waveout-api).

Fixes #139

### DIFF
--- a/src/hostapi/wmme/pa_win_wmme.c
+++ b/src/hostapi/wmme/pa_win_wmme.c
@@ -321,6 +321,11 @@ static void PaMme_SetLastSystemError( DWORD errorCode )
     PaMme_SetLastSystemError( errorCode )
 
 
+static int PaMme_IsWindowsVistaOrGreater() {
+    return LOBYTE(LOWORD(GetVersion())) >= 6;
+}
+
+
 /* PaError returning wrappers for some commonly used win32 functions
     note that we allow passing a null ptr to have no effect.
 */
@@ -1771,7 +1776,7 @@ static void InitializeSingleDirectionHandlesAndBuffers( PaWinMmeSingleDirectionH
 static PaError InitializeWaveHandles( PaWinMmeHostApiRepresentation *winMmeHostApi,
         PaWinMmeSingleDirectionHandlesAndBuffers *handlesAndBuffers,
         unsigned long winMmeSpecificFlags,
-        unsigned long bytesPerHostSample,
+        PaSampleFormat hostSampleFormat,
         double sampleRate, PaWinMmeDeviceAndChannelCount *devices,
         unsigned int deviceCount, PaWinWaveFormatChannelMask channelMask, int isInput );
 static PaError TerminateWaveHandles( PaWinMmeSingleDirectionHandlesAndBuffers *handlesAndBuffers, int isInput, int currentlyProcessingAnError );
@@ -1796,14 +1801,13 @@ static void InitializeSingleDirectionHandlesAndBuffers( PaWinMmeSingleDirectionH
 static PaError InitializeWaveHandles( PaWinMmeHostApiRepresentation *winMmeHostApi,
         PaWinMmeSingleDirectionHandlesAndBuffers *handlesAndBuffers,
         unsigned long winMmeSpecificFlags,
-        unsigned long bytesPerHostSample,
+        PaSampleFormat hostSampleFormat,
         double sampleRate, PaWinMmeDeviceAndChannelCount *devices,
         unsigned int deviceCount, PaWinWaveFormatChannelMask channelMask, int isInput )
 {
     PaError result;
     MMRESULT mmresult;
     signed int i, j;
-    PaSampleFormat sampleFormat;
     int waveFormatTag;
 
     /* for error cleanup we expect that InitializeSingleDirectionHandlesAndBuffers()
@@ -1832,9 +1836,7 @@ static PaError InitializeWaveHandles( PaWinMmeHostApiRepresentation *winMmeHostA
             ((HWAVEOUT*)handlesAndBuffers->waveHandles)[i] = 0;
     }
 
-    /* @todo at the moment we only use 16 bit sample format */
-    sampleFormat = paInt16;
-    waveFormatTag = SampleFormatAndWinWmmeSpecificFlagsToLinearWaveFormatTag( sampleFormat, winMmeSpecificFlags );
+    waveFormatTag = SampleFormatAndWinWmmeSpecificFlagsToLinearWaveFormatTag( hostSampleFormat, winMmeSpecificFlags );
 
     for( i = 0; i < (signed int)deviceCount; ++i )
     {
@@ -1852,14 +1854,14 @@ static PaError InitializeWaveHandles( PaWinMmeHostApiRepresentation *winMmeHostA
                         if this fails we fall back to WAVEFORMATEX */
 
                     PaWin_InitializeWaveFormatExtensible( &waveFormat, devices[i].channelCount,
-                            sampleFormat, waveFormatTag, sampleRate, channelMask );
+                            hostSampleFormat, waveFormatTag, sampleRate, channelMask );
                     break;
 
                 case 1:
                     /* retry with WAVEFORMATEX */
 
                     PaWin_InitializeWaveFormatEx( &waveFormat, devices[i].channelCount,
-                            sampleFormat, waveFormatTag, sampleRate );
+                            hostSampleFormat, waveFormatTag, sampleRate );
                     break;
             }
 
@@ -2329,6 +2331,13 @@ static PaError OpenStream( struct PaUtilHostApiRepresentation *hostApi,
     unsigned long outputDeviceCount = 0;                /* contains all devices and channel counts as local host api ids, even when PaWinMmeUseMultipleDevices is not used */
     char throttleProcessingThreadOnOverload = 1;
 
+    /* On Windows Vista and greater, MME will accept any format that can be represented
+       in WAVEFORMATEX and will internally convert if and when necessary.
+       On older Windows versions, the story is less clear, so we restrict ourselves to
+       Int16 for maximum compatibility.
+     */
+    const PaSampleFormat kNativeFormats = PaMme_IsWindowsVistaOrGreater() ?
+        paUInt8 | paInt16 | paInt24 | paInt32 | paFloat32 : paInt16;
 
     if( inputParameters )
     {
@@ -2356,7 +2365,7 @@ static PaError OpenStream( struct PaUtilHostApiRepresentation *hostApi,
         if( result != paNoError ) return result;
 
         hostInputSampleFormat =
-            PaUtil_SelectClosestAvailableFormat( paInt16 /* native formats */, inputSampleFormat );
+            PaUtil_SelectClosestAvailableFormat( kNativeFormats, inputSampleFormat );
 
         if( inputDeviceCount != 1 ){
             /* always use direct speakers when using multi-device multichannel mode */
@@ -2406,7 +2415,7 @@ static PaError OpenStream( struct PaUtilHostApiRepresentation *hostApi,
         if( result != paNoError ) return result;
 
         hostOutputSampleFormat =
-            PaUtil_SelectClosestAvailableFormat( paInt16 /* native formats */, outputSampleFormat );
+            PaUtil_SelectClosestAvailableFormat( kNativeFormats, outputSampleFormat );
 
         if( outputDeviceCount != 1 ){
             /* always use direct speakers when using multi-device multichannel mode */
@@ -2549,7 +2558,7 @@ static PaError OpenStream( struct PaUtilHostApiRepresentation *hostApi,
     {
         result = InitializeWaveHandles( winMmeHostApi, &stream->input,
                 winMmeSpecificInputFlags,
-                stream->bufferProcessor.bytesPerHostInputSample, sampleRate,
+                hostInputSampleFormat, sampleRate,
                 inputDevices, inputDeviceCount, inputChannelMask, 1 /* isInput */ );
         if( result != paNoError ) goto error;
     }
@@ -2558,7 +2567,7 @@ static PaError OpenStream( struct PaUtilHostApiRepresentation *hostApi,
     {
         result = InitializeWaveHandles( winMmeHostApi, &stream->output,
                 winMmeSpecificOutputFlags,
-                stream->bufferProcessor.bytesPerHostOutputSample, sampleRate,
+                hostOutputSampleFormat, sampleRate,
                 outputDevices, outputDeviceCount, outputChannelMask, 0 /* isInput */ );
         if( result != paNoError ) goto error;
     }


src/hostapi/wmme/pa_win_wmme.c
```diff
@@ -1832,8 +1831,6 @@ static PaError InitializeWaveHandles( PaWinMmeHostApiRepresentation *winMmeHostA
             ((HWAVEOUT*)handlesAndBuffers->waveHandles)[i] = 0;
     }
 
-    /* @todo at the moment we only use 16 bit sample format */
```
philburk, line 1835: This was added when supporting AC3 and SPDIF passthrough. We have to make sure we don't break that.

https://github.com/PortAudio/portaudio/commit/1de1b421f97302064ce2e30c7604e185a28da1ee

src/hostapi/wmme/pa_win_wmme.c
```diff
@@ -2329,6 +2326,8 @@ static PaError OpenStream( struct PaUtilHostApiRepresentation *hostApi,
     unsigned long outputDeviceCount = 0;                /* contains all devices and channel counts as local host api ids, even when PaWinMmeUseMultipleDevices is not used */
     char throttleProcessingThreadOnOverload = 1;
 
+    /* These are all the formats that can be represented in WAVEFORMATEX */
+    const PaSampleFormat nativeFormats = paUInt8 | paInt16 | paInt24 | paInt32 | paFloat32;
```
philburk, line 2330: Can this be gated on OS version? Before or after Vista+?

philburk, line 2330: Could be kNativeFormats.

src/hostapi/wmme/pa_win_wmme.c
dechamps: Good point. I just checked and it doesn't look like this change can break this use case, because existing callers have to specify `paInt16` anyway, making this PR a no-op in this case.

https://github.com/PortAudio/portaudio/blob/68e963a990da19bb013133dcbad59c2ed8ea0cf9/examples/paex_wmme_ac3.c#L167

It makes sense that callers would have to use `paInt16`, because otherwise conversion would take place, resulting in corruption of the bitstream. So we can safely assume that every bitstreaming user is already passing `paInt16` here anyway, and will not be broken by this PR.

src/hostapi/wmme/pa_win_wmme.c
dechamps: Why? [Microsoft themselves say](https://learn.microsoft.com/en-us/windows-hardware/drivers/audio/background-of-non-pcm-support#waveout-api) that, all the way back on Windows 98 SE, MME already supported automatic conversions for all these formats through KMixer. I don't see any reason to believe this will break even on pre-Vista Windows.

RossBencina:
PortAudio is still intended to work just fine on pre-Vista systems and also on WINE and ReactOS. Especially in the case of the legacy WMME API this compatibility rather than performance should be the highest priority. What I mean is, supporting output beyond 16-bit is a good improvement, but doing it in a way that assumes Vista+ or is optimised for Vista+ is not an improvement since it degrades backward compatibility. At this point, backward compatibility is the main reason to continue maintaining the MME back end.

> because modern Windows versions (Vista+) convert the data to floating point behind the scenes before it is handed off to the hardware.

Do we have a reliable reference for this claim? I seem to remember even Windows 98 had some floating point support. But I am (fairly) confident that even if Windows converted to float, it would provide transparent passthrough up to 24-bit.

dechamps:
> Especially in the case of the legacy WMME API this compatibility rather than performance should be the highest priority.

Please keep in mind that MME is the default PortAudio Host API on Windows. Which means that any defects in MME will impact anyone who just uses the defaults. That is presumably a large portion of PortAudio users.

In this particular case, this means any PortAudio user who just uses default settings will get their 24-bit or 32-bit float audio lossily downconverted to 16-bit instead of being passed through as-is to the Windows audio engine. This is surprising, inefficient, user-hostile, and completely unnecessary.

> it degrades backward compatibility

There is no evidence that it degrades backward compatiblity.

Can we please only block PRs when there is *clear, actual, substantiated concern*, as opposed to speculation? Rejecting PRs (or demanding an increase in code complexity) based on highly hypothetical "what ifs" (which, even if confirmed, would only affect extremely dubious use cases) is extremely demotivating and not conducive to progress.

> At this point, backward compatibility is the main reason to continue maintaining the MME back end.

I disagree. Besides compatibility, users might still want to use MME today because:

1. It is the default and they don't feel the need to change it, or don't know what they're missing out on.
2. In my experience it is by far the most reliable PortAudio Windows Host API:
   - If you try `paloopback` on MME with various parameters, you will see that it is pretty much impossible to break (aside from the glitches with small buffer sizes, which are not strictly speaking a defect given there is no user expectation that MME will support low latency use cases).
   - DirectSound comes close but still has at least a couple of issues (#775, #770).
   - Last time I checked, the WASAPI Host API was still quite brittle and easy to break under a variety of parameters. I do plan on looking into these but haven't gotten around to it yet.
   - KS is very brittle (see #763) and in any case doesn't really fulfill the same use cases because it uses the device in exclusive mode.
3. The PortAudio MME Host API is the only one to support [outputting to multiple devices at the same time](https://github.com/PortAudio/portaudio/blob/master/include/pa_win_wmme.h#L60).
4. If the user changes their default audio device in Windows, the default MME "wave mapper" device will automatically and transparently redirect the audio, making for a very neat user experience. This is also true for DS (which you also seem to consider "legacy"), but it is not true for WASAPI. PortAudio users might go for MME or DS to get that improved user experience when switching devices in Windows.

Based on the above, I don't think it's fair to assume that users only use MME for compatibility reasons.

> Do we have a reliable reference for this claim? I seem to remember even Windows 98 had some floating point support.

I'm going to copy-paste from [my comment on the equivalent DS PR](https://github.com/PortAudio/portaudio/pull/774#issuecomment-1449046691):

DirectSound hands off the samples to the Windows Audio Engine (through WASAPI internally), where the signal is filtered by APOs, resampled (if necessary), mixed, possibly gain-adjusted if software volume control is in use, passed through [CAudioLimiter](https://www.audiosciencereview.com/forum/index.php?threads/ending-the-windows-audio-quality-debate.19438/) (which does "soft" clipping), and finally converted to integer (with dithering) for handoff to the hardware. (See Microsoft docs e.g. [Windows Audio Architecture](https://learn.microsoft.com/en-us/windows-hardware/drivers/audio/windows-audio-architecture), [Audio Processing Object Architecture](https://learn.microsoft.com/en-us/windows-hardware/drivers/audio/audio-processing-object-architecture), my [APO notes](https://github.com/dechamps/APO), etc.)

> But I am (fairly) confident that even if Windows converted to float, it would provide transparent passthrough up to 24-bit.

If you want a 24-bit sample to make its way untouched throughout the entire shared Windows audio pipeline, that is technically possible since 24-bit and 32-bit float can be converted losslessly to each other. However, in practice that kind of setup would be extremely brittle as you will need to get all the stars to align *just right*:

- In the Windows audio device settings, the device output format must be set to 24-bit and the sample rate and channel count have to match, otherwise automatic conversions will take place.
- No mixing (i.e. no other application playing audio).
- All sAPOs that come with the driver should be disabled.
- If your device ends up relying on [Windows software volume control](https://github.com/dechamps/WinSoftVol), that will need to be set to 100%.
- If your samples get very close to 0 dBFS, your samples will get altered by [CAudioLimiter](https://www.audiosciencereview.com/forum/index.php?threads/ending-the-windows-audio-quality-debate.19438/), which can only be disabled by [hacking around in the registry](https://www.audiosciencereview.com/forum/index.php?threads/ending-the-windows-audio-quality-debate.19438/page-24#post-1194253).

If you do all the above, then you can, with difficulty, get a [bit-perfect pipeline](https://www.audiosciencereview.com/forum/index.php?threads/ending-the-windows-audio-quality-debate.19438/page-24#post-1194466) at 24-bit. If you try the same thing at 16-bit, Windows will also do dithering on top of the above, and that cannot be disabled (as far as I know).

Needless to say, none of the above is particularly user-friendly. For this kind of use case it makes way more sense to use an exclusive backend such as WASAPI Exclusive or WDM-KS, which are precisely designed to be bit-perfect and are the right tools for this job.

RossBencina:
Sorry for the delay, here's our position statement. There are two issues:

1. Given that MME is the legacy API it should always work. As it stands, this patch will lead to failures if the OS can't handle floats (or whatever >16bit format the user passes to PortAudio). We should implement error handling with a fallback to 16-bit. Phil and I agree on this.

2. There is a second issue, which goes way beyond this PR and which Phil and I have different opinions about, which is whether PortAudio should always be responsible for float<->integer conversions. As it stands this patch lets the OS perform these conversions, and I would like PortAudio to always do these conversions so that the scaling factors, dithering and clipping are controlled and predictable across all PortAudio host APIs. Phil thinks it's best to pass through floats to the OS.

In the case of this patch, the 2nd issue is further complicated by the fact that _any_ float passthrough is better than downscaling to 16bit. Phil and I agree on this. That said, I'd prefer the conversions to be done in PA (even integer <-> integer conversions) and Phil would prefer to pass floats through to the OS if the OS accepts them.

RossBencina:
I've created #825 to track the broader policy issue.

RossBencina:
> There is no evidence that it degrades backward compatiblity.

I have been writing code against the MME API since 1995. And I remember when float was not supported, and when it was added.

philburk:
The current code seems dependent on WAVEFORMATEX.
it will not run on systems prior to WAVEFORMATEX.
So an essential question is whether all systems that support WAVEFORMATEX support FLOAT.
If so then this code will not cause a regression.

dechamps:
> > There is no evidence that it degrades backward compatiblity.
> 
> I have been writing code against the MME API since 1995. I remember when float was not supported, and when it was added.

Sure. You said earlier in this thread that you believe float was supported all the way back to Windows 98. I have no reason to doubt you. Especially since [MSDN concurs](https://learn.microsoft.com/en-us/windows-hardware/drivers/audio/background-of-non-pcm-support#waveout-api) (Windows 98 SE to be precise). 

I'm not sure what point you are trying to make by saying that, though. If you're saying that my statement "There is no evidence that it degrades backward compatibility" is technically incorrect, then sure, no problem, let me rephrase my statement right away and correct the record:

*There is no evidence that this change degrades backward compatibility on any OS that any user could possibly care about in 2023.*

There. Can we agree on this?

If not, can you please cite one, even just one, user who would reasonably expect code written in 2023 to run out-of-the-box on Windows 95? I certainly can't. In fact, I don't know, nor know people who know, anyone who even still *runs* Windows 95 in any capacity. I can speculate that there might be an extremely small number of people out there who are forced to use Windows 95 because they want to run very old software or very old hardware. Assuming these people exist, what reason could they possibly have to want to run a 2023 version of PortAudio on these systems??

> Given that MME is the legacy API it should always work. As it stands, this patch will lead to failures if the OS can't handle floats (or whatever >16bit format the user passes to PortAudio). We should implement error handling with a fallback to 16-bit.

Honestly I am *extremely* reluctant to add a fallback, which in itself is a non-trivial error handling mechanism with its own complexity and potential for bugs, to guard against a situation that will never occur in practice outside of systems that (1) are *extremely* rare, if they even still exist and (2) have no reason to use a modern version of PortAudio in the first place. It's just completely pointless code, and I don't like writing pointless code.

Right now I'm torn between two options: (1) dropping this patch and telling my users they can't use 24-bit with MME or (2) forking PortAudio for my own project so that I can fix these bugs without having to spend precious time and energy running around in circles in threads like this one. Option (3), doing what you asked, i.e. writing pointless code (under protest) that is never going to run on any platform that any reasonable user could possibly want to use, well… that option comes in a very distant third.

> So an essential question is whether all systems that support WAVEFORMATEX support FLOAT.

[MSDN says](https://learn.microsoft.com/en-us/windows/win32/api/mmeapi/ns-mmeapi-waveformatex) `WAVEFORMATEX` was introduced with Windows 2000. [MSDN also says](https://learn.microsoft.com/en-us/windows-hardware/drivers/audio/background-of-non-pcm-support#waveout-api) that floating-point support was added to KMixer in Windows 98 SE and Windows 2000. So I believe the answer to your question is yes.

src/hostapi/wmme/pa_win_wmme.c
```diff
@@ -1796,14 +1796,13 @@ static void InitializeSingleDirectionHandlesAndBuffers( PaWinMmeSingleDirectionH
 static PaError InitializeWaveHandles( PaWinMmeHostApiRepresentation *winMmeHostApi,
         PaWinMmeSingleDirectionHandlesAndBuffers *handlesAndBuffers,
         unsigned long winMmeSpecificFlags,
-        unsigned long bytesPerHostSample,
+        PaSampleFormat sampleFormat,
```
RossBencina, line 1799: Please change this parameter to be named `hostSampleFormat`. I think that would be clearer.

RossBencina:
> Honestly I am extremely reluctant to add a fallback, which in itself is a non-trivial error handling mechanism with its own complexity and potential for bugs

I agree with this. This is the main reason I didn't get around to implementing anything more complex than hardcoding `paInt16`.

philburk:
It seems Ross is worried about breaking apps on old OS versions that do not support FLOAT.
And Etiene is worried about apps on modern OSs not being able to take advantage of FLOAT when it is available.

I propose that we make this change conditional on OS version. Something like this pseudo-code:

    if (version < WINDOWS_VISTA)  {
        kNativeFormats = paInt16;
    } else {
        kNativeFormats = paUInt8 | paInt16 | paInt24 | paInt32 | paFloat32;
    }
    
 I don't know how to check the version on Windows. Ross and I looked at this and it is non-trivial.
 See "pa_win_wasapi.c".

RossBencina:
Phil's proposal sounds ok to me, but...

> I don't know how to check the version on Windows. Ross and I looked at this and it is non-trivial.

If we could just use `GetVersion()` it would be a simple change, but right now Microsoft seems to be strongly advising against it, and Dmitry is dynamically loading the symbol in `pa_win_wasapi.c`.

https://learn.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-getversion

dechamps:
> I agree with this. This is the main reason I didn't get around to implementing anything more complex than hardcoding `paInt16`.

So when you wrote this code, you had to make a choice between only passing `paInt16` or passing all possible formats. You chose the former for compatibility reasons. I agree this likely made sense when you made this decision 16+ years ago. It doesn't make sense today.

> I propose that we make this change conditional on OS version.

Thanks Phil. I can get behind this. I still believe this is pointless, and it will still result in unnecessary conversions for people using PortAudio on, say, XP. But it's not like I really care about XP, and the extra code complexity is trivial, so I'm fine with this.

> If we could just use `GetVersion()` it would be a simple change, but right now Microsoft seems to be strongly advising against it

There is precedent for using `GetVersionEx()` in DirectSound:

https://github.com/PortAudio/portaudio/blob/929e2e8f7af281c5eb4fa07758930d542ec43d97/src/hostapi/dsound/pa_win_ds.c#L348

In practice I don't think using `GetVersion()` is really a problem for this use case. Microsoft warns about using `GetVersion()` because it cannot be used to detect Windows versions greater than Windows 8.1, but that's perfectly fine here. [This other MSDN doc](https://learn.microsoft.com/en-us/windows/win32/sysinfo/targeting-your-application-at-windows-8-1) explicitly states (emphasis mine):

> In Windows 8.1 and later, the [GetVersion](https://learn.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-getversion) and [GetVersionEx](https://learn.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-getversionexa) functions have been deprecated. As of Windows 10, the [VerifyVersionInfo](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-verifyversioninfoa) function is also deprecated. **You can still call the deprecated functions**, but if your application doesn't specifically target Windows 8.1 or later, then the functions will return the Windows 8 version (6.2).

> Dmitry is dynamically loading the symbol in `pa_win_wasapi.c`.

I suspect this is purely because `GetVersion()` and `GetVersionEx()` can only be used in "desktop" apps, not in "Universal Windows apps" (a.k.a WinRT).  I don't think this is a problem for MME or DS because PortAudio never promised these would be usable in UWP apps. In fact, I suspect MME itself cannot be used in UWP apps ([MSDN says](https://learn.microsoft.com/en-us/windows/win32/api/mmeapi/nf-mmeapi-waveoutopen) "desktop apps only" for MME functions).

I'll update the PR to use `GetVersion()` when I get the chance.

dechamps:
> I'll update the PR to use `GetVersion()` when I get the chance.

Done.

philburk (approved):
This looks great! I'll let Ross give final approval.

dmitrykos:
> > Dmitry is dynamically loading the symbol in `pa_win_wasapi.c`.
> 
> I suspect this is purely because `GetVersion()` and `GetVersionEx()` can only be used in "desktop" apps, not in "Universal Windows apps" (a.k.a WinRT).

WASAPI's implementation does not attempt to load dynamically GetVersion when compiled for UWP, that part of the code is excluded with `ifndef` and is hardcoded to return either `WINDOWS_10_SERVER2016` or `WINDOWS_8_SERVER2012`: https://github.com/PortAudio/portaudio/blob/929e2e8f7af281c5eb4fa07758930d542ec43d97/src/hostapi/wasapi/pa_win_wasapi.c#L1348-L1352

The reason of the complex implementation of WASAPI's `GetWindowsVersion` which is loading dynamically `RtlGetVersion`, or `GetVersion`, or `VerifyVersionInfoA` is to be able to detect Windows Vista, Windows 7, Windows 8, Windows 10 when application is compiled as Win32. Unfortunately Microsoft made it really fragmented by deprecating functions and introducing new.

> MME itself cannot be used in UWP apps

Indeed, only WASAPI.

I noticed from discussion that DS and now MME host implementations will use `GetVersion` which is very unportable. My proposal is to extract WASPI's `GetWindowsVersion` https://github.com/PortAudio/portaudio/blob/929e2e8f7af281c5eb4fa07758930d542ec43d97/src/hostapi/wasapi/pa_win_wasapi.c#L1233 and move it to the new `/os/win/pa_win_version.c` file which would serve portable and reliable detection of Windows version for all host APIs?

If this proposal is acceptable I could take this task and propose the PR for that covering WASAPI, MME and DS host APIs. Please let me know your opinion.

philburk:
@dmitrykos - I think it would be great to have a portable GetWindowsVersion() available as a utility.
Thank you for offering.  We will probably go ahead and merge this PR. We can then change it to call your GetWindowsVersion()  when it is ready.

dmitrykos:
@dechamps you could have a look at implementation https://github.com/PortAudio/portaudio/blob/929e2e8f7af281c5eb4fa07758930d542ec43d97/src/hostapi/wasapi/pa_win_wasapi.c#L1233
it is quite self explanatory.

My previous comment about portability issues was not related to `GetVersion` usage particularly for this PR, portability issues start from Windows Vista and up.

RossBencina:
Thanks for your patience. We're going to merge this, but two comments:

1 A minor detail: I'm not sure how we arrived at windows Vista as the target, I would have thought 7 was fine. Not a big deal, but it does seem strange to me.
2. I did raise this earlier, but maybe I was not clear enough, the latest Microsoft docs contain this caveat on `GetVersion`:

> `GetVersion` may be altered or unavailable for releases after Windows 8.1. Instead, use the [Version Helper functions](https://learn.microsoft.com/en-us/windows/desktop/SysInfo/version-helper-apis). For Windows 10 apps, please see [Targeting your applications for Windows](https://learn.microsoft.com/en-us/windows/win32/sysinfo/targeting-your-application-at-windows-8-1).

https://learn.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-getversion

Hopefully Dmitry's version will help with this issue.

RossBencina (approved):

dechamps:
Thanks for finally merging this. Progress!

> 1 A minor detail: I'm not sure how we arrived at windows Vista as the target, I would have thought 7 was fine. Not a big deal, but it does seem strange to me.

Vista is the first Windows version to include the modern Windows audio stack, so it seems natural to use it as the boundary.

> 2. I did raise this earlier, but maybe I was not clear enough, the latest Microsoft docs contain this caveat on `GetVersion`:
> 
> > `GetVersion` may be altered or unavailable for releases after Windows 8.1. Instead, use the [Version Helper functions](https://learn.microsoft.com/en-us/windows/desktop/SysInfo/version-helper-apis). For Windows 10 apps, please see [Targeting your applications for Windows](https://learn.microsoft.com/en-us/windows/win32/sysinfo/targeting-your-application-at-windows-8-1).

I did notice, and I did respond to your point in a previous comment. I strongly suspect that sentence in the Microsoft docs is just poorly phrased - Microsoft almost certainly means to say that `GetVersion` cannot be used to distinguish between post-8.1 Windows versions, not that `GetVersion` itself would suddenly stop working. It would make no sense for them to break the *existing* behaviour of `GetVersion` as it would compromise application compatibility for zero benefit to them. Historically, Microsoft tends to take Windows app compatibility very seriously.

Issue #35, opened 2006-11-22
some conversion functions are not implemented in pa_converters.c


PortAudio-admin:
*Issue created by @RossBencina*

Conversion functions in src/common/pa_converters.c copy audio data between PortAudio native buffers and host buffers.

https://www.assembla.com/code/portaudio/subversion/nodes/1930/portaudio/trunk/src/common/pa_converters.c

Some conversion functions are marked IMPLEMENT ME and do not pass any audio, some are marked REVIEW and may not currently work as expected.

All IMPLEMENT_ME functions are dithered variants of existing non-dither functions, so the non-dithered versions can be used as a starting point.

See also /test/patest_converters.c

dechamps:
Hello from 2023. Just FYI, this is still an issue for a small number of dithered converters. Here's the output of `patest_converters` on latest master (4bccdfbb68a509e2a96d916046d911959f7d935c):

```
= Sine wave in, something out =


== flags = paNoFlag ==

=== Output contains non-zero data ===
Key: . - pass, X - fail
{{{
in|  out:      f32     i32     i24     i16      i8     ui8
f32             .       .       .       .       .       .
i32             .       .       X       .       .       X
i24             .       .       .       .       .       X
i16             .       .       .       .       X       X
 i8             .       .       .       .       .       .
ui8             .       .       .       .       .       .
}}}

=== Combined dynamic range (src->dest->float32) ===
Key: Noise amplitude in dBfs, X - fail (either above failed or dest->float32 failed)
{{{
in|  out:      f32     i32     i24     i16      i8     ui8
 f32          -inf  -180.6  -138.5   -79.5   -31.5   -31.5
 i32        -144.5  -144.5     X     -85.9   -42.1     X
 i24        -138.5  -138.5  -138.5   -86.2   -42.1     X
 i16         -84.8   -84.8   -84.8   -84.8     X       X
  i8         -36.6   -36.6   -36.6   -36.6   -36.6   -36.6
 ui8         -36.4   -36.4   -36.4   -36.4   -36.4   -36.4
}}}

== flags = paClipOff ==

=== Output contains non-zero data ===
Key: . - pass, X - fail
{{{
in|  out:      f32     i32     i24     i16      i8     ui8
f32             .       .       .       .       .       .
i32             .       .       X       .       .       X
i24             .       .       .       .       .       X
i16             .       .       .       .       X       X
 i8             .       .       .       .       .       .
ui8             .       .       .       .       .       .
}}}

[...]

PR #798, opened 2023-03-20
Issue #35 - implemented missing Int32_to_Int24_Dither function

whyismynamerudy:
Title. Implemented the missing function "Int32_To_Int24_Dither" in "pa_converters.c" using the specified coding guidelines.

I had a few questions:
1. Is doing conversions differently for Little Endian vs Big Endian machines required for all conversion functions? I noticed that some functions did make the distinction while some did not, and I cannot figure out why that is the case.
2. What is the purpose of the sourceStride and the destinationStride variables in the conversion functions? Based off of how the other functions were implemented I gather that they're used for navigating across the bits of the binary numbers (I think?) but I am not sure. If that is the case, then why is it so that some strides are directly added onto src while others (such as the sourceStride in the function Int24_To_Float32) are multiplied by a certain number before adding them onto src?

### DIFF
--- a/src/common/pa_converters.c
+++ b/src/common/pa_converters.c
@@ -943,6 +943,7 @@ static void Int32_To_Int24_Dither(
 
     while ( count-- )
     {
+        /* REVIEW */
         dither = PaUtil_Generate16BitTriangularDither(ditherGenerator);
         *dest = (signed char) ((((*src) >> 1) + dither) >> 7);
         
--- a/src/common/pa_converters.c
+++ b/src/common/pa_converters.c
@@ -935,13 +935,20 @@ static void Int32_To_Int24_Dither(
     void *sourceBuffer, signed int sourceStride,
     unsigned int count, struct PaUtilTriangularDitherGenerator *ditherGenerator )
 {
-    (void) destinationBuffer; /* unused parameters */
-    (void) destinationStride; /* unused parameters */
-    (void) sourceBuffer; /* unused parameters */
-    (void) sourceStride; /* unused parameters */
-    (void) count; /* unused parameters */
     (void) ditherGenerator; /* unused parameters */
-    /* IMPLEMENT ME */
+
+    PaInt32 *src = (PaInt32*)sourceBuffer;
+    signed char *dest = (signed char*)destinationBuffer;
+    PaInt32 dither;
+
+    while ( count-- )
+    {
+        dither = PaUtil_Generate16BitTriangularDither(ditherGenerator);
+        *dest = (signed char) ((((*src) >> 1) + dither) >> 7);
+        
+        src += sourceStride;
+        dest += destinationStride;
+    }
 }
 
 /* -------------------------------------------------------------------------- */


src/common/pa_converters.c
```diff
@@ -935,13 +935,21 @@ static void Int32_To_Int24_Dither(
...
+
+    PaInt32 *src = (PaInt32*)sourceBuffer;
+    signed char *dest = (signed char*)destinationBuffer;
+    PaInt32 dither;
+
+    while ( count-- )
+    {
+        /* REVIEW */
+        dither = PaUtil_Generate16BitTriangularDither(ditherGenerator);
+        *dest = (signed char) ((((*src) >> 1) + dither) >> 7);
```
RossBencina, line 948: - review dither scaling
- assign to the three bytes of dest: `dest[0]`, `dest[1]`, `dest[2]`
- will need separate big/little endian cases

src/common/pa_converters.c
```diff
@@ -935,13 +935,21 @@ static void Int32_To_Int24_Dither(
...
+        
+        src += sourceStride;
+        dest += destinationStride;
```
RossBencina, line 951: * 3

RossBencina:
@whyismynamerudy I'll keep helping out here, but I have marked this PR as a draft so that other people don't feel the need to review it.

For the future, here are instructions for marking preliminary work in PRs as draft:

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-stage-of-a-pull-request

dechamps:
@whyismynamerudy I would strongly recommend you use the program in `test/patest_converters` to test your code. If I run that test on your code, it spits out `5.5` as measured dynamic range for i32→i24, which is clearly wrong and indicates your converter is broken, most likely for the reasons @RossBencina mentioned.

> Is doing conversions differently for Little Endian vs Big Endian machines required for all conversion functions? I noticed that some functions did make the distinction while some did not, and I cannot figure out why that is the case.

If you look closely, you will notice that the only converters that care about endianness are the converters that deal with 24-bit data. Other converters don't need to care about endianness because they can use the standard C types to represent the samples: Int16 (`short`), Float (`float`). (And of course the question of endianness doesn't apply to (U)Int8 at all.) When manipulating numbers using the standard C types, C takes care of the endianness for you by always reading and writing the data using the endianness of the host. But, sadly, there is no standard C type for a 24-bit integer. Code that wants to read and write 24-bit integers has no choice but to read or write the 3 individual bytes manually. Such code has to pay attention to endianness in order to deal with the bytes in the correct order.

> What is the purpose of the sourceStride and the destinationStride variables in the conversion functions?

The term "stride" in this context means [step size](https://en.wikipedia.org/wiki/Stride_of_an_array): how many elements to skip on every iteration. This matters for interleaved audio (i.e. audio where the samples from all channels are interleaved, as in `[left1 right 1 left2 right2 etc]` as opposed to `[left1 left2 etc] [right1 right2 etc]`), which is the most common case. Depending on how the dithering is done, it might make sense to dither one channel at a time, so the converter is called on each channel successively. On each converter call, the stride parameter is used to ensure the converter "skips over" the other channels.

In this particular piece of code, the stride is expressed in samples. For non-interleaved audio, the stride is always 1. For interleaved mono audio, the converter will be called with a stride of 1; for interleaved stereo (2 channels), the stride is 2; for interleaved 5.1 (6 channels), the stride is 6; etc.

> why is it so that some strides are directly added onto src while others (such as the sourceStride in the function Int24_To_Float32) are multiplied by a certain number before adding them onto src

This again has to do with C number types and how C does pointer arithmetic, and again you will notice that only 24-bit converters do this kind of multiplication.

In converters that can get away with using standard C types (e.g. `char`, `short`, `float`) to represent the samples, they can simply add the `sourceStride` to a `T*` pointer because the stride is expressed in samples (not bytes), and in C that means advancing the pointer to by `sourceStride * sizeof(T)` bytes, so there is no need to multiply - C already does the multiplication for you based on the pointer type. But 24-bit converters cannot use standard C types and have to deal with the individual bytes directly (see above), and therefore have to multiply the stride by the sample size in bytes (in this case 3) themselves when doing pointer arithmetic.

whyismynamerudy:
@RossBencina @dechamps Thank you both for you feedback on my code and answering my questions. I apologize for not being as active for the past 2 months - I had exams, but now that they are over I look forward to contributing to this project again. 

Currently, I'm looking through the dither generation functions to see how I would be able to implement `PaUtil_Generate24BitTriangularDither` - it looks like an interesting challenge. Initially I struggled to understand why certain things were being done in the other dither generation functions, but as I spend more time trying to understand it, it begins to make more sense. I'll aim to include the function in my next pull request.

As for the conversion functions, I am working on implementing the feedback I've gotten. For now (until the `PaUtil_Generate24BitTriangularDither` function I hope to implement gets approved), I will manually try to scale the dither and use the temporary 32 bit int that Ross suggested earlier to spread it out among the 3 bytes of dest depending on the endianness.

Again, thanks to both of you for the feedback - it's helping me learn quite a bit about coding and tackling issues I've never encountered before.

whyismynamerudy:
Pushed the changes I discussed in my previous comment. 

If the `PaUtil_Generate24BitTriangularDither` function gets approved, I may need help coming up with the function documentation comment - I'm unsure as to what to include in it.

dmitrykos:
@whyismynamerudy check how dither is applied in this example:
https://github.com/PortAudio/portaudio/blob/f1733cc39041337122507002a701af5e94631354/src/common/pa_converters.c#L1280

You do not need to shift dither in your implementation.

dmitrykos:
I think you need to remove `PaUtil_Generate24BitTriangularDither` from this PR because it is currently unused (dangling code) and due to that PR will not be accepted. Just keep it for the future, if it ever makes sense to use it.

src/common/pa_converters.c
```diff
@@ -935,13 +935,31 @@ static void Int32_To_Int24_Dither(
...
-    (void) destinationStride; /* unused parameters */
-    (void) sourceBuffer; /* unused parameters */
-    (void) sourceStride; /* unused parameters */
-    (void) count; /* unused parameters */
     (void) ditherGenerator; /* unused parameters */
-    /* IMPLEMENT ME */
+
+    PaInt32 *src = (PaInt32*)sourceBuffer;
+    signed char *dest = (signed char*)destinationBuffer;
+    signed char *scaledDitherResult = (signed char*)destinationBuffer;
```
philburk, line 942: ScaledDitherResult needs to be an PaInt32 so that it can hold the modified 32-bit value. If you shift signed char by 24, you'll have nothing left.

src/common/pa_converters.c
```diff
@@ -935,13 +935,31 @@ static void Int32_To_Int24_Dither(
...
+    PaInt32 dither;
+
+    while ( count-- )
+    {
+        /* REVIEW */
+        dither = PaUtil_Generate16BitTriangularDither(ditherGenerator);
+        *scaledDitherResult = (signed char) ((((*src) >> 1) + dither) >> 7);
+
+#if defined(PA_LITTLE_ENDIAN)
+        dest[0] = (unsigned char)(*scaledDitherResult >> 8);
```
philburk, line 952: You already shifted by eight up above so you can shift by eight less when doing char writes.

src/common/pa_converters.c
```diff
@@ -935,13 +935,31 @@ static void Int32_To_Int24_Dither(
...
-    (void) destinationBuffer; /* unused parameters */
-    (void) destinationStride; /* unused parameters */
-    (void) sourceBuffer; /* unused parameters */
-    (void) sourceStride; /* unused parameters */
-    (void) count; /* unused parameters */
     (void) ditherGenerator; /* unused parameters */
-    /* IMPLEMENT ME */
+
+    PaInt32 *src = (PaInt32*)sourceBuffer;
+    signed char *dest = (signed char*)destinationBuffer;
```
philburk, line 941: Dest should be unsigned.

philburk (changes requested):
Thanks for tackling this. Data type conversions, particularly ones involving packed 24-bit data, are very tricky.

philburk (changes requested):
Thanks for tackling this. Data type conversions, particularly ones involving packed 24-bit data, are very tricky.

whyismynamerudy:
Thanks for the feedback! I've implemented the changes suggested by @philburk and removed `PaUtil_Generate24BitTriangularDither` for now - I still have the code, and I can put it back if needed.

If `Int32_To_Int24_Dither` looks good now, I'll begin working on some of the other unimplemented conversion functions.

src/common/pa_converters.c
```diff
@@ -935,13 +935,30 @@ static void Int32_To_Int24_Dither(
...
-    (void) destinationStride; /* unused parameters */
-    (void) sourceBuffer; /* unused parameters */
-    (void) sourceStride; /* unused parameters */
-    (void) count; /* unused parameters */
-    (void) ditherGenerator; /* unused parameters */
-    /* IMPLEMENT ME */
+
+    PaInt32 *src = (PaInt32*)sourceBuffer;
+    unsigned char *dest = (unsigned char*)destinationBuffer;
+    PaInt32 *scaledDitherResult = (PaInt32*)destinationBuffer;
```
RossBencina, line 941: The most obvious red flag here is that you're casting a non-aligned ptr to a 32-bit aligned pointer. But also, why is the scaled dither result being stored in the destination buffer? In any case there's no reason to be storing a temporary result in the destination buffer.

```suggestion
```

src/common/pa_converters.c
```diff
@@ -935,13 +935,30 @@ static void Int32_To_Int24_Dither(
...
+    PaInt32 dither;
+
+    while ( count-- )
+    {
+        /* REVIEW */
+        dither = PaUtil_Generate16BitTriangularDither(ditherGenerator);
```
RossBencina, line 947:
```suggestion
        PaInt32 dither = PaUtil_Generate16BitTriangularDither(ditherGenerator);
```

src/common/pa_converters.c
```diff
@@ -935,13 +935,30 @@ static void Int32_To_Int24_Dither(
...
-    (void) sourceBuffer; /* unused parameters */
-    (void) sourceStride; /* unused parameters */
-    (void) count; /* unused parameters */
-    (void) ditherGenerator; /* unused parameters */
-    /* IMPLEMENT ME */
+
+    PaInt32 *src = (PaInt32*)sourceBuffer;
+    unsigned char *dest = (unsigned char*)destinationBuffer;
+    PaInt32 *scaledDitherResult = (PaInt32*)destinationBuffer;
+    PaInt32 dither;
```
RossBencina, line 942:
```suggestion
```

src/common/pa_converters.c
```diff
@@ -935,13 +935,30 @@ static void Int32_To_Int24_Dither(
...
+
+    while ( count-- )
+    {
+        /* REVIEW */
+        dither = PaUtil_Generate16BitTriangularDither(ditherGenerator);
+        *scaledDitherResult = (PaInt32) ((((*src) >> 1) + dither) >> 7);
```
RossBencina, line 948:
```suggestion
        PaInt32 scaledDitherResult = (PaInt32) ((((*src) >> 1) + dither) >> 7);
```

src/common/pa_converters.c
```diff
@@ -935,13 +935,30 @@ static void Int32_To_Int24_Dither(
...
+
+#if defined(PA_LITTLE_ENDIAN)
+        dest[0] = (unsigned char)(*scaledDitherResult);
```
RossBencina, line 951: remove the `*` now.

src/common/pa_converters.c
```diff
@@ -1055,16 +1072,19 @@ static void Int32_To_UInt8_Dither(
...
-    (void)ditherGenerator; /* unused parameter */
+    PaInt32 *src = (PaInt32*)sourceBuffer;
+    unsigned char *dest =  (unsigned char*)destinationBuffer;
+    PaInt32 dither;
 
     while( count-- )
     {
-        /* IMPLEMENT ME */
+        /* REVIEW */
+        dither = PaUtil_Generate16BitTriangularDither( ditherGenerator );
```
RossBencina, line 1082: declare variable here as in previous function.

RossBencina (changes requested):
I've made some suggestions.

Have you run this on the test program?

src/common/pa_converters.c
```diff
@@ -1073,16 +1072,19 @@ static void Int32_To_UInt8_Dither(
...
 
-        /* src += sourceStride;
-        dest += destinationStride; */
+        *dest = (unsigned char)((((*src) >> 1) + dither) >> 23);
```
philburk, line 1084: UInt8 audio is centered around 128 and ranges from 0 to 255.
So silence would be all 128s.

philburk (changes requested):
Good progress.

src/common/pa_converters.c
philburk: To clarify, you need to add an offset to convert Int32 to Uint8 format.

philburk:
@whyismynamerudy - Good progress so far. Do you want to finish this up?

whyismynamerudy:
> @whyismynamerudy - Good progress so far. Do you want to finish this up?

Hi @philburk , sorry I was inactive for the past few months. I got busy with some other things over summer. I'm getting back to this and I hope to have provided an update over the next two weeks.

PR #799, opened 2023-03-22
Fix some low-hanging C warnings

dechamps:
This PR is a spin-off of #780 that focuses on fixing a number of trivial low-hanging fruit warnings.

Note that this PR only fixes a few classes of warnings, not all of them. It's a step in the right direction, though.

### DIFF
--- a/test/patest_dsound_find_best_latency_params.c
+++ b/test/patest_dsound_find_best_latency_params.c
@@ -46,7 +46,7 @@
 //#include <mmsystem.h>   /* required when using pa_win_wmme.h */
 
 #include <conio.h>      /* for _getch */
-
+#include <tchar.h>
 
 #include "portaudio.h"
 #include "pa_win_ds.h"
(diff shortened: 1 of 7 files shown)


test/patest_dsound_find_best_latency_params.c
```diff
@@ -201,7 +202,7 @@ static void printWindowsVersionInfo( FILE *fp )
 
 
     fprintf( fp, "OS name and edition: %s %s\n", osName, osProductType );
-    fprintf( fp, "OS version: %d.%d.%d %S\n",
+    _ftprintf( fp, TEXT("OS version: %d.%d.%d %s\n"),
```
philburk, line 205: Why "_fprintf"?

philburk (approved):
I notice the same fixes are in #780.
If we merge this first then we should rebase that PR.

test/patest_dsound_find_best_latency_params.c
dechamps: Because `osProductType` comes from `GetVersionEx()` (as opposed to `GetVersionExA()` or `GetVersionExW()`), the string type is a [`TCHAR`](https://learn.microsoft.com/en-us/cpp/text/generic-text-mappings-in-tchar-h?view=msvc-170). The correct way to print a TCHAR is [`_tprintf`](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/printf-printf-l-wprintf-wprintf-l?view=msvc-170#generic-text-routine-mappings).

The corresponding warning is:

```
portaudio\test\patest_dsound_find_best_latency_params.c(205): warning C4477: 'fprintf' : format string '%S' requires an argument of type 'unsigned short *', but variadic argument 4 has type 'CHAR *'
```

It can be debated whether we should get rid of TCHAR code and switch to Unicode Win32 APIs everywhere, but I doubt we want this debate to be in scope for this PR, so for now I'm just making sure the code is at least consistent.

dechamps:
> I notice the same fixes are in https://github.com/PortAudio/portaudio/pull/780. If we merge this first then we should rebase that PR.

Yes, [this is the plan](https://github.com/PortAudio/portaudio/pull/780#issuecomment-1480233793).

RossBencina (approved):
Looks good, thanks.

PR #800, opened 2023-03-22
Remove -Werror from autotools darwin build

dechamps:
Warnings should not be turned into errors by default as that can cause difficulties for people who just want to build PortAudio, not work on it. Also this is inconsistent with other platforms.

This PR is a spin-off of #780.

### DIFF
--- a/configure
+++ b/configure
@@ -15866,7 +15866,7 @@ case "${host_os}" in
         $as_echo "#define PA_USE_COREAUDIO 1" >>confdefs.h
 
 
-        CFLAGS="$CFLAGS -I\$(top_srcdir)/src/os/unix -Wno-deprecated -Werror"
+        CFLAGS="$CFLAGS -I\$(top_srcdir)/src/os/unix -Wno-deprecated"
         LIBS="-framework CoreAudio -framework AudioToolbox -framework AudioUnit -framework CoreFoundation -framework CoreServices"
 
         if test "x$enable_mac_universal" = "xyes" ; then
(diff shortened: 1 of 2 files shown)


philburk (approved):

RossBencina:
I checked #636, here are the pertinent comments from there:

r-burns https://github.com/PortAudio/portaudio/pull/636#issuecomment-913055557 wrote:

> Werror is good to enable for development, but unless your CI checks every possible compiler out there, warnings can always sneak by. For example, Nixpkgs ran into some new warnings with gcc 10. I think the recommended approach is to keep Werror off by default, and enable it on the CI using CPPFLAGS. 

@philburk https://github.com/PortAudio/portaudio/pull/636#issuecomment-913249765 wrote:

> I am in favor of merging this after we add -Werror to the CI.

And later in the thread https://github.com/PortAudio/portaudio/pull/636#issuecomment-937346049 :

> Ross and I discussed this and we are both OK with removing -Werror from configure and cmake.
> So we need to have the -Werror in the CI builds.

@RossBencina https://github.com/PortAudio/portaudio/pull/636#issuecomment-915668929 wrote:

> I think this is heading in the right direction. It's OK to limit -Werror to CI.

And later in the thread https://github.com/PortAudio/portaudio/pull/636#discussion_r723763544 :

> The CI builds should use -Werror for both debug and release builds.

RossBencina:
So from #636 we already agreed to removing `-Werror` from our distributed build files, but we want to retain it in CI.

I've created a PR with options for adding it back to autotools CI:
https://github.com/PortAudio/portaudio/pull/803

We should discuss the approach on that PR.

RossBencina (approved):
OK, but let's fix CI too. See #803 for that.

PR #802, opened 2023-03-25
oss: Clean up header handling

brad0:
Clean up some of the OSS header handling. machine/soundcard.h is from FreeBSD releases that are over 20 years old and linux/soundcard.h is used by ancient Linux. Modern FreeBSD / NetBSD and Linux with OSSv4 is good with sys/soundcard.h.

### DIFF
--- a/cmake/modules/FindOSS.cmake
+++ b/cmake/modules/FindOSS.cmake
@@ -10,11 +10,9 @@ Imported Targets
 This module provides the following imported targets, if found:
 
 ``OSS::oss``
-  Target for the OSS header include directory. One of the following
-  compile definitions is added to the target:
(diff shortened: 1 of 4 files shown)


philburk (commented):
Thanks for this. But one of the main goals for PortAudio is portability across
multiple APIs and OS versions.
This PR removes compatibility with old versions of Unix.
So this could break some very old builds.
Does it fix a problem that occurs on more modern systems?

philburk:
@brad0 - Are you still interested in this PR? There are some unanswered questions.

philburk:
Closing for inactivity. We want to keep compatibility with old systems.
If this fixes a problem with current system, please file an Issue.

PR #803, opened 2023-03-28
[DRAFT] example of setting -Werror in autotools ./configure CI script

RossBencina:
Either include in #800 or follow up from 800

I don't intend to use both options. Choose one option, either

1. configure step, OR
2. make step

See discussion at https://stackoverflow.com/questions/32792692/cflags-in-configure-script

### DIFF
--- a/.github/workflows/autotools.yml
+++ b/.github/workflows/autotools.yml
@@ -15,6 +15,6 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: configure
-      run: ./configure
+      run: ./configure CFLAGS=-Werror
     - name: make
-      run: make
+      run: make CFLAGS=-Werror


RossBencina:
Right now I don't have an environment to test this. Seems to be failing CI. @dechamps if feasible could you check whether either of these options work for re-enabling `-Werror`?

dechamps:
> @dechamps if feasible could you check whether either of these options work for re-enabling `-Werror`?

It works if you remove the change to `make` and only pass `CFLAGS=-Werror` to `configure`. I'm not too familiar with autoconf, but I suspect the reason is because if you do `make CFLAGS=…` then you are overriding `CFLAGS` that are generated by `configure`, including critical options like include paths (hence the `portaudio.h: No such file or directory` error).

Of course, once you fix that, you will then also need to reproduce this allowlist for the build to succeed: https://github.com/PortAudio/portaudio/blob/65df2c4943d4ebd513e2f5d25724d725bb7ded4c/CMakeLists.txt#L36-L39

Which then begs the question of whether we really want to have this list be in two places at the same time and having to keep both in sync. (This is where I would typically add "this is why it's a bad idea to try to maintain multiple build systems" but we both know how this discussion will end, so let's not go there.)

RossBencina:
Thanks for checking. I've reverted to setting environment variables rather than passing parameters to `./configure`. In both cases the CI passed. Strange.

> Which then begs the question of whether we really want to have this list be in two places at the same time and having to keep both in sync.

To be honest, I'd prefer to not have such a list at all. `-Wno-error=deprecated-declarations` is probably unavoidable.

PR #812, opened 2023-04-15
Remove some warnings from allowlist

dechamps:
These warnings are not currently being triggered. Their presence is probably the result of merge confusion between #780 and #799.


RossBencina (approved):

RossBencina:
Awesome, thanks.

Issue #810, opened 2023-04-10
Fix compilation warnings


RossBencina:
Ideally we would like to not have any compiler warnings. Right now the CI will suppress them in order for the build to pass. See the list in this and later commits: #780

PR #821, opened 2023-06-04
Fix MSVC warning C4018 signed/unsigned mismatch

dechamps:
See #810


philburk (approved):
Thanks for this cleanup.

RossBencina (approved):
Upon second review I think using unsigned long in pa_jack.c is good.

Issue #828, opened 2023-07-30
Don't link to macOS frameworks with an absolute path


daschuer:
**Describe the bug**
Portaudio links to macOS framework with an absolute path. This breaks dynamic linked builds by updating Xcode. 

**To Reproduce**
Build Portaudio on macOS and look at the PortAudioTargets.cmake file: 
```
set_target_properties(PortAudio::PortAudio PROPERTIES
  INTERFACE_COMPILE_DEFINITIONS "PA_USE_JACK=1;PA_USE_COREAUDIO=1"
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include"
  INTERFACE_LINK_LIBRARIES "\$<LINK_ONLY:Threads::Threads>;\$<LINK_ONLY:JACK::jack>;\$<LINK_ONLY:m>;/Applications/Xcode_12.4.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.1.sdk/System/Library/Frameworks/CoreAudio.framework;/Applications/Xcode_12.4.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.1.sdk/System/Library/Frameworks/AudioToolbox.framework;/Applications/Xcode_12.4.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.1.sdk/System/Library/Frameworks/AudioUnit.framework;/Applications/Xcode_12.4.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.1.sdk/System/Library/Frameworks/CoreFoundation.framework;/Applications/Xcode_12.4.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.1.sdk/System/Library/Frameworks/CoreServices.framework"
)
```
**Expected behavior**
instead macOS framework shall be linked with the -framework option 

**Actual behavior**
Building is broken after updating Xcode. A Rebuild of portaudio is required. 

**Desktop (please complete the following information):**
 - OS: macOS 
 - OS: Tested with 12.4
 - PortAudio version: 19.7 

**Additional context**
Here is a GitHub workflow that demonstrates the issue: https://github.com/daschuer/mixxx/actions/runs/5688916805
Here is a fixed workflow with a patched Portaudio: https://github.com/daschuer/mixxx/actions/runs/5702341115
[...]

PR #829, opened 2023-08-01
Don't use apsolute path when linking to macOS frameworks

daschuer:
This fixes https://github.com/PortAudio/portaudio/issues/828

From CMake 3.24 we can use a special generator expression for frameworks. 
But since PortAudio is on 3.1.0 I did changed it to the -Wl syntax explicit.


RossBencina (approved):

philburk (approved):

philburk:
@daschuer - Thanks for finding and fixing this serious bug.

PR #830, opened 2023-08-05
win: New PaWinUtil_GetOsVersion() function for getting Windows OS version

dmitrykos:
As per discussion in #796 introducing a single API for getting Windows OS version in PA that will improve client code readability and reliability.

Additionally: 
- refactored WASAPI, DS, MME and WDMKS host back-ends to use `PaWinUtil_GetOsVersion()` instead of direct OS API


src/os/win/pa_win_version.c
```diff
@@ -0,0 +1,240 @@
...
+
+    // RTL_OSVERSIONINFOW equals OSVERSIONINFOW but it is missing inb MinGW winnt.h header,
+    // thus use OSVERSIONINFOW for greater portability
+    typedef NTSTATUS (WINAPI *LPFN_RTLGETVERSION)(POSVERSIONINFOW lpVersionInformation);
+    LPFN_RTLGETVERSION fnRtlGetVersion;
+
+    #define NTSTATUS_SUCCESS ((NTSTATUS)0x00000000L)
+
+    // RtlGetVersion must be able to provide true Windows version (Windows 10 may be reported as Windows 8
+    // by GetVersion API)
```
MichalPetryka, line 113: `RtlGetVersion` does in fact also lie about the version in some cases, for example when Windows decides to use compatibility shims.

dmitrykos:
@MichalPetryka thank you for your comment, I amended inline comment  accordingly.

src/os/win/pa_win_version.c
dmitrykos: @MichalPetryka thank you for your comment, I amended inline comment accordingly.

src/os/win/pa_win_version.h
```diff
@@ -0,0 +1,80 @@
...
+    paOsVersionWindowsNT4,
+    paOsVersionWindows2000,
+    paOsVersionWindowsXP,
+    paOsVersionWindowsXPServer2003,
+    paOsVersionWindowsVistaServer2008,
+    paOsVersionWindows7Server2008R2,
+    paOsVersionWindows8Server2012,
+    paOsVersionWindows8_1Server2012R2,
+    paOsVersionWindows10Server2016,
+    paOsVersionWindowsFuture
```
philburk, line 66: If we add new versions later on then paOsVersionWindowsFuture will change.
Maybe we should reserve some space:

    paOsVersionWindowsFuture = 1000,

philburk (changes requested):
Nice. Thanks for doing this.

src/os/win/pa_win_version.h
dmitrykos: My assumption is that `PaWinUtil_GetOsVersion()` is library-private therefore `paOsVersionWindowsFuture` does not need specific hardcoded border value. But, why not doing it like you proposed, it potentially may simplify debugging when OS version is reported/logged with some specific easily noticeable large value.

MichalPetryka: Does it make sense to have a `future` value at all? Nothing should ever be checking for this AFAIR.

dmitrykos: It makes sense to me as you can easily extend implementation to the next known future version. Also, client code has possibility to make assertion if knowledge about future version is desired for the implementation.

```c
assert(PaWinUtil_GetOsVersion() < paOsVersionWindowsFuture);
```

philburk (commented):
LGTM, I'll wait for Ross to review.

philburk (approved):

dmitrykos:
Overlooked that WDMKS is also using `GetVersion`, just added it into the PR.

src/hostapi/wasapi/pa_win_wasapi.c
```diff
@@ -2458,7 +2290,7 @@ PaError PaWasapi_Initialize( PaUtilHostApiRepresentation **hostApi, PaHostApiInd
 
 #ifndef PA_WINRT
     // Fail safely for any Windows version below Windows Vista
```
RossBencina, line 2292: This comment now seems incorrect, since paOsVersionWindowsUnknown no longer means "version below Vista"

Not sure whether it's the code or the comment that needs to change.

RossBencina (changes requested):
Hi Dmitry,

In addition to my one review comment, could you please let us know what tests you have run on this new code. Thankyou!

src/hostapi/wasapi/pa_win_wasapi.c
dmitrykos: Hi Ross! Thank you for noticing this, made correction to detect OS below Vista.

dmitrykos:
> In addition to my one review comment, could you please let us know what tests you have run on this new code.

I tested on Windows 10:
```c
PortAudio Test: output sine wave. SR = 44100, BufSize = 64
before paHostApiInitializers[0].
getting Windows version with RtlGetVersion(): major=10, minor=0, build=19045, platform=2
Windows version=10
```
also Windows UWP and the lowest I have is Windows Server 2003 R2. I did not test on Windows Vista, 7, 8.

Windows Server 2003 R2 is practically latest Windows XP with 3790 build (https://en.wikipedia.org/wiki/List_of_Microsoft_Windows_versions) which is detected correctly:
```c
PortAudio Test: output sine wave. SR = 44100, BufSize = 64
before paHostApiInitializers[0].
getting Windows version with RtlGetVersion(): major=5, minor=2, build=3790, platform=2
Windows version=5
```

Unfortunately I did not implement any automated testing every Windows version that could be done with a help of Docker due to lack of time resource. The underlying implementation is copy-paste from the WASAPI's implementation which was tested with time on Windows Vista, 7, 8.

RossBencina (approved):

PR #834, opened 2023-09-06
alsa: do not link with lasound in the PA_ALSA_DYNAMIC case

jcelerier:
it kinda defeats the point of the option (making binaries that will still run on systems without libasound.so.2)


RossBencina:
LGTM but I'm not competent to sign off on a CMake change. @Be-ing ?

If anyone with CMake chops could comment that would be a great help.

dmitrykos:
> If anyone with CMake chops could comment that would be a great help.

@RossBencina the change looks correct and can be safely merged. When `PA_ALSA_DYNAMIC` is `ON` the `libasound` should not be linked to the binary and @jcelerier is right about it. In the past in 9c4c92fef6f0ef65682062870c87a29eedd79f9e I added dynamic loading possibility for Linux driver just for this reason.

dmitrykos (approved):

RossBencina (approved):

philburk (approved):

jcelerier:
thanks!

PR #841, opened 2023-09-21
Implemented Int32_To_UInt8_Dither

ngoiyaeric:
Implemented a method to convert data from int32_To_UnInt8_Dither. Port audio did not offer 32bitTriangularDither so I used the available 16bitTriangularDither. Would like to know how to test this code and how to check if it is implemented correctly. The build for port audio works with these changes.

[work towards issue #35]


RossBencina:
Thanks for this. I'm not sure that you have the dither scaling factor correct, could you provide a rationale please?

Running the following tests should be sufficient:

- https://github.com/PortAudio/portaudio/blob/master/test/patest_dither.c
- https://github.com/PortAudio/portaudio/blob/master/test/patest_converters.c

ngoiyaeric:
Hi, thanks for the corrections. I made the code changes you have suggested. What dynamic range code should I expect after running test/patest_converters for the i32 to ui8 conversions, what logic should I use to generalize this indicator across various conversion tests.

philburk (changes requested):
The new math looks good.
But the CI found some trailing whitespace:

error: src/common/pa_converters.c(1016) trailing whitespace:
b'    unsigned char *dest =  (unsigned char*)destinationBuffer; '
error: src/common/pa_converters.c(1021) trailing whitespace:
b'       '
error: src/common/pa_converters.c(1028) trailing whitespace:
b'        dest += destinationStride; '

RossBencina:
> What dynamic range code should I expect after running test/patest_converters for the i32 to ui8 conversions

Same as other ui8 conversions I guess.

Could you paste the output of this test here please and we'll scratch our heads about it:

https://github.com/PortAudio/portaudio/blob/master/test/patest_converters.c

ngoiyaeric:
> > What dynamic range code should I expect after running test/patest_converters for the i32 to ui8 conversions
> 
> Same as other ui8 conversions I guess.
> 
> Could you paste the output of this test here please and we'll scratch our heads about it:
> 
> https://github.com/PortAudio/portaudio/blob/master/test/patest_converters.c

Hi, 
When I run this command 
gcc -l/portaudio/src/common/ patest_converters.c -o your_output_executable -lm -lportaudio
to test the module, I get this error. 
patest_converters.c:53:10: fatal error: 'pa_converters.h' file not found
#include "pa_converters.h"
         ^~~~~~~~~~~~~~~~~
1 error generated.

Any Ideas how I could fix it?

dmitrykos:
@ngoiyaeric it depends on how `src/common` folder is related to the path from which you are running GCC build command. If you are running GCC build from inside `portaudio` folder then it would be `-Isrc/common`, if from outside but `portaudio` folder is not in the root of the storage then `-Iportaudio/src/common` and so on.

ngoiyaeric:
> @ngoiyaeric it depends on how `src/common` folder is related to the path from which you are running GCC build command. If you are running GCC build from inside `portaudio` folder then it would be `-Isrc/common`, if from outside but `portaudio` folder is not in the root of the storage then `-Iportaudio/src/common` and so on.

hi, this does not remedy the issue, the same error is still thrown.

dmitrykos:
@ngoiyaeric in any case GCC is struggling to locate `pa_converters.h` in your build configuration, it is the only reason. To avoid dealing with compiler configuration directly you could try building `patest_converters` via CMake generated build project.

dmitrykos:
@ngoiyaeric have you managed to test your implementation? Also this commit (https://github.com/PortAudio/portaudio/pull/841/commits/273c99101954004ff01f93d5f50423b379c21bed) has to be removed from PR.

RossBencina:
We'd like to see this tested and finalised but given the lack of input, we're not going to hold up the next milestone release for this. We can revisit if someone steps up to fix it.

.github/workflows/codesee-arch-diagram.yml
```diff
@@ -0,0 +1,23 @@
+# This workflow was added by CodeSee. Learn more at https://codesee.io/
```
philburk, line 1: This file should be removed from this PR.
It is unrelated to the purpose of the PR.

philburk (changes requested):

src/common/pa_converters.c
```diff
@@ -1012,16 +1012,20 @@ static void Int32_To_UInt8_Dither(
     void *sourceBuffer, signed int sourceStride,
     unsigned int count, struct PaUtilTriangularDitherGenerator *ditherGenerator )
 {
-    /* PaInt32 *src = (PaInt32*)sourceBuffer;
-    unsigned char *dest =  (unsigned char*)destinationBuffer; */
+    PaInt32 *src = (PaInt32*)sourceBuffer;
+    unsigned char *dest =  (unsigned char*)destinationBuffer; 
     (void)ditherGenerator; /* unused parameter */
```
philburk, line 1017: Remove this line. Used on line: 1022.

RossBencina:
Closing as incomplete / abandoned. We can't merge this as-is and there's been no response from the contributor for over one year.

PR #843, opened 2023-09-30
pulseaudio: Move Pulseaudio include in correct place when using autoconf

illuusio:
As Jack and Pulseaudio both needs Ringbuffer that include can be done in same place. In configure.in also Pulseaudio header file was included before it was sure that it was really needed.

Commit makes sure that Pulseaudio include is available only if it's needed as it can cause failing in build if Pulseaudio develoment files are not available.


philburk (approved):
Thanks. Ross and I were curious about this.

Issue #201, opened 2011-12-08
Write a test that prints out the default latency values for each device


PortAudio-admin:
*Issue created by @RossBencina*

Write a test that prints out host api and default latency values for each device. The output of this test need to be reviewed for each host API.

For each device print name, host API and the four default values:
			
    device index, device name, host api name, default high input, default low input, default high output, default low output

This is needed as part of the process of resolving ticket #202, #97

Vimal-Krishna:
Hi. I hope I can take this up and complete it. I have something working and will share an update soon.

Vimal-Krishna:
Created a PR with screenshot of the outputs - https://github.com/PortAudio/portaudio/pull/844

PR #844, opened 2023-10-02
Adding test to resolve Issue 201

Vimal-Krishna:
Adding a new test - patest_enumerate_default_latency

Issue - https://github.com/PortAudio/portaudio/issues/201

Outputs in this format to the cmd line:
```
| X |  42 |  == 1 - LG HDR 4K (AMD High Definition Audio Device) [Loopback] == | Windows WASAPI |   0.0100 |   0.0030 |   0.0000 |   0.0000 |

|   |  68 |  == Headset (@System32\drivers\bthhfenum.sys,#2;%1 Hands-Free AG Audio%0;(HC 2000BNC)) == | Windows WDM-KS |   0.0853 |   0.0100 |   0.0853 |   0.0100 |
```

Included the '==' markdown to disable textile formatting within the name field so that special characters like @ do not cause a formatting issue.

image

Tested the output on https://textile-lang.com/ 

image


RossBencina (commented):
Thankyou! I've added some feedback for your consideration.

test/patest_enumerate_default_latency.c
```diff
@@ -0,0 +1,139 @@
...
+#ifdef WIN32
+        {   /* Use wide char on windows, so we can show UTF-8 encoded device names */
+            wchar_t wideName[MAX_PATH];
+            MultiByteToWideChar(CP_UTF8, 0, deviceInfo->name, -1, wideName, MAX_PATH - 1);
+            printf(" == ");
+            print_clean_wstring(wideName);
+            printf(" == | ");            
+        }
+#else
+        printf(" == ");
```
philburk, line 119:
```suggestion
        printf(" == %s == | ", deviceInfo->name);
```

philburk (changes requested):
Thanks for adding this new test program. Looks great!
Just some minor suggestions.

RossBencina:
Please fix the compilation errors in CI too.

Vimal-Krishna:
Thanks for the feedback folks! I think the CI issues should be fixed as well. Let me know if the headers are ok.
(snipped to show entries with 'X')

image

RossBencina:
Thanks this is looking good. The only thing I'm thinking is that it's a bit hard to interpret the 0.0000 entries because we can't see whether it's an input-only or output-only device where 0 is correct, or whether it's a bad value.

What do you think about adding another column `Input/Output Channels` with the channel counts for example `2/0` for a stereo input-only device.

Vimal-Krishna:
> Thanks this is looking good. The only thing I'm thinking is that it's a bit hard to interpret the 0.0000 entries because we can't see whether it's an input-only or output-only device where 0 is correct, or whether it's a bad value.
> 
> What do you think about adding another column `Input/Output Channels` with the channel counts for example `2/0` for a stereo input-only device.

Yes, seems like a useful addition. I can make the change this weekend. How about we put in a value like N/A instead of 0.0000 for the output latency of an input-only device? And also exclude it when deciding if a device needs the 'X' marker. So the 'X' s will still be the only thing to look for when looking at the report.

Vimal-Krishna:
Devices 12 to 23 had an X marker before. Now they don't since the channel count is zero for either input or output.
textile-lang com_

test/patest_enumerate_default_latency.c
```diff
@@ -0,0 +1,164 @@
+/** @file patest_enumerate_default_latency.c
+    @ingroup test_src
+    @brief List available devices in table format
```
RossBencina, line 3:
```suggestion
    @brief List default latencies of available devices in table format
```

test/patest_enumerate_default_latency.c
```diff
@@ -0,0 +1,164 @@
...
+                isLatencyZero = 1;
+            }
+        }
+
+        if (deviceInfo->maxOutputChannels > 0) {
+            if (defaultHighOutputLatency == 0 || defaultLowOutputLatency == 0) {
+                isLatencyZero = 1;
+            }
+        }
+
```
RossBencina, line 133:
```suggestion

```

test/patest_enumerate_default_latency.c
```diff
@@ -0,0 +1,164 @@
...
+    numDevices = Pa_GetDeviceCount();
+    if( numDevices < 0 )
+    {
+        printf( "ERROR: Pa_GetDeviceCount returned 0x%x\n", numDevices );
+        err = numDevices;
+        goto error;
+    }
+
+    // Header for the table
+    printf("|_. Default Latency Zero? |_. Device Number |_. I/O Channels |_. Device Name |_. Host API |_. Default High Input Latency |_. Default Low Input Latency |_. Default High Output Latency |_. Default Low Output latency |\n");
```
RossBencina, line 110:
```suggestion
    printf("|_. Bad Default Latency? |_. Device Number |_. I/O Channels |_. Device Name |_. Host API |_. Default High Input Latency |_. Default Low Input Latency |_. Default High Output Latency |_. Default Low Output latency |\n");
```

RossBencina (commented):
LGTM just going to merge these tweaks myself.

RossBencina (approved):

philburk (approved):

philburk:
@Vimal-Krishna - Thanks for this contribution. The code looked very nice. We appreciate you closing this old issue from 2011.

Vimal-Krishna:
@RossBencina @philburk 
Thank you! I enjoyed working on it. I hope to continue working on the project. 🙂

PR #846, opened 2023-10-02
wdmks: declare GUIDs with selectany attribute

invertego:
Match the behavior of guiddef.h in both mingw and the Windows SDK headers. This prevents linking errors caused by multiply defined symbols when linking against certain Windows SDK libs (like dxguid.lib).

### DIFF
--- a/src/hostapi/wdmks/pa_win_wdmks.c
+++ b/src/hostapi/wdmks/pa_win_wdmks.c
@@ -164,12 +164,17 @@ Default is to use the pin category.
 #define DYNAMIC_GUID(data) {data}
 #define _NTRTL_ /* Turn off default definition of DEFINE_GUIDEX */
 #undef DEFINE_GUID
+#ifdef DECLSPEC_SELECTANY
+#define PA_DECLSPEC_SELECTANY DECLSPEC_SELECTANY
+#else
+#define PA_DECLSPEC_SELECTANY
+#endif
 #if defined(__clang__) || (defined(_MSVC_TRADITIONAL) && !_MSVC_TRADITIONAL) /* clang-cl and new msvc preprocessor: avoid too many arguments error */
-    #define DEFINE_GUID(n, ...) EXTERN_C const GUID DECLSPEC_SELECTANY n = {__VA_ARGS__}
+    #define DEFINE_GUID(n, ...) EXTERN_C const GUID PA_DECLSPEC_SELECTANY n = {__VA_ARGS__}
     #define DEFINE_GUID_THUNK(n, ...) DEFINE_GUID(n, __VA_ARGS__)
     #define DEFINE_GUIDEX(n) DEFINE_GUID_THUNK(n, STATIC_##n)
 #else
-    #define DEFINE_GUID(n, data) EXTERN_C const GUID DECLSPEC_SELECTANY n = {data}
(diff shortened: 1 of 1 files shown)


invertego:
I encountered this issue when building MAME, which currently [recommends disabling portaudio](https://github.com/mamedev/mame/commit/8d4bbaddd435e991882287e7bc1259dc53c507d5) when building with VS.

I'm honestly not sure this whole section of code - which seems to be just duplicating portions of guiddef.h - even ought to exist in portaudio. It appears to go back at least 16 years in age.  That said, not being familiar with your build requirements, I thought I'd offer a minimally disruptive change.

mingw reference:
https://github.com/mingw-w64/mingw-w64/blob/master/mingw-w64-headers/include/guiddef.h

src/hostapi/wdmks/pa_win_wdmks.c
```diff
@@ -165,11 +165,11 @@ Default is to use the pin category.
 #define _NTRTL_ /* Turn off default definition of DEFINE_GUIDEX */
 #undef DEFINE_GUID
 #if defined(__clang__) || (defined(_MSVC_TRADITIONAL) && !_MSVC_TRADITIONAL) /* clang-cl and new msvc preprocessor: avoid too many arguments error */
```
RossBencina, line 167:
```suggestion
#ifdef DECLSPEC_SELECTANY
#define PA_DECLSPEC_SELECTANY DECLSPEC_SELECTANY
#else
#define PA_DECLSPEC_SELECTANY
#endif
#if defined(__clang__) || (defined(_MSVC_TRADITIONAL) && !_MSVC_TRADITIONAL) /* clang-cl and new msvc preprocessor: avoid too many arguments error */
```

src/hostapi/wdmks/pa_win_wdmks.c
```diff
@@ -165,11 +165,11 @@ Default is to use the pin category.
...
-    #define DEFINE_GUID(n, ...) EXTERN_C const GUID n = {__VA_ARGS__}
+    #define DEFINE_GUID(n, ...) EXTERN_C const GUID DECLSPEC_SELECTANY n = {__VA_ARGS__}
```
RossBencina, line 168:
```suggestion
    #define DEFINE_GUID(n, ...) EXTERN_C const GUID PA_DECLSPEC_SELECTANY n = {__VA_ARGS__}
```

src/hostapi/wdmks/pa_win_wdmks.c
```diff
@@ -165,11 +165,11 @@ Default is to use the pin category.
...
     #define DEFINE_GUID_THUNK(n, ...) DEFINE_GUID(n, __VA_ARGS__)
     #define DEFINE_GUIDEX(n) DEFINE_GUID_THUNK(n, STATIC_##n)
 #else
-    #define DEFINE_GUID(n, data) EXTERN_C const GUID n = {data}
+    #define DEFINE_GUID(n, data) EXTERN_C const GUID DECLSPEC_SELECTANY n = {data}
```
RossBencina, line 172:
```suggestion
    #define DEFINE_GUID(n, data) EXTERN_C const GUID PA_DECLSPEC_SELECTANY n = {data}
```

philburk (approved):
I approve this with Ross's changes.

PR #850, opened 2023-10-10
wasapi: Fix wrong sign in timeInfo.inputBufferAdcTime calculation.

kleinerm:
ADC capture time must be earlier than currentTime, as the capture time of any received samples in a buffer must be in the past.

Fix the wrong sign in calculation to fix this.


philburk (approved):

RossBencina (approved):

RossBencina:
@kleinerm thanks!

@dmitrykos Phil and I think this looks good, if you're ok with it would you like to go ahead and merge it please.

dmitrykos:
@RossBencina, sorry noticed this thread only now and the fix looks correct. Merged.

dmitrykos (commented):
Looks correct to me.

PR #852, opened 2023-10-14
pulseaudio: fix PaPulseAudio_Initialize unlocking

### LINKED ISSUE #848: PaPulseAudio_Initialize does not unlock correctly on error path
This PR fixes an access-after free when calling `PaPulseAudio_UnLock`

https://github.com/PortAudio/portaudio/pull/847

However the bigger issue is that `PaPulseAudio_UnLock` is not called consistently. Either `PaPulseAudio_UnLock` needs to be called before or after `goto error` but not both or neither. On line 563 it gets locked, but there is also a `goto error` prior to that on line 550. Some later `goto error` paths unlock, some don't.

One possible fix is to have an `error_unlock:` label prior to the `error` label that is used for paths that should unlock.

illuusio:
When PaPulseAudio_Initialize fails it does not currently do it very well as bug report  #848 tells. Now Only free lock if it's initialized enough to be unlocked.

Unlocking is done after allocations are done. That is because Locking is just after allocations and next `goto error` should have allocated memory and drop to that. Also now Group freeing code is one place.

Resolves #848

### DIFF
--- a/src/hostapi/pulseaudio/pa_linux_pulseaudio.c
+++ b/src/hostapi/pulseaudio/pa_linux_pulseaudio.c
@@ -211,11 +211,17 @@ void PaPulseAudio_Free( PaPulseAudio_HostApiRepresentation * ptr )
         ptr->timeEvent = NULL;
     }
 
-
     if( ptr->mainloop )
     {
         pa_threaded_mainloop_free( ptr->mainloop );
         ptr->mainloop = NULL;
+    }
+
+    if( ptr->allocations )
+    {
+        PaUtil_FreeAllAllocations( ptr->allocations );
+        PaUtil_DestroyAllocationGroup( ptr->allocations );
+        ptr->allocations = NULL;
     }
 
     PaUtil_FreeMemory( ptr );
@@ -529,6 +535,7 @@ PaError PaPulseAudio_Initialize( PaUtilHostApiRepresentation ** hostApi,
     int i;
     int deviceCount;
     int ret = 0;
+    int lockTaken = 0;
     PaPulseAudio_HostApiRepresentation *pulseaudioHostApi = NULL;
     PaDeviceInfo *deviceInfoArray = NULL;
 
@@ -561,6 +568,8 @@ PaError PaPulseAudio_Initialize( PaUtilHostApiRepresentation ** hostApi,
 
     /* Connect to server */
     PaPulseAudio_Lock( pulseaudioHostApi->mainloop );
+    lockTaken = 1;
+
     ret = pa_context_connect( pulseaudioHostApi->context,
                                  NULL,
                                  0,
@@ -570,8 +579,7 @@ PaError PaPulseAudio_Initialize( PaUtilHostApiRepresentation ** hostApi,
     {
         PA_PULSEAUDIO_SET_LAST_HOST_ERROR( 0,
                                            "PulseAudio_Initialize: Can't connect to server");
-        result = paNotInitialized;
-        PaPulseAudio_UnLock( pulseaudioHostApi->mainloop );
+        result = paUnanticipatedHostError;
         goto error;
     }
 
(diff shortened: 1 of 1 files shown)


src/hostapi/pulseaudio/pa_linux_pulseaudio.c
```diff
@@ -750,10 +755,8 @@ PaError PaPulseAudio_Initialize( PaUtilHostApiRepresentation ** hostApi,
     {
         if( pulseaudioHostApi->allocations )
```
philburk, line 756:
```suggestion
        if( lockTaken )
```

philburk (changes requested):

RossBencina (commented):
Just a suggestion...

src/hostapi/pulseaudio/pa_linux_pulseaudio.c
RossBencina: I like this. It is clearer.

add
```
    int lockTaken = 0;
```

to top of function, and

```
    lockTaken = 1;
```

after the call to `PaPulseAudio_Lock`

src/hostapi/pulseaudio/pa_linux_pulseaudio.c
```diff
@@ -211,13 +211,18 @@ void PaPulseAudio_Free( PaPulseAudio_HostApiRepresentation * ptr )
...
     {
         pa_threaded_mainloop_free( ptr->mainloop );
         ptr->mainloop = NULL;
     }
 
+    if( ptr->allocations )
+    {
+        PaUtil_FreeAllAllocations( ptr->allocations );
+        PaUtil_DestroyAllocationGroup( ptr->allocations );
+    }
```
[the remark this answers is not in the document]
illuusio: Ok i've added this one too

src/hostapi/pulseaudio/pa_linux_pulseaudio.c
illuusio: Ok this should be also resolved

kleinerm:
@illuusio Changes look good to me. I also gave it a quick test and it worked.

src/hostapi/pulseaudio/pa_linux_pulseaudio.c
```diff
@@ -589,6 +597,7 @@ PaError PaPulseAudio_Initialize( PaUtilHostApiRepresentation ** hostApi,
                 break;
             case PA_CONTEXT_TERMINATED:
             case PA_CONTEXT_FAILED:
+                result = paUnanticipatedHostError;
```
RossBencina, line 600: I think there's a missing call to `PaUtil_SetLastHostErrorInfo` here. perhaps separate calls for each case.

RossBencina: It's fine to do additional cleanup later, but this PR adds a `paUnanticipatedHostError`. Our policy is to not accept PRs that introduce new problems, so we'd appreciate it if you could fix this case in this PR please.

illuusio: Ok now it should be solved

RossBencina: Thanks, but now the first parameter to `PA_PULSEAUDIO_SET_LAST_HOST_ERROR` is not correct. It should be the native error code definitely not a PortAudio error code. For this PR, you could just set it to 0 if you don't have anything better. For further info check the definition of `PA_PULSEAUDIO_SET_LAST_HOST_ERROR` in `pa_linux_pulseaudio_internal.h` and then the definition of `PaUtil_SetLastHostErrorInfo` in `pa_util.h`.

Also, `initiliaze` is mis-spelled on line 606.

While I'm here, note that you'll need to fix the `PA_PULSEAUDIO_SET_LAST_HOST_ERROR` in a later PR to pass the correct host api type.

illuusio: I've correct the errors and spelling and actually hostapi type is fixed in another PR but I can fix it here if it's needed.

illuusio: I've removed those new error codes and now it behaves as it was before only it goes to error (problem is that it just return `PaNoError` but as it did worked like that lately it should be ok.

RossBencina (approved):

philburk (approved):

PR #854, opened 2023-10-15
Fix hifiberry dacandadcpro

hoene:
This is the patch proposed in #431


illuusio:
Will this affect on other ALSA drivers? Why does not normal samplerate setup won't work?

RossBencina:
This patch was already proposed (and rejected) in #432. Please don't repost rejected patches.

See the discussion at #431 and #432 for direction on acceptable fixes.

PR #859, opened 2023-10-30
workflows: Add Pulseaudio to Github workflow build

illuusio:
As Pulseaudio is not currently build on Github workflow build. PR Adds it at least for CMake build

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -82,7 +82,7 @@ jobs:
     - name: "[Ubuntu] install dependencies"
       run: |
         sudo apt-get update
-        sudo apt-get install libasound2-dev ${{ matrix.dependencies_extras }}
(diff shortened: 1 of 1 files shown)


kleinerm:
Looks good to me, and I just tested it by pulling into my repo and GitHub CI with the Cmake Ubuntu gcc build seems to build proper libportaudio.a with Pulseaudio backend etc.

One thing I noticed is that some of the install steps does not seem to copy the pulseaudio hostapi specific header file into place, while it does so for JACK and ALSA. Ie. pa_jack.h, pa_linux_alsa.h, pa_unix_oss.h gets installed into /usr/include/ but not pa_linux_pulseaudio.h. I guess that would need some refinement for the `make install` step.

philburk (approved):

RossBencina (approved):

Issue #856, opened 2023-10-20
PaStreamParameters suggestedLatency < 0.0 behavior is not specified and not handled correctly


RossBencina:
The documentation for `PaStreamParameters suggestedLatency` does not mention negative latencies.

I have not reviewed all host APIs but at least `pa_win_wmme.c` and `pa_win_ds.c` assume that `suggestedLatency` is non-negative, and directly assign to an unsigned long, for example in `pa_win_ds.c`:

```
unsigned long suggestedInputLatencyFrames, suggestedOutputLatencyFrames;
...
suggestedInputLatencyFrames = (unsigned long)(inputParameters->suggestedLatency * sampleRate);
...
suggestedOutputLatencyFrames = (unsigned long)(outputParameters->suggestedLatency * sampleRate);
```

There is equivalent code in `pa_win_wmme.c`.

The fix should be to clamp to 0, e.g.

```
if (inputParameters->suggestedLatency >= 0.0)
   suggestedInputLatencyFrames = (unsigned long)(inputParameters->suggestedLatency * sampleRate);
else
    suggestedInputLatencyFrames = 0;
```

There is already a PR to fix this in PA/PulseAudio #849
See also #855

PR #863, opened 2023-11-17
pulseaudio: Rework some initialization code and have better error reporting

illuusio:
Rework Pulseaudio stream starting and initialization code little bit more readable form. Commit also reworks stream starting to work more reliable.

PR reworks also error reporting as it should be first class citizen if there is better stream handling and initialization.

This should overcome problems from PR #857

### DIFF
--- a/src/hostapi/pulseaudio/pa_linux_pulseaudio.c
+++ b/src/hostapi/pulseaudio/pa_linux_pulseaudio.c
@@ -67,29 +67,19 @@ extern char *__progname;
 /* PulseAudio specific functions */
 int PaPulseAudio_CheckConnection( PaPulseAudio_HostApiRepresentation * ptr )
 {
-    pa_context_state_t state;
+    int retCode = paNotInitialized;
 
+    if ( ptr == NULL )
+    {
+        return retCode;
+    }
 
-    /* Sanity check if ptr if NULL don't go anywhere or
-     * it will SIGSEGV
+    pa_context_state_t state = pa_context_get_state( ptr->context );
+
+    /* Check if contenxt if good.
+     * If it is then check state
      */
-    if( !ptr )
-    {
-        PA_PULSEAUDIO_SET_LAST_HOST_ERROR( 0,
-                                           "PaPulseAudio_CheckConnection: Host API is NULL! Can't do anything about it" );
-        return -1;
-    }
-
-    if( !ptr->context || !ptr->mainloop )
-    {
-        PA_PULSEAUDIO_SET_LAST_HOST_ERROR( 0,
-                                           "PaPulseAudio_CheckConnection: PulseAudio context or mainloop are NULL" );
-        return -1;
-    }
-
-    state = pa_context_get_state(ptr->context);
-
-    if( !PA_CONTEXT_IS_GOOD(state) )
+    if( PA_CONTEXT_IS_GOOD(state) )
     {
         switch( state )
         {
@@ -97,20 +87,45 @@ int PaPulseAudio_CheckConnection( PaPulseAudio_HostApiRepresentation * ptr )
              * https://freedesktop.org/software/pulseaudio/doxygen/def_8h.html
              */
 
+            case PA_CONTEXT_READY:
+                retCode = paNoError;
+            break;
+
+            case PA_CONTEXT_CONNECTING:
+            case PA_CONTEXT_AUTHORIZING:
+            case PA_CONTEXT_SETTING_NAME:
+                /* These are just here if they are needed in future */
+            break;
+        }
+    }
+    else
+    {
+        retCode = paUnanticipatedHostError;
+
+        switch( state )
+        {
+            /* These can be found from
+             * https://freedesktop.org/software/pulseaudio/doxygen/def_8h.html
+             */
+
             case PA_CONTEXT_UNCONNECTED:
-               PA_PULSEAUDIO_SET_LAST_HOST_ERROR( 0,
-                                                  "PaPulseAudio_CheckConnection: The context hasn't been connected yet (PA_CONTEXT_UNCONNECTED)" );
+                PA_DEBUG( ("Portaudio %s: Can't connect to server. (PA_CONTEXT_UNCONNECTED)",
+                           __FUNCTION__) );
+            break;
+
+            case PA_CONTEXT_TERMINATED:
+                PA_DEBUG( ("Portaudio %s: Connection terminated to Pulseaudio server. (PA_CONTEXT_TERMINATED)",
+                           __FUNCTION__) );
             break;
 
             case PA_CONTEXT_FAILED:
-               PA_PULSEAUDIO_SET_LAST_HOST_ERROR( 0,
-                                                  "PaPulseAudio_CheckConnection: The connection failed or was disconnected. (PA_CONTEXT_FAILED)" );
+                PA_DEBUG( ("Portaudio %s: Connection failed or was disconnected from Pulseaudio server. (PA_CONTEXT_FAILED)",
+                           __FUNCTION__) );
             break;
         }
-
-        return -1;
     }
-    return 0;
+
+    return retCode;
 }
 
 /* Create HostAPI presensentation */
@@ -447,7 +462,10 @@ void PaPulseAudio_SourceListCb( pa_context * c,
 void PaPulseAudio_StreamStateCb( pa_stream * s,
                                  void *userdata )
 {
+    PaPulseAudio_Stream *stream = (PaPulseAudio_Stream *) userdata;
     const pa_buffer_attr *pulseaudioBufferAttr = NULL;
+    pa_stream_state_t state = PA_STREAM_UNCONNECTED;
+
     /* If you need debug pring enable these
      * char cmt[PA_CHANNEL_MAP_SNPRINT_MAX], sst[PA_SAMPLE_SPEC_SNPRINT_MAX];
      */
@@ -461,11 +479,10 @@ void PaPulseAudio_StreamStateCb( pa_stream * s,
         return;
     }
 
-    switch( pa_stream_get_state(s) )
-    {
-        case PA_STREAM_TERMINATED:
-            break;
+    state = pa_stream_get_state(s);
 
+    switch( state )
+    {
         case PA_STREAM_CREATING:
             break;
 
@@ -475,8 +492,6 @@ void PaPulseAudio_StreamStateCb( pa_stream * s,
                 PA_DEBUG( ("Portaudio %s: Can get buffer attr: '%s'\n",
                            __FUNCTION__,
                            pa_strerror(pa_context_errno(pa_stream_get_context(s) ) )) );
-                PA_PULSEAUDIO_SET_LAST_HOST_ERROR( 0,
-                                           "PaPulseAudio_StreamStateCb: Can't get Stream pa_buffer_attr" );
             }
             else
             {
@@ -487,9 +502,15 @@ void PaPulseAudio_StreamStateCb( pa_stream * s,
             }
             break;
 
+        case PA_STREAM_TERMINATED:
+            PA_DEBUG( ("Portaudio %s: PA_STREAM_TERMINATED '%s'\n",
+                      __FUNCTION__,
+                      pa_strerror( pa_context_errno( pa_stream_get_context( s ) ) )) );
+            break;
+
         case PA_STREAM_FAILED:
         default:
-            PA_DEBUG( ("Portaudio %s: FAILED '%s'\n",
+            PA_DEBUG( ("Portaudio %s: PA_STREAM_FAILED '%s'\n",
                       __FUNCTION__,
                       pa_strerror( pa_context_errno( pa_stream_get_context( s ) ) )) );
 
@@ -511,17 +532,17 @@ void PaPulseAudio_StreamUnderflowCb( pa_stream *s,
     /* If this is null we have big problems and we probably are out of memory */
     if( !s )
     {
-        PA_PULSEAUDIO_SET_LAST_HOST_ERROR( 0,
-                                           "PaPulseAudio_StreamUnderflowCb: Invalid stream" );
+        PA_DEBUG( ("Portaudio %s: Invalid stream",
+                   __FUNCTION__) );
         return;
     }
 
     stream->outputUnderflows++;
     pulseaudioOutputSampleSpec = (pa_buffer_attr *)pa_stream_get_buffer_attr(s);
-    PA_DEBUG( ("Portaudio %s: PulseAudio '%s' with delay: %ld stream has underflowed\n", __FUNCTION__, pa_stream_get_device_name(s), pulseaudioOutputSampleSpec->tlength) );
-
-    PA_PULSEAUDIO_SET_LAST_HOST_ERROR( 0,
-                                       "PaPulseAudio_StreamUnderflowCb: Pulseaudio stream underflow");
+    PA_DEBUG( ("Portaudio %s: PulseAudio '%s' with delay: %ld stream has underflowed\n",
+               __FUNCTION__,
+               pa_stream_get_device_name(s),
+               pulseaudioOutputSampleSpec->tlength) );
 
     pa_threaded_mainloop_signal( stream->mainloop,
                                  0 );
@@ -577,6 +598,8 @@ PaError PaPulseAudio_Initialize( PaUtilHostApiRepresentation ** hostApi,
 
     if( ret < 0 )
     {
+        PA_DEBUG( ("Portaudio %s: Can't connect to server",
+                   __FUNCTION__) );
         PA_PULSEAUDIO_SET_LAST_HOST_ERROR( 0,
                                            "PulseAudio_Initialize: Can't connect to server");
         result = paUnanticipatedHostError;
@@ -590,24 +613,16 @@ PaError PaPulseAudio_Initialize( PaUtilHostApiRepresentation ** hostApi,
     {
         pa_threaded_mainloop_wait( pulseaudioHostApi->mainloop );
 
-        switch( pa_context_get_state( pulseaudioHostApi->context ) )
+        result = PaPulseAudio_CheckConnection( pulseaudioHostApi );
+
+        if( result != paNotInitialized && result != paNoError )
         {
-            case PA_CONTEXT_READY:
-                ret = 1;
-                break;
-            case PA_CONTEXT_TERMINATED:
-                PA_DEBUG( "PaPulseAudio_Initialize: Pulseaudio server terminated stream" );
-                goto error;
-                break;
-            case PA_CONTEXT_FAILED:
-                PA_DEBUG( "PaPulseAudio_Initialize: Pulseaudio server failed to initialize" );
-                goto error;
-                break;
-            case PA_CONTEXT_UNCONNECTED:
-            case PA_CONTEXT_CONNECTING:
-            case PA_CONTEXT_AUTHORIZING:
-            case PA_CONTEXT_SETTING_NAME:
-                break;
+            goto error;
+        }
+
+        if( result == paNoError )
+        {
+            ret = 1;
         }
     }
 
@@ -1224,6 +1239,10 @@ PaError OpenStream( struct PaUtilHostApiRepresentation *hostApi,
             pa_stream_set_started_callback( stream->outputStream,
                                             PaPulseAudio_StreamStartedCb,
                                             stream );
+            pa_stream_set_underflow_callback( stream->outputStream,
+                                              PaPulseAudio_StreamUnderflowCb,
+                                              stream);
+
         }
 
         else
--- a/src/hostapi/pulseaudio/pa_linux_pulseaudio.c
+++ b/src/hostapi/pulseaudio/pa_linux_pulseaudio.c
@@ -600,7 +600,7 @@ PaError PaPulseAudio_Initialize( PaUtilHostApiRepresentation ** hostApi,
     {
         PA_DEBUG( ("Portaudio %s: Can't connect to server",
                    __FUNCTION__) );
-        PA_PULSEAUDIO_SET_LAST_HOST_ERROR( 0,
+        PA_PULSEAUDIO_SET_LAST_HOST_ERROR( ret,
                                            "PulseAudio_Initialize: Can't connect to server");
         result = paUnanticipatedHostError;
         goto error;
--- a/src/hostapi/pulseaudio/pa_linux_pulseaudio.c
+++ b/src/hostapi/pulseaudio/pa_linux_pulseaudio.c
@@ -68,7 +68,7 @@ extern char *__progname;
 int PaPulseAudio_CheckConnection( PaPulseAudio_HostApiRepresentation * ptr )
 {
     /*
-     * This is bit hackish as -1 is consired as error
+     * This is bit hackish as -1 is considered as an error
      * but PA_ERR_* are positive and PA_OK is zero.
      * That is why return -1 for we are waiting something
      * to happen is 'correct'
--- a/src/hostapi/pulseaudio/pa_linux_pulseaudio.c
+++ b/src/hostapi/pulseaudio/pa_linux_pulseaudio.c
@@ -67,7 +67,7 @@ extern char *__progname;
 /* PulseAudio specific functions */
 int PaPulseAudio_CheckConnection( PaPulseAudio_HostApiRepresentation * ptr )
 {
-    int retCode = paNotInitialized;
+    int retCode = -1;
 
     if ( ptr == NULL )
     {
@@ -88,7 +88,7 @@ int PaPulseAudio_CheckConnection( PaPulseAudio_HostApiRepresentation * ptr )
              */
 
             case PA_CONTEXT_READY:
-                retCode = paNoError;
+                retCode = PA_OK;
             break;
 
             case PA_CONTEXT_CONNECTING:
@@ -100,7 +100,7 @@ int PaPulseAudio_CheckConnection( PaPulseAudio_HostApiRepresentation * ptr )
     }
     else
     {
-        retCode = paUnanticipatedHostError;
+        retCode = PA_ERR_ACCESS;
 
         switch( state )
         {
@@ -615,12 +615,12 @@ PaError PaPulseAudio_Initialize( PaUtilHostApiRepresentation ** hostApi,
 
         result = PaPulseAudio_CheckConnection( pulseaudioHostApi );
 
-        if( result != paNotInitialized && result != paNoError )
+        if( result > PA_OK )
         {
             goto error;
         }
 
-        if( result == paNoError )
+        if( result == PA_OK )
         {
             ret = 1;
         }
--- a/src/hostapi/pulseaudio/pa_linux_pulseaudio.c
+++ b/src/hostapi/pulseaudio/pa_linux_pulseaudio.c
@@ -67,6 +67,12 @@ extern char *__progname;
 /* PulseAudio specific functions */
 int PaPulseAudio_CheckConnection( PaPulseAudio_HostApiRepresentation * ptr )
 {
+    /*
+     * This is bit hackish as -1 is consired as error
+     * but PA_ERR_* are positive and PA_OK is zero.
+     * That is why return -1 for we are waiting something
+     * to happen is 'correct'
+     */
     int retCode = -1;
 
     if ( ptr == NULL )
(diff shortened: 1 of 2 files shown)


src/hostapi/pulseaudio/pa_linux_pulseaudio.c
```diff
@@ -931,7 +932,7 @@ PaError PaPulseAudio_BlockingInitRingBuffer( PaUtilRingBuffer * rbuf,
 
     if( ringbufferBuffer == NULL )
     {
-        PA_PULSEAUDIO_SET_LAST_HOST_ERROR( 0,
+        PA_PULSEAUDIO_SET_LAST_HOST_ERROR( paInsufficientMemory,
```
RossBencina, line 935: The only time you should be calling `PA_PULSEAUDIO_SET_LAST_HOST_ERROR` is if you return `paUnanticipatedHostError`. If you can map to a generic PortAudio error, as in this case `paInsufficientMemory`, then you should not call `PA_PULSEAUDIO_SET_LAST_HOST_ERROR`.

illuusio: I've corrected errors as requested

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
```diff
@@ -825,45 +850,24 @@ PaError PaPulseAudio_StartStreamCb( PaStream * s )
...
-                    pa_stream_set_underflow_callback( stream->outputStream,
-                                                      PaPulseAudio_StreamUnderflowCb,
-                                                      stream);
-                    pa_stream_set_started_callback( stream->outputStream,
-                                                    PaPulseAudio_StreamStartedCb,
-                                                    stream );
-                }
-                else
-                {
+                    PA_PULSEAUDIO_SET_LAST_HOST_ERROR( 0,
```
RossBencina, line 860: Still broken here. `PA_PULSEAUDIO_SET_LAST_HOST_ERROR` must only be used when returning `paUnanticipatedHostError` which is not the case here. Also, passing 0 to `PA_PULSEAUDIO_SET_LAST_HOST_ERROR` makes no sense, that is not a pulseaudio error code. You should be passing the error code from pulse audio.

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
```diff
@@ -717,41 +760,23 @@ PaError PaPulseAudio_StartStreamCb( PaStream * s )
...
                                             &stream->inputBufferAttr,
                                             pulseaudioStreamFlags ) )
             {
-                pa_stream_set_started_callback( stream->inputStream,
-                                                PaPulseAudio_StreamStartedCb,
-                                                stream );
-            }
-            else
-            {
+                PA_PULSEAUDIO_SET_LAST_HOST_ERROR( 0,
```
RossBencina, line 768: Still broken here. `PA_PULSEAUDIO_SET_LAST_HOST_ERROR` must only be used when returning `paUnanticipatedHostError` which is not the case here. Also, passing 0 to `PA_PULSEAUDIO_SET_LAST_HOST_ERROR` makes no sense, that is not a pulseaudio error code. You should be passing the error code from pulse audio.

src/hostapi/pulseaudio/pa_linux_pulseaudio.c
philburk: There is still a problem.

        PA_PULSEAUDIO_SET_LAST_HOST_ERROR( 0,
                                           "PaPulseAudio_BlockingInitRingBuffer: Not enough memory to handle request" );
        return paInsufficientMemory;

The code should not use PA_PULSEAUDIO_SET_LAST_HOST_ERROR when there is not a host error code.
If you are passing back a "pa" error then you should not set it.
You could use PA_DEBUG() instead of PA_PULSEAUDIO_SET_LAST_HOST_ERROR() if you want to log at that point.

RossBencina: Also, always prefer returning PA error codes to `paUnanticipatedHostError`

illuusio: Ok now everything is PA_DEBUG and no more error reporting with `PA_PULSEAUDIO_SET_LAST_HOST_ERROR` . I have learn how `ENSURE_` works on OSS and ALSA to understand how they should behave but now it should be ok as no more `paUnanticipatedHostError`.

src/hostapi/pulseaudio/pa_linux_pulseaudio.c
```diff
@@ -92,14 +92,14 @@ int PaPulseAudio_CheckConnection( pa_context_state_t state )
             break;
 
             case PA_CONTEXT_TERMINATED:
-                PA_PULSEAUDIO_SET_LAST_HOST_ERROR( 0,
-                                                  "Connection terminated to Pulseaudio server. (PA_CONTEXT_TERMINATED)" );
+                PA_DEBUG( ("Portaudio %s: Connection terminated to Pulseaudio server. (PA_CONTEXT_TERMINATED)",
+                           __FUNCTION__) );
                 retCode = paUnanticipatedHostError;
```
philburk, line 97: This code returns paUnanticipatedHostError.
So it should be setting the LastHostErrorInfo.

https://files.portaudio.com/docs/v19-doxydocs/portaudio_8h.html#aad573f208b60577f21d2777a7c5054e0

So it should be the way it was before.

src/hostapi/pulseaudio/pa_linux_pulseaudio.c
```diff
@@ -92,14 +92,14 @@ int PaPulseAudio_CheckConnection( pa_context_state_t state )
...
             break;
 
             case PA_CONTEXT_FAILED:
-                PA_PULSEAUDIO_SET_LAST_HOST_ERROR( 0,
-                                                  "Connection failed or was disconnected from Pulseaudio server. (PA_CONTEXT_FAILED)" );
+                PA_DEBUG( ("Portaudio %s: Connection failed or was disconnected from Pulseaudio server. (PA_CONTEXT_FAILED)",
+                           __FUNCTION__) );
```
philburk, line 102: ditto

philburk (changes requested):

kleinerm:
Just gave this pull request a test and it doesn't work at all, at least on Ubuntu 20.04.6-LTS.

Playback in callback mode just hangs, with the paCallback not ever getting called.

Capture dies with "Assertion 's' failed at pulse/stream.c:335, function pa_stream_get_state(). Aborting."

Backtrace in the capture abort:
```
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff5408859 in __GI_abort () at abort.c:79
#2  0x00007ffff5174973 in pa_stream_get_state () at /lib/x86_64-linux-gnu/libpulse.so.0
#3  0x00007ffff7fb1bae in _PaPulseAudio_WaitStreamState () at ./libportaudio.so.19.8
#4  0x00007ffff7fb1eef in PaPulseAudio_StartStreamCb () at ./libportaudio.so.19.8
#5  0x00007ffff7f97643 in Pa_StartStream () at ./libportaudio.so.19.8
#6  0x00007fffc62ab45c in PSYCHPORTAUDIOStartAudioDevice () at Common/PsychPortAudio/PsychPortAudio.c:4928
#7  0x00007fffc629f19a in mexFunction (nlhs=<optimized out>, plhs=<optimized out>, nrhs=5, prhs=0x7fffbcb83120) at Common/Base/PsychScriptingGlueMatlab.c:368
#8  0x00007ffff7798524 in call_mex(octave_mex_function&, octave_value_list const&, int) () at /lib/x86_64-linux-gnu/liboctinterp.so.7
#9  0x00007ffff72686a5 in octave_mex_function::call(octave::tree_evaluator&, int, octave_value_list const&) () at /lib/x86_64-linux-gnu/liboctinterp.so.7
```

It kind of looks as if the same problem as in PR #857 exists, but I don't have time to debug your new changes. See my proposal in that PR though for something that I think might solve that problem in a simple manner.

Sorry for not bringing more pleasant feedback as a christmas present :/.

I still hope that a Portaudio release with the pulseaudio backend could make it into the next big Ubuntu 24.04-LTS release. That would be splendid, but would require new portaudio packages to land in Debian unstable before February 29 2024, which is when Ubuntu stops importing packages from Debian.

Have nice christmas and happy new year!

dmitrykos:
@illuusio please integrate changes from the master to make CI work.

illuusio:
> @illuusio please integrate changes from the master to make CI work/succeed.

Thank you for fixing some not working stuff on master. I've updated to this PR to current HEAD.

illuusio:
> Just gave this pull request a test and it doesn't work at all, at least on Ubuntu 20.04.6-LTS.
> 
> Playback in callback mode just hangs, with the paCallback not ever getting called.
> 
> Capture dies with "Assertion 's' failed at pulse/stream.c:335, function pa_stream_get_state(). Aborting."
> 
> Backtrace in the capture abort:
> 
> ```
> (gdb) bt
> #0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
> #1  0x00007ffff5408859 in __GI_abort () at abort.c:79
> #2  0x00007ffff5174973 in pa_stream_get_state () at /lib/x86_64-linux-gnu/libpulse.so.0
> #3  0x00007ffff7fb1bae in _PaPulseAudio_WaitStreamState () at ./libportaudio.so.19.8
> #4  0x00007ffff7fb1eef in PaPulseAudio_StartStreamCb () at ./libportaudio.so.19.8
> #5  0x00007ffff7f97643 in Pa_StartStream () at ./libportaudio.so.19.8
> #6  0x00007fffc62ab45c in PSYCHPORTAUDIOStartAudioDevice () at Common/PsychPortAudio/PsychPortAudio.c:4928
> #7  0x00007fffc629f19a in mexFunction (nlhs=<optimized out>, plhs=<optimized out>, nrhs=5, prhs=0x7fffbcb83120) at Common/Base/PsychScriptingGlueMatlab.c:368
> #8  0x00007ffff7798524 in call_mex(octave_mex_function&, octave_value_list const&, int) () at /lib/x86_64-linux-gnu/liboctinterp.so.7
> #9  0x00007ffff72686a5 in octave_mex_function::call(octave::tree_evaluator&, int, octave_value_list const&) () at /lib/x86_64-linux-gnu/liboctinterp.so.7
> ```
> 
> It kind of looks as if the same problem as in PR #857 exists, but I don't have time to debug your new changes. See my proposal in that PR though for something that I think might solve that problem in a simple manner.
> 
> Sorry for not bringing more pleasant feedback as a christmas present :/.
> 
> I still hope that a Portaudio release with the pulseaudio backend could make it into the next big Ubuntu 24.04-LTS release. That would be splendid, but would require new portaudio packages to land in Debian unstable before February 29 2024, which is when Ubuntu stops importing packages from Debian.
> 
> Have nice christmas and happy new year!

Thank you for testing. I've little bit reworked code and hopefully now it works.

RossBencina:
@kleinerm is it working for you now?

src/hostapi/pulseaudio/pa_linux_pulseaudio.c
```diff
@@ -67,50 +67,65 @@ extern char *__progname;
...
-        PA_PULSEAUDIO_SET_LAST_HOST_ERROR( 0,
-                                           "PaPulseAudio_CheckConnection: Host API is NULL! Can't do anything about it" );
-        return -1;
+        return retCode;
     }
 
-    if( !ptr->context || !ptr->mainloop )
+    pa_context_state_t state = pa_context_get_state( ptr->context );
+
+    /* Check if contenxt if good.
```
philburk, line 79: c/contenxt/context/

src/hostapi/pulseaudio/pa_linux_pulseaudio.c
```diff
@@ -67,50 +67,65 @@ extern char *__progname;
...
+            case PA_CONTEXT_CONNECTING:
+            case PA_CONTEXT_AUTHORIZING:
+            case PA_CONTEXT_SETTING_NAME:
+                /* These are just here if they are needed in future */
+            break;
+        }
+    }
+    else
     {
+        retCode = paUnanticipatedHostError;
```
RossBencina, line 103:
```suggestion
```

Just delete this line and continue with `paNotInitialized` from above. You can't return `paUnanticipatedHostError;` without setting the host error.

src/hostapi/pulseaudio/pa_linux_pulseaudio.c
```diff
@@ -577,6 +598,8 @@ PaError PaPulseAudio_Initialize( PaUtilHostApiRepresentation ** hostApi,
 
     if( ret < 0 )
     {
+        PA_DEBUG( ("Portaudio %s: Can't connect to server",
+                   __FUNCTION__) );
         PA_PULSEAUDIO_SET_LAST_HOST_ERROR( 0,
```
RossBencina, line 603:
```suggestion
        PA_PULSEAUDIO_SET_LAST_HOST_ERROR( ret,
```

RossBencina (changes requested):
If the two changes I've suggested are accepted, and @kleinerm says it works (in either order) then I'm ready to approve this.

src/hostapi/pulseaudio/pa_linux_pulseaudio.c
kleinerm: Also, "is good" instead of "if good"

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
```diff
@@ -717,16 +760,10 @@ PaError PaPulseAudio_StartStreamCb( PaStream * s )
...
                                             pulseaudioStreamFlags ) )
-            {
-                pa_stream_set_started_callback( stream->inputStream,
-                                                PaPulseAudio_StreamStartedCb,
-                                                stream );
-            }
-            else
             {
                 PA_DEBUG( ("Portaudio %s: Can't read audio!\n",
                           __FUNCTION__) );
```
kleinerm, line 769:
```suggestion
                          __FUNCTION__) );
                   PaPulseAudio_UnLock( pulseaudioHostApi->mainloop );
```
UnLock in error else-case is missing.

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
```diff
@@ -825,45 +848,22 @@ PaError PaPulseAudio_StartStreamCb( PaStream * s )
...
-                                                      PaPulseAudio_StreamUnderflowCb,
-                                                      stream);
-                    pa_stream_set_started_callback( stream->outputStream,
-                                                    PaPulseAudio_StreamStartedCb,
-                                                    stream );
-                }
-                else
                 {
                     PA_DEBUG( ("Portaudio %s: Can't write audio!\n",
                               __FUNCTION__) );
```
kleinerm, line 859:
```suggestion
                              __FUNCTION__) );
                       PaPulseAudio_UnLock( pulseaudioHostApi->mainloop );
```
UnLock in error else-case is missing.

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
```diff
@@ -735,23 +772,9 @@ PaError PaPulseAudio_StartStreamCb( PaStream * s )
             }
             PaPulseAudio_UnLock( pulseaudioHostApi->mainloop );
 
-            for( waitLoop = 0; waitLoop < 100; waitLoop ++ )
+            if( _PaPulseAudio_WaitStreamState( pulseaudioHostApi->mainloop, stream->outputStream ) != paNoError )
```
kleinerm, line 775:
```suggestion
            if( _PaPulseAudio_WaitStreamState( pulseaudioHostApi->mainloop, stream->inputStream ) != paNoError )
```
This is the setup and wait path for inputStream, not outputStream.

kleinerm:
Apart from my review comments, the problem persists, with exactly the same symptoms as mentioned before:

Hangs when trying to start playback, callback never called.
Dies with assertion when trying to start capture - updated backtrace now:

```
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff5406859 in __GI_abort () at abort.c:79
#2  0x00007ffff5172973 in pa_stream_get_state () at /lib/x86_64-linux-gnu/libpulse.so.0
#3  0x00007ffff7fb0d45 in _PaPulseAudio_WaitStreamState () at ./libportaudio.so.19.8
#4  0x00007ffff7fb1086 in PaPulseAudio_StartStreamCb () at ./libportaudio.so.19.8
#5  0x00007ffff7f96643 in Pa_StartStream () at ./libportaudio.so.19.8
#6  0x00007fffc62ab3fc in PSYCHPORTAUDIOStartAudioDevice ()
    at /home/kleinerm/projects/OpenGLPsychtoolbox/Psychtoolbox-3/Psychtoolbox/PsychBasic/Octave5LinuxFiles64/PsychPortAudio.mex
#7  0x00007fffc629f13a in mexFunction ()
    at /home/kleinerm/projects/OpenGLPsychtoolbox/Psychtoolbox-3/Psychtoolbox/PsychBasic/Octave5LinuxFiles64/PsychPortAudio.mex
#8  0x00007ffff7796524 in call_mex(octave_mex_function&, octave_value_list const&, int) () at /lib/x86_64-linux-gnu/liboctinterp.so.7
#9  0x00007ffff72666a5 in octave_mex_function::call(octave::tree_evaluator&, int, octave_value_list const&) () at /lib/x86_64-linux-gnu/liboctinterp.so.7

```

src/hostapi/pulseaudio/pa_linux_pulseaudio.c
illuusio: I messed one line badly.. :smile:

illuusio:
@RossBencina  and @kleinerm thank you for reviewing this. I'll update the suggestion as I have some time to what's them thru. They seem to all to good stuff..

RossBencina:
Please fix the whitespace issues flagged by CI:

```
error: src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c(770) bad indent: [19](https://github.com/PortAudio/portaudio/actions/runs/7728303666/job/21068771570?pr=863#step:4:20)
b'                   PaPulseAudio_UnLock( pulseaudioHostApi->mainloop );'
error: src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c(861) bad indent: 23
b'                       PaPulseAudio_UnLock( pulseaudioHostApi->mainloop );'
error: src/hostapi/pulseaudio/pa_linux_pulseaudio.c(104) trailing whitespace:
b'        '
```

illuusio:
> Please fix the whitespace issues flagged by CI:
> 
> ```
> error: src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c(770) bad indent: [19](https://github.com/PortAudio/portaudio/actions/runs/7728303666/job/21068771570?pr=863#step:4:20)
> b'                   PaPulseAudio_UnLock( pulseaudioHostApi->mainloop );'
> error: src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c(861) bad indent: 23
> b'                       PaPulseAudio_UnLock( pulseaudioHostApi->mainloop );'
> error: src/hostapi/pulseaudio/pa_linux_pulseaudio.c(104) trailing whitespace:
> b'        '
> ```

Yes I've noted that and didn't got time to fix. It had unfinished things anyway so was not so biggie anyway. I've little bit cleaned up things as I by myself could not remember why I added `isActive` and `isStopped` as they are same thing. Long story short logic behind starting playback and record should now work as it should.. I've tested limited and if @kleinerm  could once again see it my changes makes any difference.

RossBencina:
> could not remember why I added isActive and isStopped as they are same thing

They are not necessarily the same thing, see: https://www.portaudio.com/docs/proposals/010-ClarifyStreamStateMachine.html

kleinerm:
So this is still totally broken, to no surprise at all, as you ignored my proposed and tested fix a 3rd time, and even more broken since your latest commit from yesterday.

If I revert your latest commit "[pulseaudio: Rework callback starting](https://github.com/PortAudio/portaudio/pull/863/commits/96f1235507020b52cd4085b4bdd280bdb4e1507d)", and then apply my proposed fix, stuff works again in playback, recording and simultaneous playback & recording, both half-duplex and full-duplex.

Given that it is now pretty unlikely that a new Portaudio release with your pulseaudio backend will make it into important spring 2024 Linux releases like Ubuntu 24.04-LTS, which was my goal and hope in helping you, and I have now spent over 99 hours of work testing, debugging and helping, I can't justify putting more work into this.

Some recent testing by myself against the current Ubuntu 24.04 daily snapshots with Pipewire 1.0 and its new ALSA pipewire plugin, which acts as a new default ALSA audio device, it looks like going Portaudio ALSA hostApi -> ALSA pipewire default audio device -> Pipewire on upcoming 2024 distros might be an as good as or better solution for many Portaudio sound apps.

I'm happy though to test one last time, after others have successfully reviewed and tested the code to satisfaction.

illuusio:
> Some recent testing by myself against the current Ubuntu 24.04 daily snapshots with Pipewire 1.0 and its new ALSA pipewire plugin, which acts as a new default ALSA audio device, it looks like going Portaudio ALSA hostApi -> ALSA pipewire default audio device -> Pipewire on upcoming 2024 distros might be an as good as or better solution for many Portaudio sound apps.

Yes there should be Pipewire Portaudio HostAPI. When I started writing Pulseaudio HostAPI it was not sure how it goes with Pipewire.  It should be much easier to Hostapi than Pulseaudio which goes very badly with API like Portaudio. Pulseaudio API will be just kind of curiosity and for situation when you don't have anything else. Of course there should be  someone to write HostAPI..

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
```diff
@@ -237,20 +237,19 @@ static int _PaPulseAudio_ProcessAudio(PaPulseAudio_Stream *stream,
...
 
-    /*
-     * This can be called before we have reached out
-     * starting Portaudio stream or Portaudio stream
-     * is stopped
-     */
-    if( !stream->isActive && stream->pulseaudioIsActive )
+    if( !stream->isActive && stream->pulseaudioIsActive && stream->outputStream)
     {
         if(stream->outputStream)
```
kleinerm, line 254: This `if(stream-outputStream)` is now redundant due to the new `&& stream->outputStream` in the enclosing if statement above. Could be removed, but is purely cosmetic.

illuusio:
> Wrt. the change in suggestedLatency handling, I'll leave this to @RossBencina and @philburk , as it doesn't affect my software in any way - I don't ever pass in negative or zero suggestedLatency values. But treating a zero suggestedLatency as DEFAULT_MIN_LATENCY is not strictly what the developers docs for Portaudio recommend - See https://github.com/PortAudio/portaudio/wiki/BufferingLatencyAndTimingImplementationGuidelines#user-model-latency-fields-and-parameters and the new issue #856 as a result of previous discussions we had here a few months ago.
> 
> So I'm not sure if working around application bugs by mapping a requested latency of zero to 15 msecs is really the proper thing to do. But I personally don't care.

Yes this one thing that I've though when implemented. It's not proper thing to do but as 0 latency works incorrect this lesser bad than un working audio but if there is some other advice it would help a lot..

philburk:
Thanks for persevering with this host implementation. It sounds like you are close.

Clipping zero requested latency to a minimum value is a good thing.
It is best if the application stays between the recommended min and max values.
If an app tries to push outside that range then results are not guaranteed.
A value of zero is, of course, impossible and should be considered a request for the lowest possible latency.
If the latency is pushed extremely low then an app will have to expect an occasional glitch.

src/hostapi/pulseaudio/pa_linux_pulseaudio.c
```diff
@@ -1122,13 +1147,13 @@ PaError OpenStream( struct PaUtilHostApiRepresentation *hostApi,
...
         /* Convert positive suggestedLatency from seconds to microseconds, otherwise default to zero. */
-        if (inputParameters->suggestedLatency >= 0)
+        if ( inputParameters->suggestedLatency > 0.0 )
         {
             stream->suggestedLatencyUSecs = (unsigned int) (inputParameters->suggestedLatency * 1e6 + 1.0f);
         }
         else
         {
-            stream->suggestedLatencyUSecs = 0;
+            stream->suggestedLatencyUSecs = (unsigned int) (PA_PULSEAUDIO_DEFAULT_MIN_LATENCY * 1e6 + 1.0f);
```
RossBencina, line 1156: `inputParameters->suggestedLatency` of `0` should result in the same buffer size as suggested latency `0 + epsilon`. So maybe just set this to 1usec here if 0 isn't working.

RossBencina:
I don't think the zero suggested latency handling is correct but we plan to merge it next week anyhow since the main issue is resolved and the suggested latency handling can be addressed in a separate PR.

illuusio:
Ok I'll revert my changes and let latency stuff be merged. Then I'll can figure this out if needed or fix this straight to pure data as latency they determine in `pd-gui` is 25 ms but it's set somehow to 0 when starting portaudio stream which is not correct.

src/hostapi/pulseaudio/pa_linux_pulseaudio.c
illuusio: Ok this in reverted now. Sorry I didn't understand the epsilon thing.

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
illuusio: Ok I've fixed this. Thank you for spotting.

src/hostapi/pulseaudio/pa_linux_pulseaudio.c
illuusio: It's not done anymore so I assume this is fixed.

illuusio:
As now it's working again I'm starting to hope this will be merged soon but let's see if there's something to fix still..

kleinerm:
Retested on Ubuntu 20.04.6-LTS to be still fine.

Fwiw, I looked at some pulseaudio server code, and lines like these ...
https://gitlab.freedesktop.org/pulseaudio/pulseaudio/-/blob/master/src/pulsecore/memblockq.c?ref_type=heads#L863
... suggest that a tlength == 0 is treated like tlength == -1, which is highest latency possible ~2 seconds.

Other code fragments, e.g., https://gitlab.freedesktop.org/pulseaudio/pulseaudio/-/blob/master/src/pulsecore/memblockq.c?ref_type=heads#L852 in combination with https://gitlab.freedesktop.org/pulseaudio/pulseaudio/-/blob/master/src/pulsecore/memblockq.c?ref_type=heads#L81 suggest that the lowest implementable value may be one audio frame size, ie. `pa_frame_size(sample_spec)` in that code. Similar, e.g., https://gitlab.freedesktop.org/pulseaudio/pulseaudio/-/blob/master/src/pulsecore/protocol-native.c#L412

So what could be worth trying is making sure that 
``stream->inputBufferAttr.fragsize`` and ``stream->outputBufferAttr.tlength``
never get lower than `pa_frame_size(stream->inputSampleSpec)` and `pa_frame_size(stream->outputSampleSpec)`.
Ie. add a check and override assignment in `PaPulseAudio_StartStreamCb()`.

This way 0 usecs or too small usecs would result in an assignment of at least pa_frame_size() bytes.

RossBencina:
> So what could be worth trying is making sure that
> stream->inputBufferAttr.fragsize and stream->outputBufferAttr.tlength
> never get lower than pa_frame_size(stream->inputSampleSpec) and pa_frame_size(stream->outputSampleSpec).
> Ie. add a check and override assignment in PaPulseAudio_StartStreamCb().
> 
> This way 0 usecs or too small usecs would result in an assignment of at least pa_frame_size() bytes.

Sounds good to me.

Is there anything remaining for this PR before we merge it? Are we ok to merge this in 7 days?

kleinerm:
@RossBencina From my side and testing, merging is fine.

illuusio:
Same to me. I'll address these latency problems in other PR.

philburk (approved):

src/hostapi/pulseaudio/pa_linux_pulseaudio.c
```diff
@@ -67,50 +67,75 @@ extern char *__progname;
 /* PulseAudio specific functions */
 int PaPulseAudio_CheckConnection( PaPulseAudio_HostApiRepresentation * ptr )
 {
-    pa_context_state_t state;
-
-
-    /* Sanity check if ptr if NULL don't go anywhere or
-     * it will SIGSEGV
+    /*
+     * This is bit hackish as -1 is consired as error
```
RossBencina, line 71:
```suggestion
     * This is bit hackish as -1 is considered as an error
```

philburk (approved):

RossBencina (approved):
Seems good to me we've done enough hopefully we didn't miss anything.

RossBencina:
Thank you everyone :)

Issue #777, opened 2023-02-27
why is vcpkg a dependency on macOS cmake CI builds?


RossBencina:
We just experienced a CI failure due to missing vcpkg dependency (see #767).

Why do we have this dependency on macOS? It caused a CI failure. Can we remove the dependency to avoid future issues.

RossBencina:
@Be-ing assigning this to you for comment since you introduced vcpkg.

Be-ing:
vcpkg is used for JACK. If you remove it, you'll need to setup another way to test JACK on CI.

RossBencina:
@Be-ing I see. I guess we use vcpkg for JACK on Windows already and that's why it's being done that way? 

Do you see a problem with using brew for JACK on Mac OS? (I mean, if we rely on brew to install vcpkg to install JACK, why not just install JACK directly with brew).

Be-ing:
I suppose you could replace JACK from vcpkg with JACK from Homebrew for macOS.

PR #866, opened 2023-12-02
Add a test that tries to use the generated CMake libs [continuation].

dmitrykos:
It is a continuation of stalled (and thus closed) PR #671.


dmitrykos:
Besides fix of #671 this PR also fixes CI pipeline due to outdated vcpkg commit Id that was causing MSVC and MinGW build failure although was working well before. The reason was missing dependency `msys-mingw-w64-i686-libwinpthread-git-9.0.0.6373.5be8fcd83-1-any.pkg.tar.zst` of vcpkg.

Now it is set to the latest (https://github.com/microsoft/vcpkg/commit/ca9ac0ba65965937fb66783c4f726c2c755ad9d9) and works well.

RossBencina:
Does this remove the vcpkg dependency on Mac CI? (i.e. does it close #777)

.github/workflows/cmake.yml
```diff
@@ -10,28 +10,23 @@ jobs:
...
             asio_sdk_cache_path: asiosdk.zip
             dependencies_extras: mingw-w64
-            # FIXME: linking JACK fails with vcpkg. Switch the CMake toolchain file to use vcpkg. The
-            # toolchain file in this repository is not needed when using a MinGW triplet with vcpkg.
             vcpkg_triplet: x64-mingw-static
             cmake_generator: "Unix Makefiles"
+            # FIXME: linking vcpkg Jack fais due to MinGW-W64 linker assertion. Disable Jack for this build.
             cmake_options:
               -DPA_USE_ASIO=ON
+              -DPA_USE_JACK=OFF
```
RossBencina, line 26: Could you create a ticket regarding this issue please?

philburk (approved):
Thanks for figuring this out!
I don't fully understand it but CI is passing.
So that is a good thing.

RossBencina (approved):
Thanks Dmitry. I don't understand what you've done here but I'm OK moving forward with this.

.github/workflows/cmake.yml
dmitrykos: There is linker error, so it is outside PA and either Jack dependency must be fixed or MinGW-64 linker has to be changed. I can create ticket to trace this issue although we could simply revise `DPA_USE_JACK=OFF` after some time.

PR #881, opened 2024-01-29
Add support for iOS

### LINKED ISSUE #156: PortAudio implementation for iOS

Create a PortAudio implementation for iOS

A number of people on the mailing list have code for audio i/o on iOS that could be combined into the basis of a PA implementation.

### LINKED ISSUE #749: Ios deprecation
Hello all.

I'm trying to build portaudio for Ios, and I think I face some deprecated types.
```
../vendor/portaudio/src/hostapi/coreaudio/pa_mac_core_utilities.h:115:5: error: unknown type name 'AudioHardwarePropertyID'; did you mean 'AudioFormatPropertyID'?
    AudioHardwarePropertyID inPropertyID,
    ^~~~~~~~~~~~~~~~~~~~~~~
    AudioFormatPropertyID
```

```
../vendor/portaudio/src/hostapi/coreaudio/pa_mac_core_utilities.h:124:5: error: unknown type name 'AudioDeviceID'; did you mean 'AudioFileID'?
    AudioDeviceID         inDevice,
    ^~~~~~~~~~~~~
    AudioFileID
``
`


These defines are in AudioHardwareDeprecated.h, which not included in CoreAudio.h

What is the procedure to solve the problem :) ?

Thank you

### PR DESCRIPTION
**Pull request by fwcd**
### Fixes #156, fixes #749

This is a rebased version of Hans Petter Selasky's, adding support for Core Audio on iOS. I have had success building this for an experimental iOS port of the DJ app Mixxx ([see here](https://github.com/fwcd/m1xxx/issues/16)), but am marking this as draft for now, since for some reason only headphone output worked for me, not the speakers.

Something else to be investigated would be whether https://github.com/PortAudio/portaudio/issues/749#issuecomment-1502440205 contains any further patches that we might want to integrate. The posted Xcode project might prove useful for testing too (though that of course doesn't belong into this repo).

In any case, here's a PR for early review.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -296,14 +296,25 @@ elseif(UNIX)
 
   if(APPLE)
     set(CMAKE_MACOSX_RPATH 1)
-    target_sources(PortAudio PRIVATE
-      src/hostapi/coreaudio/pa_mac_core.c
-      src/hostapi/coreaudio/pa_mac_core_blocking.c
-      src/hostapi/coreaudio/pa_mac_core_blocking.h
-      src/hostapi/coreaudio/pa_mac_core_internal.h
-      src/hostapi/coreaudio/pa_mac_core_utilities.c
-      src/hostapi/coreaudio/pa_mac_core_utilities.h
-    )
+    if(IOS)
+      target_sources(PortAudio PRIVATE
+        src/hostapi/coreaudio_ios/pa_ios_core.c
+        src/hostapi/coreaudio_ios/pa_ios_core_blocking.c
+        src/hostapi/coreaudio_ios/pa_ios_core_blocking.h
+        src/hostapi/coreaudio_ios/pa_ios_core_internal.h
+        src/hostapi/coreaudio_ios/pa_ios_core_utilities.c
+        src/hostapi/coreaudio_ios/pa_ios_core_utilities.h
+      )
+    else()
+      target_sources(PortAudio PRIVATE
+        src/hostapi/coreaudio/pa_mac_core.c
+        src/hostapi/coreaudio/pa_mac_core_blocking.c
+        src/hostapi/coreaudio/pa_mac_core_blocking.h
+        src/hostapi/coreaudio/pa_mac_core_internal.h
+        src/hostapi/coreaudio/pa_mac_core_utilities.c
+        src/hostapi/coreaudio/pa_mac_core_utilities.h
+      )
+    endif()
     target_include_directories(PortAudio PRIVATE src/hostapi/coreaudio)
     set(PORTAUDIO_PUBLIC_HEADERS "${PORTAUDIO_PUBLIC_HEADERS}" include/pa_mac_core.h)
 
@@ -315,8 +326,12 @@ elseif(UNIX)
         -Wl,-framework,CoreFoundation
         -Wl,-framework,CoreServices
     )
-    target_compile_definitions(PortAudio PUBLIC PA_USE_COREAUDIO=1)
-    set(PKGCONFIG_CFLAGS "${PKGCONFIG_CFLAGS} -DPA_USE_COREAUDIO=1")
+    if(IOS)
+      target_compile_definitions(PortAudio PUBLIC PA_USE_COREAUDIO_IOS=1)
+    else()
+      target_compile_definitions(PortAudio PUBLIC PA_USE_COREAUDIO=1)
+      set(PKGCONFIG_CFLAGS "${PKGCONFIG_CFLAGS} -DPA_USE_COREAUDIO=1")
+    endif()
 
     # Use C11 so that we can make use of atomic library and avoid deprecation errors.
     set_property(TARGET PortAudio PROPERTY C_STANDARD 11)
--- a/src/hostapi/coreaudio_ios/pa_ios_core.c
+++ b/src/hostapi/coreaudio_ios/pa_ios_core.c
@@ -1,0 +1,1429 @@
+/*
+ * Implementation of the PortAudio API for Apple AUHAL
+ *
+ * PortAudio Portable Real-Time Audio Library
+ * Latest Version at: http://www.portaudio.com
+ *
+ * Written by Bjorn Roche of XO Audio LLC, from PA skeleton code.
+ * Portions copied from code by Dominic Mazzoni (who wrote a HAL implementation)
+ *
+ * Dominic's code was based on code by Phil Burk, Darren Gibbs,
+ * Gord Peters, Stephane Letz, and Greg Pfiel.
+ *
+ * The following people also deserve acknowledgements:
+ *
+ * Olivier Tristan for feedback and testing
+ * Glenn Zelniker and Z-Systems engineering for sponsoring the Blocking I/O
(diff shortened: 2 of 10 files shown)

### CODE REVIEW

**Comment by fwcd** [comment]
Did my best to try fixing all of the formatting issues (manually, the patch contained a strange mix of 3-space, 4-space and tab indentation and the VSCode formatter would have preferred a different style to what the rest of the project uses).

The `pa_whitelint.py` check now passes.

File: src/hostapi/coreaudio_ios/pa_ios_core.c [hunk 1]
```diff
@@ -0,0 +1,1429 @@
...
+    const AudioTimeStamp * inTimeStamp,
+    UInt32 inBusNumber,
+    UInt32 inNumberFrames,
+    AudioBufferList * ioData);
+static double GetStreamCpuLoad(PaStream * stream);
+
+/*
+ * Callback called when starting or stopping a stream.
+ */
+static void
```
RossBencina, line 108: Typically we put the qualifier and return type on the same line as the function name, could you do that please?

RossBencina:
> iOS does set up a default audio session

In fact, the audio session is a singleton, so it's not really a question of whether to set one up, but whether we're happy with the defaults. The docs say "Although the default audio session provides useful behavior, it generally doesn’t provide the audio behavior a media app needs. To change the default behavior, you configure your app’s audio session category." [1]

I notice that you mention the playback category, but if full-duplex is required that will not work.[2]

Let's stew on this. 

I'd like to get some other people to confirm this PR works for them before merging.

[1] https://developer.apple.com/documentation/avfaudio/avaudiosession
[2] https://developer.apple.com/documentation/avfaudio/avaudiosession/category

philburk:
@fwcd - thanks for doing this! it will be great to have iOS support.

> This is a rebased version of Hans Petter Selasky's patch (originally taken from [here](https://stackoverflow.com/questions/71481460/how-to-patch-portaudio-for-low-latency-ios-usage)),

Did your first commit have changes beyond just those needed for rebasing?

CMakeLists.txt
```diff
@@ -296,14 +296,25 @@ elseif(UNIX)
...
     set(CMAKE_MACOSX_RPATH 1)
-    target_sources(PortAudio PRIVATE
-      src/hostapi/coreaudio/pa_mac_core.c
-      src/hostapi/coreaudio/pa_mac_core_blocking.c
-      src/hostapi/coreaudio/pa_mac_core_blocking.h
-      src/hostapi/coreaudio/pa_mac_core_internal.h
-      src/hostapi/coreaudio/pa_mac_core_utilities.c
-      src/hostapi/coreaudio/pa_mac_core_utilities.h
-    )
+    if(IOS)
```
philburk, line 299: At some point, we will need a GitHub Actions for Continuous Integration testing for iOS.

philburk (changes requested):

fwcd:
> In fact, the audio session is a singleton, so it's not really a question of whether to set one up, but whether we're happy with the defaults.

Yeah, I might have phrased it badly, I meant setting up the singleton (as in the code snippet). The playback category is just an example there. The defaults behaving weirdly, in the sense that speaker output wouldn't work, was something that I didn't (and still don't) really understand. So I would appreciate if someone could comment on that.

Another, more minor, issue I see with configuring the audio session from PortAudio is that it will introduce a new dependency on a much higher-level framework (`AVFAudio`) compared to Core Audio, which PortAudio currently uses. I am not familiar enough with iOS' audio stack to comment on whether there is a way to express this solely in terms of Core Audio's API, but that is probably something we would not want to do anyway, given that using the `AVAudioSession` API is generally the recommended approach for iOS apps.

> Did your first commit have changes beyond just those needed for rebasing?

No, the first commit is 1-to-1 the HPS patch.

RossBencina:
> The defaults behaving weirdly, in the sense that speaker output wouldn't work, was something that I didn't (and still don't) really understand. 

The docs say that the default won't output audio if the phone is in silent mode, could that be the issue?

fwcd:
Ah, looks like I missed that part, that would definitely explain it :D

But yeah, I guess that just underlines that it's rarely the right behavior for media apps.

CMakeLists.txt
fwcd: I have added an experimental iOS matrix configuration (which is pretty similar to macOS)

RossBencina:
> that would definitely explain it :D

Could you revise and simplify the PR description to reflect the updated state of your understanding please?

fwcd:
Hm, the CI error is strange, I can't reproduce it locally, neither with Makefiles, nor with Ninja, nor with vcpkg:

```
 CMake Error at /usr/local/Cellar/cmake/3.28.0/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find Threads (missing: Threads_FOUND)
```

I found this upstream issue, however: https://gitlab.kitware.com/cmake/cmake/-/issues/18993

src/hostapi/coreaudio_ios/pa_ios_core.c
```diff
@@ -0,0 +1,1412 @@
...
+    }
+    auhalHostApi->allocations = PaUtil_CreateAllocationGroup();
+    if (auhalHostApi->allocations == NULL) {
+        result = paInsufficientMemory;
+        goto error;
+    }
+    *hostApi = &auhalHostApi->inheritedHostApiRep;
+
+    (*hostApi)->info.structVersion = 1;
+    (*hostApi)->info.type = paCoreAudio;
```
fwcd, line 175: Should iOS be using `paInDevelopment` and later get its own type id or is it okay to use the same ID as the macOS Core Audio implementation, given that the two should never be available at the same time?

RossBencina:
Failing iOS clang CI. Please you could you rebase and check. Possibly could be fixed with the fix we did on master: https://github.com/PortAudio/portaudio/commit/7de41c1e07cdc506f4438e5ec9a1315a4ac9330c

jona-1993:
Hello, 

I used your solution for .NET 6 iOS (with dllimport).

But I think you should add this "set(CMAKE_OSX_DEPLOYMENT_TARGET "14.5")" in CMakeLists.txt. Because without this we need the latest version of iOS.

And I have an issue for build that with my iOS application:

1>Xamarin.Shared.Sdk.targets(1560,3): Error  : clang++ exited with code 1:
Undefined symbols for architecture arm64:
  "_PaPthreadUtil_GetTime", referenced from:
      _PaUnixThread_New in portaudio.a(pa_unix_util.c.o)
  "_PaPthreadUtil_NegotiateCondAttrClock", referenced from:
      _PaUnixThread_New in portaudio.a(pa_unix_util.c.o)
ld: symbol(s) not found for architecture arm64
clang: error: linker command failed with exit code 1 (use -v to see invocation)

I think that pa_pthread_util is not built. 
Do you have a solution for this?

calvin2021y:
any update ?

jona-1993:
There is a problem with this. The script to use the native csproj library for iOS is not correct.
And there is another one, you can't publish the app on the Apple Store if you don't use a binding project. I started the development of the binding project but I have little time to finish it.

Issue #641, opened 2021-09-11
pa_jack.c uses deprecated jack_port_get_latency API


Be-ing:
```
be@gsuss-iMac portaudio % cmake -D CMAKE_BUILD_TYPE=Debug -S . -B cmake_build

-- The C compiler identification is AppleClang 12.0.5.12050022
-- The CXX compiler identification is AppleClang 12.0.5.12050022
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /Library/Developer/CommandLineTools/usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /Library/Developer/CommandLineTools/usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Looking for pthread.h
-- Looking for pthread.h - found
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE  
-- Found Regex: POSIX regex.h  
-- Found JACK: /usr/local/lib/libjack.dylib;Regex::regex  
-- Configuring done
-- Generating done
-- Build files have been written to: /Users/be/sw/portaudio/cmake_build
be@gsuss-iMac portaudio % cmake --build cmake_build

...

/Users/be/sw/portaudio/src/hostapi/jack/pa_jack.c:637:17: warning: 'jack_port_get_latency' is deprecated [-Wdeprecated-declarations]
                jack_port_get_latency( p ) / globalSampleRate;
                ^
/usr/local/include/jack/jack.h:1233:58: note: 'jack_port_get_latency' has been explicitly marked deprecated here
jack_nframes_t jack_port_get_latency (jack_port_t *port) JACK_OPTIONAL_WEAK_DEPRECATED_EXPORT;
                                                         ^
/usr/local/include/jack/weakmacros.h:80:61: note: expanded from macro 'JACK_OPTIONAL_WEAK_DEPRECATED_EXPORT'
#define JACK_OPTIONAL_WEAK_DEPRECATED_EXPORT __attribute__((__deprecated__))
                                                            ^
[...]

PR #886, opened 2024-03-02
JACK: do not subtract buffer size from port latency

stefanha:
The latency calculation subtracts the size of one buffer from JACK's port latency value. I suspect this adjustment was made so that jackd's "asynchronous mode" appears to have the user's requested latency value even though jackd adds one extra buffer, thus increasing the real latency.

When jackd runs in "synchronous mode" or the more recent pipewire-jack implementation is used, PortAudio reports incorrect latency values of 0 seconds because the latency is exactly one buffer.

Fix the calculation to produce the real latency.

### DIFF
--- a/src/hostapi/jack/pa_jack.c
+++ b/src/hostapi/jack/pa_jack.c
@@ -1365,11 +1365,9 @@ static PaError OpenStream( struct PaUtilHostApiRepresentation *hostApi,
 
     if( stream->num_incoming_connections > 0 )
         stream->streamRepresentation.streamInfo.inputLatency = (jack_port_get_latency( stream->remote_output_ports[0] )
-                - jack_get_buffer_size( jackHostApi->jack_client )  /* One buffer is not counted as latency */
             + PaUtil_GetBufferProcessorInputLatencyFrames( &stream->bufferProcessor )) / sampleRate;
     if( stream->num_outgoing_connections > 0 )
         stream->streamRepresentation.streamInfo.outputLatency = (jack_port_get_latency( stream->remote_input_ports[0] )
-                - jack_get_buffer_size( jackHostApi->jack_client )  /* One buffer is not counted as latency */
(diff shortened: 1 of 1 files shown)


RossBencina:
related #641

stefanha:
@RossBencina I have pushed a new commit that replaces the deprecated `jack_port_get_latency()` API will `jack_port_get_latency_range()` in order to fix #641.

stefanha:
Hi @RossBencina, please let me know if you have any comments. Thanks!

RossBencina:
I've assigned this to myself since I need to put in the time to review our specifications and think about why this is correct.

RossBencina:
@stefanha sorry about the delay.

I've the calculations against our documentation (https://github.com/PortAudio/portaudio/wiki/BufferingLatencyAndTimingImplementationGuidelines) and Phil and I have agreed that this looks like the correct fix, given that the old code was computing zero latency in both the input and output cases.

One more thing: I'm unclear about the "asynchronous mode" vs "synchronous mode" thing. Is it the case that PortAudio/JACK no longer ever uses "asynchronous mode" under any circumstances?

philburk (approved):
Looks good. Thanks!

stefanha:
> One more thing: I'm unclear about the JACK "asynchronous mode" vs "synchronous mode" thing. Is it the case that PortAudio/JACK no longer ever uses "asynchronous mode" under any circumstances?

@RossBencina Thanks for taking a look at this! The latency reported by the PortAudio JACK host API was incorrect in both modes. Asynchronous mode adds latency, so it is not accurate to subtract the extra latency. I believe the the state of JACK implementations is:
- Modern Linux machines use PipeWire's JACK implementation, which behaves like jackd's synchronous mode.
- Some pro-audio setups still use jackd and the default is asynchronous mode. I guess they will eventually switch to PipeWire.

Either way, the latency reported by the JACK host API was not accurate and this fix will ensure that the real latency is reported.

RossBencina (approved):

RossBencina:
Merged! Thank you!

Issue #366, opened 2020-12-17
JACK client name and port names nondescript/confusing


mxmilkiib:
**Describe the bug**

The JACK client name and port names PortAudio allows/provides are not helpful.

This is [Mixxx](www.mixxx.org):
![image](https://user-images.githubusercontent.com/108225/102420930-28df5c00-3ffb-11eb-9582-9b1a21c161a9.png)

**Expected behavior**

That there could be a way to set an appropriate JACK client name and meaningful port [metadata](https://jackaudio.org/api/group__Metadata.html).

**Desktop (please complete the following information):**
 - OS: multiplatform
 - Host API: JACK

RossBencina:
"That there could be a way to set an appropriate JACK client name and meaningful port metadata." I agree.

mxmilkiib:
Ah, the client name is via the JACK bridge that comes with [alsa-plugins](https://www.archlinux.org/packages/extra/x86_64/alsa-plugins/), and it works regularly to show "Mixxx" with the correct settings.

It would be nice still if the pretty-name port [properties](https://manpages.debian.org/testing/jackd1/jack_property.1.en.html) could be set.

RossBencina:
This is a reasonable enhancement request for a host-specific API to set the port names of a stream. This could done when the stream is opened (parameters to `Pa_OpenStream`) or once the stream is opened (e.g. `PaJack_StreamSetPortName"). I think the latter is better if that's easy to implement.

Anyone is free to submit a PR.

mxmilkiib:
mildly cheeky bump, as this came up yet again in Mixxx community discussions on Zulip

PR #888, opened 2024-03-26
JACK: try to use stable port names

stefanha:
JACK connection managers save/load the connection graph so users don't need to create connections manually every time. This depends on stable port names but PortAudio currently generates incrementing port names throughout the lifetime of a running process.

Reset the port name counters when all streams are closed so that the same port names are generated when the streams are opened again.

This is not a perfect solution because applications that open streams in different orders each time or never close all open streams before restarting audio processing will not reset the port counters. However, typical applications should work with this simple approach and it's an improvement over current behavior.

Tested with Audacity 3.3.3 on Fedora 39.

### DIFF
--- a/src/hostapi/jack/pa_jack.c
+++ b/src/hostapi/jack/pa_jack.c
@@ -1390,6 +1390,26 @@ error:
 }
 
 /*
+ * Reset inputBase and outputBase when all streams have been closed. This makes port names stable for typical
+ * applications that open streams in the same order every time. Applications that open streams in a different order
+ * every time do not get stable port names. Stable port names allow JACK connection managers to save/load connections
+ * between ports, saving the user the trouble of manually connecting ports each time they use the application.
+ */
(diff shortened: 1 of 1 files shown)


RossBencina (approved):
Looks OK to me. I can't test it though. Would be good to get some feedback from PA/JACK users.

RossBencina:
Phil and I are slightly ambivalent on merging this because it is a behavior change that could break user expectations in the case where the reopened streams are unreleated and should *not* be reconnected. Our assessment is:

- Simple applications will not change behavior after this PR
- Applications with only one mode will benefit from this PR
- The above two cases probably cover most applications
- The remaining class of applications with more complex stream usage will have behavior changed from not being re-connected by connection managers to being potentially incorrectly reconnected. It is not clear whether this is actually a worse outcome than the current behavior. Not is it clear whether there are applications that would be impacted.

My position is that for serious use, there should be a way for the application to set the stream names as discussed in #366.

PR #890, opened 2024-04-02
Upgrade Python 2 code to Python 3

cclauss:
Print is a function in Python 3.  https://docs.python.org/3/whatsnew/3.0.html


RossBencina:
Maybe best to add `from __future__ import print_function` to avoid gratuitously breaking Python 2 support.

cclauss:
These changes to NOT break Python 2.

cclauss:
macOS no longer ships with Python.

RossBencina (approved):

RossBencina:
Thank you!

PR #898, opened 2024-04-19
file() was removed in Python 3 in favor of open()

cclauss:
https://portingguide.readthedocs.io/en/latest/builtins.html#removed-file

https://docs.python.org/3/whatsnew/3.0.html#builtins
> Removed the `file` type. Use [`open()`](https://docs.python.org/3/library/functions.html#open). There are now several different kinds of streams that open can return in the [`io`](https://docs.python.org/3/library/io.html#module-io) module.


cclauss:
@RossBencina

Like #890
* #890

RossBencina (approved):

RossBencina:
Thank you!

Issue #895, opened 2024-04-16
PulseAudio:  time is spent perhaps due to insufficient checking of state in PaPulseAudio_CloseStreamCb function


tatsuki-makino:
**Describe the bug**
It uses about 50 seconds (usleep(10000) * 5000) per stream, probably due to insufficient checking of state in PaPulseAudio_CloseStreamCb function.

**To Reproduce**
1. Set up PulseAudio as appropriate.
2. Build PortAudio with PulseAudio enabled.
3. Launch Audacity 3.4.2.
4. Wait until the window appears.

**Expected behavior**
Audacity works smoothly.

**Actual behavior**
The user has to wait about 10 minutes before the window appears.
When touching a device associated with PulseAudio, we get the feeling that a glitch is accumulating.

**Desktop (please complete the following information):**
 - OS: FreeBSD
 - OS Version 12.4-STABLE (It has been tested. All versions are affected.)
 - PortAudio version: 88ab584e7bf4358599744cd662cfbc978f41efbf

**Additional context**
A patch of the experiment is attached.
Instead of just one state, have the inline function PA_STREAM_IS_GOOD check the condition.
However, I do not know if this condition is correct :)
It needs to be verified by someone who is familiar with PulseAudio.
switch and PA_DEBUG are included to know the state.
[patch.txt](https://github.com/PortAudio/portaudio/files/14986527/patch.txt)

illuusio:
Could you please open PR for this as it's easier to follow than just having patch..

illuusio:
I can verify this is happening. One can't use `PA_STREAM_IS_GOOD` as it exited when stream is in state `PA_STREAM_CREATING`. Which means it still not functional but user can start using it.  But as there is 'easy' way to reproduce with Audacity I suppose it can be found why it get stuck.

illuusio:
Now there is fix mostly based on your patch: #897

RossBencina:
#897 has been merged. @tatsuki-makino please confirm that it fixes this issue

tatsuki-makino:
Thank you for accepting the fix.
The code at the commit of db778cf can ostensibly use the input and output from Audacity to PulseAudio without any problems.
But, I don't know which states the pa_stream_disconnect and pa_stream_unref can use for streams and which states it can switch to, so I can't say that this is entirely correct :)
Well, this issue would have reached a state where it could be closed.

One thing that still bothers me is that if we leave the condition for exiting this loop after 
[...]

PR #914, opened 2024-05-20
Pulseaudio: Make start-stop work again by unifying close and stop

illuusio:
PR fixes start and stop by unifying closing, stop and abort. Also by unifying handling process waiting by macro. Fixes also last part of bug #895 as it tries to prevent glitches at the end of stream.

### DIFF
--- a/src/hostapi/pulseaudio/pa_linux_pulseaudio.c
+++ b/src/hostapi/pulseaudio/pa_linux_pulseaudio.c
@@ -51,14 +51,13 @@
 */
 
 
-#include <string.h>     /* strlen() */
-
 #include "pa_linux_pulseaudio_cb_internal.h"
 #include "pa_linux_pulseaudio_block_internal.h"
 
 /* PulseAudio headers */
 #include <stdio.h>
 #include <string.h>
+#include <unistd.h>
 #include <pulse/pulseaudio.h>
 
 /* This is used to identify process name for PulseAudio. */
@@ -563,8 +562,8 @@ PaError PaPulseAudio_Initialize( PaUtilHostApiRepresentation ** hostApi,
                                  PaHostApiIndex hostApiIndex )
 {
     PaError result = paNoError;
-    int i;
-    int deviceCount;
+    int i = 0;
+    int deviceCount = 0;
     int ret = 0;
     int lockTaken = 0;
     PaPulseAudio_HostApiRepresentation *pulseaudioHostApi = NULL;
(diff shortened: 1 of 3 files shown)


tatsuki-makino:
I have experimented with the combination of 2edc379 and audacity-3.5.1 on FreeBSD (but FreeBSD is still 12.4-STABLE).

The behavior of having to manually stop the playback of audio after the entire track or the end of a selection is reached has been suppressed.

As far as the modifications to c420938 are concerned, the stream that has been used up will always pass through the cleanup process.
This was kind of a fart on my part, so I don't know where this is working :)
If the used-by count in the result of pacmd list-sinks drops to 0, does that mean that this problem is not occurring?

illuusio:
> I have experimented with the combination of [2edc379](https://github.com/PortAudio/portaudio/commit/2edc379932656ef88287c56f10fe798fa13754e4) and audacity-3.5.1 on FreeBSD (but FreeBSD is still 12.4-STABLE).
> 
> The behavior of having to manually stop the playback of audio after the entire track or the end of a selection is reached has been suppressed.

There is still some unwanted artifacts? Thanks for sharing how to produce them I can test now. Pity that Audacity is so slow to start. I have to kind of find test that also produces same effect. Does `patest_buffer` do the same?

illuusio:
> As far as the modifications to [c420938](https://github.com/PortAudio/portaudio/commit/c420938072567f9800b6cb2dd01d74f45b1f1524) are concerned, the stream that has been used up will always pass through the cleanup process.
> This was kind of a fart on my part, so I don't know where this is working :)
> If the used-by count in the result of pacmd list-sinks drops to 0, does that mean that this problem is not occurring?

Shouldn't stream be always be cleaned us this is client side and it should clean up us you said. When `ref` drops to zero then all allocated resources should be released on client side. Of course if server does not let us go or somehow hangs (again you are playing on remote machine which is shutdown when you are disconnecting for example) we can't do anything about the server side. Portaudio can just keep own nest clean. But please share if you have more though about this as this is just me trying to improve this and having just own opinion.

illuusio:
I've splitted that macro thing to own #924 PR as I like get that in as it makes my life easier.

illuusio:
> I have experimented with the combination of [2edc379](https://github.com/PortAudio/portaudio/commit/2edc379932656ef88287c56f10fe798fa13754e4) and audacity-3.5.1 on FreeBSD (but FreeBSD is still 12.4-STABLE).
> 
> The behavior of having to manually stop the playback of audio after the entire track or the end of a selection is reached has been suppressed.
> 
> As far as the modifications to [c420938](https://github.com/PortAudio/portaudio/commit/c420938072567f9800b6cb2dd01d74f45b1f1524) are concerned, the stream that has been used up will always pass through the cleanup process. This was kind of a fart on my part, so I don't know where this is working :) If the used-by count in the result of pacmd list-sinks drops to 0, does that mean that this problem is not occurring?

I've tested with ALSA and Jack there is also small click at then end of stream with Audacity which can be heared with headphones. Is just closing of headphone or something else inside Audacity or feature of Portaudio I can't say.

PR #922, opened 2024-06-02
Fix issues with DirectSound thread cleanup code

dechamps:
This PR fixes a couple of issues with the way the DirectSound Host API cleans up its streaming thread, and also simplifies the code as a nice side effect.

Details can be found in the commit messages.

### DIFF
--- a/src/hostapi/dsound/pa_win_ds.c
+++ b/src/hostapi/dsound/pa_win_ds.c
@@ -106,7 +106,7 @@
 
 #if !defined(__CYGWIN__) && !defined(UNDER_CE)
 #define CREATE_THREAD (HANDLE)_beginthreadex
-#undef CLOSE_THREAD_HANDLE /* as per documentation we don't call CloseHandle on a thread created with _beginthreadex */
+#define CLOSE_THREAD_HANDLE CloseHandle
 #define PA_THREAD_FUNC static unsigned WINAPI
 #define PA_THREAD_ID unsigned
 #else
@@ -3051,9 +3051,7 @@ error:
 #ifndef PA_WIN_DS_USE_WMME_TIMER
     if( stream->processingThread )
     {
-#ifdef CLOSE_THREAD_HANDLE
         CLOSE_THREAD_HANDLE( stream->processingThread ); /* Delete thread. */
-#endif
         stream->processingThread = NULL;
     }
 #endif
@@ -3092,10 +3090,8 @@ static PaError StopStream( PaStream *s )
         if( WaitForSingleObject( stream->processingThreadCompleted, 30*100 ) == WAIT_TIMEOUT )
             return paUnanticipatedHostError;
 
-#ifdef CLOSE_THREAD_HANDLE
-        CloseHandle( stream->processingThread ); /* Delete thread. */
+        CLOSE_THREAD_HANDLE( stream->processingThread ); /* Delete thread. */
         stream->processingThread = NULL;
-#endif
 
     }
 #endif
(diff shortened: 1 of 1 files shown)


philburk:
> This PR fixes a couple of issues 

Thanks for offering the fix. Could you create an Issue, or two, so that we can discuss the bug?

RossBencina:
I agree that the portaudio code is not consistent with the microsoft API docs at https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/endthread-endthreadex?view=msvc-170 I can see where the confusion comes from:

- beginthread/endthread  -> client MUST NOT call CloseHandle
- beginthreadex/endthreadex -> client MUST call CloseHandle

It also looks like the old code was not consistently using the macros supplied for creating and closing the thread (the macros exist for the purpose of cygwin compatibility when beginthreadex/endthreadex is not used). Ideally I think the macros should be retained and fixed rather than eliminating them, since they clearly abstract the differences between the two runtimes.

dechamps:
> Could you create an Issue, or two, so that we can discuss the bug?

Done in #925 and #926.

> Ideally I think the macros should be retained and fixed rather than eliminating them, since they clearly abstract the differences between the two runtimes.

Done by preserving the `CLOSE_THREAD_HANDLE` macro.

(I would argue that these macros don't really make sense in practice and we should get rid of them, because they are abstracting away differences that don't really exist - the official docs of `_beginthreadex()` do not even try to hide the fact that it's really just a thin wrapper around `CreateThread()`, and that the return value is really just a Win32 thread `HANDLE` in disguise, as evidenced by the fact that we're being told to pass it to `CloseHandle()`. Really the only difference is which function is used to create the thread - `CreateThread()` vs. `_beginthreadex()` and everything else is the same. Given this, the macros just seem to obscure the code for no apparent benefit. But I guess I don't have strong feelings about this.)

src/hostapi/dsound/pa_win_ds.c
```diff
@@ -3051,9 +3051,7 @@ static PaError StartStream( PaStream *s )
 #ifndef PA_WIN_DS_USE_WMME_TIMER
     if( stream->processingThread )
     {
-#ifdef CLOSE_THREAD_HANDLE
```
RossBencina, line 3054: The purpose of `#ifdef CLOSE_THREAD_HANDLE` is to allow the macro to be a no-op. Given that there is now no situation where it would be a no-op I am not sure that the macro should be present at all.

RossBencina (approved):
IMPORTANT: DO NOT SQUASH when merging.

I have finally made time to review this properly. Overall it looks fine, although it would have been easier to deal with if it was two separate PRs.

7398279cc34773a335addd937db45592b9319e99 (removal of event signaling thread completion) is 100% ok by me.

4a2fef468db74362feeccebf994c227e3e152eac (match `_beginthreadex` with `CloseThread`) appears correct to me. Not sure the fix is ideal, but ok.

src/hostapi/dsound/pa_win_ds.c
```diff
@@ -106,7 +106,7 @@
 
 #if !defined(__CYGWIN__) && !defined(UNDER_CE)
 #define CREATE_THREAD (HANDLE)_beginthreadex
-#undef CLOSE_THREAD_HANDLE /* as per documentation we don't call CloseHandle on a thread created with _beginthreadex */
+#define CLOSE_THREAD_HANDLE CloseHandle
```
RossBencina, line 109:
```suggestion
#define CLOSE_THREAD_HANDLE CloseHandle /* NOTE: _beginthreadex requires handle to be closed with CloseHandle */
```

RossBencina:
Merged. Thanks @dechamps

PR #943, opened 2024-07-12
undo .def file ordinal change made in 605fc89. …

### LINKED ISSUE #784: .def file ordinal change for MME symbols during dev cycle.
Since V19.7.0 was released the following changes were made:

MME symbols were introduced here Aug 19 2021:

https://github.com/PortAudio/portaudio/commit/093d7f7fdf2cd1dbed0cb888257b6b39b4bed7db

And then the ordinals were changed here Feb 7 2022:

https://github.com/PortAudio/portaudio/commit/605fc89d5763b3d7bfd4803531a23c359a668b73

I don't think that we should have allowed the second change. Once assigned, ordinals should remain stable.

To avoid issues when adding additional WASAPI symbols in future, I think we should renumber the wmme ordinals to start at 80 or 90 prior to the next release.

RossBencina:
add comments about append-only policy for ordinal changes. resolves #784

here's the change that it undoes: 605fc89

Phil and I have decided that we prefer this to #941

### DIFF
--- a/cmake/portaudio.def.in
+++ b/cmake/portaudio.def.in
@@ -39,6 +39,7 @@ Pa_GetStreamWriteAvailable          @32
 Pa_GetSampleSize                    @33
 Pa_Sleep                            @34
 Pa_GetVersionInfo                   @35
+; add new portable public API functions here. DO NOT CHANGE EXISTING ORDINALS!
 @DEF_EXCLUDE_ASIO_SYMBOLS@PaAsio_GetAvailableBufferSizes      @50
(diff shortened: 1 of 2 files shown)


philburk (approved):

RossBencina:
@dmitrykos ok with you?

RossBencina:
@dmitrykos awaiting your review

dmitrykos (approved):

dmitrykos:
> @dmitrykos awaiting your review

@RossBencina I am really sorry, missed your question! The change is ok.

Issue #946, opened 2024-07-18
Win32/WASAPI: suggestion: improve symbol names for the "stereo to mono mixer" code


RossBencina:
[Hi Dmitry, this is just a suggestion for changing internal function names. I'm not proposing to have a lengthy debate. Please close this issue if you don't want to action it]

During the review of #934 I had a bit of confusion about the meaning of the phrase "internal stereo to mono mixer" in a comment.

Upon review of the code in `pa_win_wasapi.c` I see that there is a mechanism for selecting functions for changing channel count: `GetMonoToStereoMixer`, it is called on line 3339:

```
        pSub->monoMixer = GetMonoToStereoMixer(&pSub->wavexu.ext, (output ? MIX_DIR__1TO2 : MIX_DIR__2TO1_L));
```

Note that this function, and the functions that it calls are used for 2->1 mixing or channel selection for input, or 1->2 channel duplication for output. There are two sources of confusion:

- In only one of 3 cases is this a "mixer". Otherwise it is a "selector" or a "duplicator". So I think it is better to avoid the terms "mix" and "mixer". A general term would be "converter".
- In the output case it is "mono to stereo" but for the input case it is "stereo to mono", however the text `MonoToStereo` is used throughout the code. This could be changed to use `MonoToStereo` and `StereoToMono` consistent with behavior, but since `2TO1` and `1TO2` also appear but might be possible to just avoid it in most cases. Another option is `MonoStereo` which does not name a direction.

I think it is good to aim to reduce the amount of confusion in the code base. Here are some options:

```
pSub->monoStereoConverter = GetMonoStereoConverter(&pSub->wavexu.ext, (output ? CONVERT__1TO2 : CONVERT_2TO1_L));
```
or

```
pSub->monoStereoConverter = output ? GetMonoToStereoConverter(&pSub->wavexu.ext) :  GetStereoToMonoConverter(&pSub->wavexu.ext, CONVERT__2TO1_L);
```

Then I would change the name of in individual converters:

`_StereoToMono_2TO1_8`, `_MonoToStereo_1TO2_24`

or, probably better:

`_Convert_2TO1_8`, `_Convert_1TO2_24`

Hope you see what I mean.

Thank you.

dmitrykos:
Hi Ross! I like your proposal, it will improve code readability indeed.

I would choose:

```c
[...]

PR #948, opened 2024-07-22
wasapi: Fixed Pa_ReadStream()/Pa_WriteStream() for 1-channel (Mono) I/O

### LINKED ISSUE #945: Win32/WASAPI loopback: distorted audio while capturing via sync API
Mono audio captured as `paInt16` on WASAPI loopback device via sync `Pa_ReadStream()` API is distorted, while the callback capture works without problem. It sounds like wrong format or clipped. Sampling rate and duration are correct. Capturing mic with the same program works fine.
Reproduced by @dmitrykos [here](https://github.com/PortAudio/portaudio/pull/934#issuecomment-2230143525).

dmitrykos:
- Fixes Pa_ReadStream()/Pa_WriteStream() for 1-channel (Mono) stream
- Made several examples configurable for 1-channel operation for easier testing of Mono I/O

The problem is covered by #945 issue. Modified examples can be used to test 1-channel input/output stream, including the loopback.

After the acceptance of this PR I will further address #946 issue in another PR and refactor mono-stereo mixer's implementation names.

Closes #945


arkadijs:
It's better but not completely free of cracking for some? devices.

Here are two examples:
http://0x0.st/Xp1-.wav - PortAudio master
http://0x0.st/Xp1o.wav - this PR

Each file has 4 segments of ~3sec:
1. Headphone callback
2. Headphones read
(small silence while msedge.exe switches to another output)
3. Speakers callback
4. Speakers read

While (2) is still obviously problematic it's much better now.
Test program [here](https://gist.github.com/arkadijs/7027bf8aad399f3420406600c73df9e4).

dmitrykos:
> While (2) is still obviously problematic it's much better now.

I found the problematic place, we were writing some trash/silence behind buffer when saving trailing frames into the circular buffer. Now all is fine. I used your test bench in my tests.

Would you please check from your side.

arkadijs:
Tested, it works, no problems. 🎉

dmitrykos:
That's great! I hope @RossBencina and @philburk will review this PR soon.

examples/paex_read_write_wire.c
```diff
@@ -50,10 +50,12 @@
 
 /* #define SAMPLE_RATE  (17932) // Test failure to open with this value. */
 #define SAMPLE_RATE       (44100)
-#define FRAMES_PER_BUFFER   (512)
-#define NUM_SECONDS          (10)
+#define CHANNELS          (2)
+#define FRAMES_PER_BUFFER (1024)
```
RossBencina, line 54: Why change `FRAMES_PER_BUFFER`?

examples/paex_read_write_wire.c
```diff
@@ -50,10 +50,12 @@
...
+#define NUM_SECONDS       (10)
 /* #define DITHER_FLAG     (paDitherOff)  */
-#define DITHER_FLAG           (0)
+#define DITHER_FLAG       (0)
+#define LISTEN_LOOPBACK   (0)
```
RossBencina, line 58: Better: `USE_LOOPBACK_INPUT_DEVICE`

src/hostapi/wasapi/pa_win_wasapi.c
```diff
@@ -1414,6 +1419,31 @@ static MixMonoToStereoF GetMonoToStereoMixer(const WAVEFORMATEXTENSIBLE *fmtext,
...
+    if (frames == 0)
+        frames = subStream->bufferSize;
+
+    assert(frames != 0);
+
+    // Expand buffer if necessary
+    UINT32 monoBufferSize = frames * subStream->wavexu.ext.Format.nBlockAlign;
+    if (monoBufferSize > subStream->monoBufferSize)
+    {
+        subStream->monoBuffer = PaWasapi_ReallocateMemory(subStream->monoBuffer, (subStream->monoBufferSize = monoBufferSize));
```
RossBencina, line 1437: assignment inside function call parameter expression seems unnecessarily hard to follow.

src/hostapi/wasapi/pa_win_wasapi.c
```diff
@@ -1414,6 +1419,31 @@ static MixMonoToStereoF GetMonoToStereoMixer(const WAVEFORMATEXTENSIBLE *fmtext,
     return NULL;
 }
 
+// ------------------------------------------------------------------------------------------
+static HRESULT TryAllocateMonoMixerBuffer(PaWasapiSubStream *subStream, UINT32 frames)
```
RossBencina, line 1423: I'm not sure what you're getting at with "TryAllocate" here. All allocations can fail. Maybe `ReallocateMonoMixerBuffer` or `ExpandMonoMixerBuffer`

src/hostapi/wasapi/pa_win_wasapi.c
```diff
@@ -3910,7 +3951,7 @@ static PaError OpenStream( struct PaUtilHostApiRepresentation *hostApi,
         if (stream->in.params.blocking == TRUE)
         {
             UINT32 bufferFrames = ALIGN_NEXT_POW2((stream->in.framesPerHostCallback / WASAPI_PACKETS_PER_INPUT_BUFFER) * 2);
-            UINT32 frameSize    = stream->in.wavexu.ext.Format.nBlockAlign;
+            UINT32 frameSize    = stream->in.wavexu.ext.Format.nBlockAlign / (stream->in.monoMixer != NULL ? 2 : 1);
```
RossBencina, line 3954:
```suggestion
            UINT32 frameSize    = (stream->in.monoMixer != NULL) ? (stream->in.wavexu.ext.Format.nBlockAlign / 2) : stream->in.wavexu.ext.Format.nBlockAlign;
```
More likely to get strength reduction to right shift this way.

src/hostapi/wasapi/pa_win_wasapi.c
```diff
@@ -4787,23 +4833,43 @@ static PaError ReadStream( PaStream* s, void *_buffer, unsigned long frames )
             continue;
         }
 
+        // Get buffer of the Mono to Stereo mixer, otherwise use WASAPI device buffer directly
+        if (isMonoStereoConverter)
+        {
+            // Expand buffer if necessary (normally buffer is preallocated by ActivateAudioClient())
```
RossBencina, line 4839: I get that it is normally preallocated, but this comment should also explain why it is sometimes necessary to expand the buffer here. Ideally you should never be calling an allocator inside ReadStream.

src/hostapi/wasapi/pa_win_wasapi.c
```diff
@@ -4905,20 +4972,41 @@ static PaError WriteStream( PaStream* s, const void *_buffer, unsigned long fram
...
         // Keep waiting again (on Vista it was noticed that WASAPI could SOMETIMES return NULL pointer
         // to buffer without returning AUDCLNT_E_BUFFER_TOO_LARGE instead)
-        if (wasapi_buffer == NULL)
+        if (deviceBuffer == NULL)
             continue;
 
+        // Get buffer of the Mono to Stereo mixer, otherwise use WASAPI device buffer directly
+        if (isMonoStereoConverter)
+        {
+            // Expand buffer if necessary (normally buffer is allocated by ActivateAudioClient())
```
RossBencina, line 4981: As for ReadStream: why?

src/hostapi/wasapi/pa_win_wasapi.c
```diff
@@ -6152,7 +6228,9 @@ PA_THREAD_FUNC ProcThreadPoll(void *param)
         return (UINT32)paUnanticipatedHostError;
 
     // Request fine (1 ms) granularity of the system timer functions to guarantee correct logic around WaitForSingleObject
-    SystemTimer_SetGranularity(&timer, 1);
+    // Note: while granularity is already set by OpenStream() we still must be sure that user code did not
+    // change granularity between Pa_OpenStream() and Pa_StartStream()
+    SystemTimer_SetGranularity(&stream->timer, 1);
```
RossBencina, line 6233: Doesn't this mean you are calling `timeBeginPeriod` twice, but `timeEndPeriod` only once?

> You must match each call to timeBeginPeriod with a call to timeEndPeriod, specifying the same minimum resolution in both calls

https://learn.microsoft.com/en-us/windows/win32/api/timeapi/nf-timeapi-timebeginperiod

RossBencina (commented):
I have made various comments. Also:

The patch updates examples/tests with compile-time options to select number of CHANNELS and to find and select the `[Loopback]` device for input.

    examples/paex_read_write_wire.c
    examples/paex_write_sine.c
    test/patest_read_record.c

Since the loopback device with this name is a WASAPI-specific thing I'm not sure whether it makes sense to update the generic examples with this feature. No need to change anything yet but this is something to discuss with Phil when he shows up.

examples/paex_read_write_wire.c
dmitrykos: Just typo, will revert.

examples/paex_read_write_wire.c
dmitrykos: I propose less verbose variant - `USE_LOOPBACK_INPUT­`

src/hostapi/wasapi/pa_win_wasapi.c
dmitrykos: Yes, but this prevent accidental misuse of unassigned `subStream->monoBufferSize` in `PaWasapi_ReallocateMemory`. I do not mind splitting it though as per your proposal.

src/hostapi/wasapi/pa_win_wasapi.c
dmitrykos: `TryXXX` was a leftover from its previous implementation when it was also checking `(subStream->monoMixer != NULL)`, I simplified that to avoid double checks but function name remained the same. Will change to `ReallocateXXX`.

src/hostapi/wasapi/pa_win_wasapi.c
dmitrykos: It depends on how smart compiler is, it could do right shift in my case too, if deducted, because the result of the comparison is constant (1 or 2). I would avoid doing this micro optimization because in my case the comparison is less bulky. Please let me know your opinion.

src/hostapi/wasapi/pa_win_wasapi.c
dmitrykos: I updated comments in places where `ReallocateMonoMixerBuffer` is called but prefer to leave those calls there because WASAPI is very driver dependent and they are more for a fail-safe operation.

src/hostapi/wasapi/pa_win_wasapi.c
dmitrykos: Would you please elaborate, I did not get what you mean.

src/hostapi/wasapi/pa_win_wasapi.c
dmitrykos: No, `SystemTimer_RestoreGranularity` is called in the end of each thread function/worker and also in `CloseStream` therefore all calls to `SystemTimer_SetGranularity` match `SystemTimer_RestoreGranularity`.

dmitrykos:
@RossBencina I updated implementation based on your review notes, please check.

examples/paex_read_write_wire.c
```diff
@@ -48,12 +48,14 @@
...
 
-/* #define SAMPLE_RATE  (17932) // Test failure to open with this value. */
-#define SAMPLE_RATE       (44100)
-#define FRAMES_PER_BUFFER   (512)
-#define NUM_SECONDS          (10)
-/* #define DITHER_FLAG     (paDitherOff)  */
-#define DITHER_FLAG           (0)
+/*#define SAMPLE_RATE        (17932) // Test failure to open with this value. */
+#define SAMPLE_RATE        (44100)
+#define CHANNELS           (2)
```
RossBencina, line 53:
```suggestion
#define CHANNELS           (2) /* set to 0 to use all device channels */
```

examples/paex_read_write_wire.c
```diff
@@ -120,6 +146,7 @@ int main(void)
 
     numChannels = inputInfo->maxInputChannels < outputInfo->maxOutputChannels
             ? inputInfo->maxInputChannels : outputInfo->maxOutputChannels;
+    numChannels = (CHANNELS ? CHANNELS : numChannels);
```
RossBencina, line 149: Are we sure we want to add this? Not all examples have "channels == 0  => use all channels" behavior

src/hostapi/wasapi/pa_win_wasapi.c
```diff
@@ -4787,23 +4835,44 @@ static PaError ReadStream( PaStream* s, void *_buffer, unsigned long frames )
             continue;
         }
 
+        // Get buffer of the Mono to Stereo mixer, otherwise use WASAPI device buffer directly
+        if (isMonoStereoConverter)
+        {
+            // Expand buffer if necessary (normally shall not happen, buffer is preallocated by
+            // ActivateAudioClient(), this is to circumvent unexpeced buffer size provided by the driver)
```
RossBencina, line 4842: fix typo:
```suggestion
            // ActivateAudioClient(), this is to circumvent unexpected buffer size provided by the driver)
```

Also, has this situation ever been observed? or is this just defensive programming just in case? Either way the comment should say, e.g. "which has been observed on some systems" or "never observed, but defensive measure, just in case."

src/hostapi/wasapi/pa_win_wasapi.c
RossBencina: I meant, "why expand the buffer size". the comment should explain.

src/hostapi/wasapi/pa_win_wasapi.c
RossBencina: Ok, I see that the timer Set/Restore calls are matched.

src/hostapi/wasapi/pa_win_wasapi.c
```diff
@@ -5849,7 +5930,9 @@ PA_THREAD_FUNC ProcThreadEvent(void *param)
         return (UINT32)paUnanticipatedHostError;
 
     // Request fine (1 ms) granularity of the system timer functions for precise operation of waitable timers
-    SystemTimer_SetGranularity(&timer, 1);
+    // Note: while granularity is already set by OpenStream() we still must be sure that user code did not
```
RossBencina, line 5933: This comment does not make sense to me. Once PA/WASAPI has reduced the timer granularity, it would be very difficult for client code to increase the granularity. Client code would have to issue an unmatched call to `timeEndPeriod` (already a bug in client code) with exactly the same granularity that PA/WASAPI used to set the granularity. Under these circumstances "we still must be sure" seems vague and confusing. The comment could say "while granularity has already been set by OpenStream(), we defensively set it again here to avoid issues with buggy client code." Then at least it is clear why you are doing it, but without further explanation I still see no likelihood of it being needed. It would be good for the comment to say why this could actually happen, especially if you have seen this in practice.

P.S. I checked and see that all of the Set/Restore calls are matched, so I resolved my earlier comment.

examples/paex_read_write_wire.c
```diff
@@ -105,6 +107,30 @@ int main(void)
     if( err != paNoError ) goto error2;
 
     inputParameters.device = Pa_GetDefaultInputDevice(); /* default input device */
+
+#if USE_LOOPBACK_INPUT
```
RossBencina, line 111: I don't think WASAPI-specific code belongs in our generic examples. Need to discuss with Phil.

RossBencina (commented):
Hi Dmitry, I've resolved a bunch of the old feedback and added some more.

Unfortunately Phil is absent today so I can't discuss some of the remaining issues with him.

This was very time consuming to re-review. You could help by:

1. Please do not force-push during the middle of a review. We need to be able to see what changes you made to resolve our review comments.

2. Please do not fold-in irrelevant changes such as whitespace and variable renaming. Especially when the PR is *big* like this one. I know that it is more work for you to keep things separate but it will make it much quicker for us to review the PRs.

Thank you!

examples/paex_read_write_wire.c
dmitrykos: Yes, but why not having couple of examples like these. The operation of example does not change but it is easier to reuse them for a development purpose.

src/hostapi/wasapi/pa_win_wasapi.c
dmitrykos: Yes, currently it is a defensive programming and I did not want to change the behavior of the existing implementation which was relying on buffer expansion in these places (with this PR buffer is preallocated inside `Initialize` stage). I propose to leave it as is and with another PR I can surround these blocks with `#ifdef` and make them off by default. If we do not receive any issues after that PR then those ifdefed blocks can be removed completely from the implementation.

src/hostapi/wasapi/pa_win_wasapi.c
dmitrykos: > Once PA/WASAPI has reduced the timer granularity, it would be very difficult for client code to increase the granularity.

My assumption was that usually `Pa_Initialize` is called somewhere early when app starts. Then user code can do some work and accidentally change granularity. In this case PA WASAPI is making sure that when stream is opened the granularity is on required level. In any case PA WASAPI is matching calls correctly, therefore I see no potential issue with a client app. This approach can be treated as defensive.

examples/paex_read_write_wire.c
dmitrykos: We could add comment for `USE_LOOPBACK_INPUT` that it affects WASAPI-only devices.

examples/paex_read_write_wire.c
```diff
@@ -105,6 +107,30 @@ int main(void)
...
+    for (i = Pa_GetDeviceCount() - 1; i >= 0; i--)
+    {
+        const PaDeviceInfo* device;
+        if ((device = Pa_GetDeviceInfo(i)) != NULL)
+        {
+            // Note: [Loopback] name postfix is provided by the WASAPI device only
+            if (strstr(device->name, "[Loopback]") != NULL)
```
philburk, line 118: 1) Host specific code should be avoided in generic examples.

2) Why would we want to use [Loopback] in code that copies input to output?
We end up with a closed loop: output->input->output->input...

src/hostapi/wasapi/pa_win_wasapi.c
RossBencina: To be clear: I don't think the code actually does anything because of the way windows timer API works: it will always choose the minimum of all requested granularities so it is very hard for the client to interfere with PortAudio's previous requests. So, I would like the code reviewed.

However, if you insist on retaining the code, then the comment needs to clearly explain why the code actually could have some effect. The comment right now is confusing because it contradicts the basic mechanism that the windows timer API uses.

examples/paex_read_write_wire.c
RossBencina: Phil and I agree that the generic examples should not include wasapi specific code paths. They are meant to be simple examples. So these changes need to be removed from the PR.

If there is good reason to include this code at all, it needs to go in separate files. Current practice would be to move theses file to `test/patest_read_write_wire_wasapi_loopback.c` etc.

examples/paex_write_sine.c
```diff
@@ -48,6 +48,7 @@
 
 #define NUM_SECONDS         (5)
 #define SAMPLE_RATE         (44100)
+#define CHANNELS            (2)
```
philburk, line 51: Add comment
/* Set to 1 or 2. */

philburk (changes requested):

examples/paex_read_write_wire.c
dmitrykos: I propose to have this test as it is very useful in checking WASAPI Loopback devices, so will create a new source file as proposed: `test/patest_read_write_wire_wasapi_loopback.c`

examples/paex_read_write_wire.c
dmitrykos: > 1. Host specific code should be avoided in generic examples.

Ok, I will create a separate file for WASAPI case.

> 2. Why would we want to use [Loopback] in code that copies input to output? We end up with a closed loop: output->input->output->input...

Not really, see the log below. There I play to USB DAC `Neutron HiFi DAC V1` and capture its stream via a Loopback device. Output goes to a 2-nd audio device which is `Speakers (Realtek(R) Audio)`. And it works just fine.

```
patest_read_write_wire.c
sizeof(int) = 4
sizeof(long) = 4
Input device # 5.
    Name: Speakers (2- Neutron HiFi DAC V1) [Loopback]
      LL: 0.003 s
      HL: 0.01 s
Output device # 0.
   Name: Speakers (Realtek(R) Audio)
     LL: 0.0025397 s
     HL: 0.0101587 s
Num channels = 2.
Wire on. Will run 10 seconds.
Wire off.
```

examples/paex_write_sine.c
dmitrykos: Done.

examples/paex_read_write_wire.c
dmitrykos: Done.

src/hostapi/wasapi/pa_win_wasapi.c
dmitrykos: @RossBencina you are right. I checked Microsoft docs for [timeBeginPeriod](https://learn.microsoft.com/en-us/windows/win32/api/timeapi/nf-timeapi-timebeginperiod) telling that _"For processes which call this function, Windows uses the lowest value (that is, highest resolution) requested by any process."_

I updated implementation and `SystemTimer_SetGranularity()` is called in once in `OpenStream()`.

dmitrykos:
@philburk I made all changes we discussed, would you please check and approve if all is fine now.

test/patest_read_record.c
```diff
@@ -47,13 +47,14 @@
...
-/* #define DITHER_FLAG     (paDitherOff)  */
-#define DITHER_FLAG     (0) /**/
+/* #define SAMPLE_RATE        (17932) // Test failure to open with this value. */
+#define SAMPLE_RATE        (44100)
+#define FRAMES_PER_BUFFER  (1024)
+#define NUM_SECONDS        (5)
+#define NUM_CHANNELS       (2)
+/* #define DITHER_FLAG        (paDitherOff)  */
+#define DITHER_FLAG        (0) /**/
+#define USE_LOOPBACK_INPUT (0)
```
RossBencina, line 57: Did you mean to include the changes to this file? If you want a test for wasapi loopback read_record, please move it to a wasapi-specific test like you did for read_write_wire_wasapi.

dmitrykos: Sorry, I did not notice. Reverted `patest_read_record` to original state, introduced `patest_read_record_wasapi` as separate file.

src/hostapi/wasapi/pa_win_wasapi.c
```diff
@@ -1414,6 +1419,33 @@ static MixMonoToStereoF GetMonoToStereoMixer(const WAVEFORMATEXTENSIBLE *fmtext,
...
+    if (frames == 0)
+        frames = subStream->bufferSize;
+
+    assert(frames != 0);
+
+    // Expand buffer if necessary
+    UINT32 monoBufferSize = frames * subStream->wavexu.ext.Format.nBlockAlign;
+    if (monoBufferSize > subStream->monoBufferSize)
+    {
+        subStream->monoBufferSize = monoBufferSize;
```
RossBencina, line 1437: Better to perform this assignment only once the reallocation succeeds

RossBencina (approved):
Hi Dmitry, I'm not going to hold up the merge for my latest request, more of a nice to have.

src/hostapi/wasapi/pa_win_wasapi.c
dmitrykos: No problem to adjust this implementation. Additionally I set `monoBufferSize` to 0 when memory allocation fails that looks logical because memory buffer gets freed in this case.

dmitrykos:
> Hi Dmitry, I'm not going to hold up the merge for my latest request, more of a nice to have.

Ross, I made proposed change, thank you for pointing to it.

examples/paex_read_write_wire.c
philburk: Oh! OK, so you are using two different outputs. That makes sense.

src/hostapi/wasapi/pa_win_wasapi.c
```diff
@@ -3910,7 +3956,7 @@ static PaError OpenStream( struct PaUtilHostApiRepresentation *hostApi,
         if (stream->in.params.blocking == TRUE)
         {
             UINT32 bufferFrames = ALIGN_NEXT_POW2((stream->in.framesPerHostCallback / WASAPI_PACKETS_PER_INPUT_BUFFER) * 2);
-            UINT32 frameSize    = stream->in.wavexu.ext.Format.nBlockAlign;
+            UINT32 frameSize    = stream->in.wavexu.ext.Format.nBlockAlign / (stream->in.monoMixer != NULL ? 2 : 1);
```
philburk, line 3959: Are we dividing the input bytes per frame by two to get the output bytes per frame because we are mixing from stereo to mono?

philburk (approved):
Looks great! Thanks for doing this.
Just a minor nit about an assert being a NOOP. OK to merge as is.

src/hostapi/wasapi/pa_win_wasapi.c
```diff
@@ -1414,6 +1419,36 @@ static MixMonoToStereoF GetMonoToStereoMixer(const WAVEFORMATEXTENSIBLE *fmtext,
...
+        if (subStream->monoBuffer == NULL)
+        {
+            subStream->monoBufferSize = 0;
+            return E_OUTOFMEMORY;
+        }
+
+        subStream->monoBufferSize = monoBufferSize;
+    }
+
+    assert(subStream->monoBuffer != NULL);
```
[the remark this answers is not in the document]
dmitrykos: It is a sanity check of the logic, in case buffer is not allocated at all due to `subStream->wavexu.ext.Format.nBlockAlign`. I modified assertion in the subsequent commit.

src/hostapi/wasapi/pa_win_wasapi.c
dmitrykos: Here ring buffer is serving user with mono data while WASAPI backend is operating with stereo data (`stream->in.wavexu.ext.Format.nBlockAlign`), therefore to avoid using 2x larger buffer size we shrink it to the actual size being served to the user.

philburk (approved):

arkadijs:
I've tested the version merged into master and it works for me, thank you!

Issue #957, opened 2024-08-15
CI: Test install and uninstall targets with ./configure


RossBencina:
Ideally CI should test the the install and uninstall targets work.

A minimal start would be to install + uninstall and require that there are no errors.

I think this involves extending `.github/workflows/autotools.yml` with the extra steps.

Next level would be to check that certain files are installed. e.g. by successfully compiling a little .c program that includes all of the expected includes. Testing uninstall could then check that the compilation fails after uninstall.

Raised by #956

PR #956, opened 2024-08-12
Add missing files to uninstall target

adsr:

PR #958, opened 2024-08-15
CI smoke test for autoconf make install/uninstall targets

RossBencina:
tests #956. wip for #957

### DIFF
--- a/.github/workflows/autotools.yml
+++ b/.github/workflows/autotools.yml
@@ -15,7 +15,7 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: configure
-      run: ./configure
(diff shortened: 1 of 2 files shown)


RossBencina:
This is working now. Ready to merge.

I've made it install to `~/.local` (actually, `/home/runner/.local` since the gihub CI runner user dir is `/home/runner`). It successfully compiles and runs a small test executable. It also tests that compilation fails after `make uninstall`.

illuusio:
Why manually build `patest_init`? Ain't it build with configure scripts?

philburk (approved):

RossBencina:
@illuusio the purpose of the compilation step is to test that `make install` actually installed working header files and library binaries.

Issue #939, opened 2024-07-07
Build portaudiocpp with cmake failed


donarturo11:
**Describe the bug**
Trying to create cmake build directory for portaudiocpp gets message:
```
CMake Error at cmake/modules/FindPortAudio.cmake:20 (message):
  PortAudio_FOUND but not target PortAudio::portaudio
```

**To Reproduce**
Steps to reproduce the behavior. Include code if applicable.
1. Install portaudio with cmake in any directory
```
cmake -S portaudio -B portaudio-build -DCMAKE_INSTALL_PREFIX=${PORTAUDIO_PATH}
cmake --build portaudio-build --config Release
cmake --install portaudio-build --config Release
```
2. Configure building with cmake portaudiocpp
```
cmake -S portaudio/bindings/cpp -B portaudiocpp-build -DCMAKE_PREFIX_PATH=${PORTAUDIO_PATH} 
```

**Expected behavior**
Configure build directory with no errors.

**Actual behavior**
CMake throws error
```
CMake Error at cmake/modules/FindPortAudio.cmake:20 (message):
  PortAudio_FOUND but not target PortAudio::portaudio
```

**Desktop (please complete the following information):**
 - OS: Windows
 - OS Version: 10
 - PortAudio version: current #[2fa0219](https://github.com/PortAudio/portaudio/commit/2fa021931944b141302697f0750c3e5484888726)
 - Host API : any

**Proposed solution**
Unify target names cases.

Note: PortAudio is a community supported project. If you have a solution, please create a Pull Request for us to consider.

RossBencina:
Our CI doesn't fail because our CI doesn't use portaudiocpp. But this does suggest a regression since presumably portaudiocpp used to build.

@donarturo11 would you please check PortAudio v19.7 release and confirm that it does not have this problem?

RossBencina:
It's not failing CI because portaudiocpp is not tested by CI.

My concern now is that there is a regression in the PortAudio library CMakeLists.txt, and that maybe we should undo the library library name case change in `PortAudio -> portaudio` in the main PortAudio  CMakeLists.txt.

@dechamps any thoughts?

donarturo11:
I think too that undoing change case in main CMakeList will be better solution than mine.

donarturo11:
> @donarturo11 would you please check PortAudio v19.7 release and confirm that it does not have this problem?

Yes sure. So, preparing and building portaudiocpp using source from main branch and stable version of portaudio are successful.
[...]

Issue #486, opened 2021-01-28
libportaudio_static.a is not created on Mac OS


ababo:
After [this commit](https://github.com/PortAudio/portaudio/commit/ea11061b20f637f8f707a4a225d37ff58ab54dc0) `portaudio_static` becomes broken (there's no way to build `libportaudio_static.a`).

ababo:
Basically this diff fixes the issue:
```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 5c5f748..4bcb22b 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -383,9 +383,9 @@ IF(PA_BUILD_STATIC)
   TARGET_INCLUDE_DIRECTORIES(portaudio_static PRIVATE ${PA_PRIVATE_INCLUDE_PATHS})
   TARGET_INCLUDE_DIRECTORIES(portaudio_static PUBLIC "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>" "$<INSTALL_INTERFACE:include>")
   TARGET_LINK_LIBRARIES(portaudio_static ${PA_LIBRARY_DEPENDENCIES})
-  IF(NOT PA_LIBNAME_ADD_SUFFIX)
-    SET_PROPERTY(TARGET portaudio_static PROPERTY OUTPUT_NAME portaudio)
-  ENDIF()
+  # IF(NOT PA_LIBNAME_ADD_SUFFIX)
+  #   SET_PROPERTY(TARGET portaudio_static PROPERTY OUTPUT_NAME portaudio)
+  # ENDIF()
 ENDIF()
 
 IF(WIN32 AND MSVC)
```

philburk:
@ababo - Thanks for the patch. 
I am not very familiar with CMake. So I don't understand your patch.
It looks like it is commenting out code related to "static". So it is the opposite of what I expected.

Can you explain briefly how this patch works?

ababo:
I just comment out these lines, because they prevent `libportaudio_static.a` from being created (if `PA_LIBNAME_ADD_SUFFIX` is not set, i.e. by default). I don't understand the intention of the author of these lines. Maybe he wanted to support a customisable suffix for the static library, but now be default `libportaudio_static.a` is being renamed  into `libportaudio.a`, which in turn is replaced buy the shared version of this library with the same name.

Be-ing:
[...]

PR #961, opened 2024-09-16
Fix cpp building cmake

rn3kk:
add LIBRARY_BUILD_TYPE to portaudiocpp

Need for fix static build as subproject (thrid-party)

### DIFF
--- a/bindings/cpp/CMakeLists.txt
+++ b/bindings/cpp/CMakeLists.txt
@@ -1,6 +1,12 @@
 cmake_minimum_required(VERSION 3.13)
 cmake_policy(VERSION 3.13)
 
+if(PA_BUILD_SHARED_LIBS)
+  set(LIBRARY_BUILD_TYPE SHARED)
+else()
+  set(LIBRARY_BUILD_TYPE STATIC)
+endif()
+
 project(PortAudioCpp VERSION 19.8 LANGUAGES CXX)
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake/modules")
 
@@ -65,7 +71,7 @@ endif()
 # portaudiocpp-targets
 # ##############################################################################
 
-add_library(portaudiocpp ${portaudiocpp-sources})
+add_library(portaudiocpp ${LIBRARY_BUILD_TYPE} ${portaudiocpp-sources})
 add_library(PortAudio::portaudiocpp ALIAS portaudiocpp) # For subdirectory build
 
 find_package(PortAudio MODULE REQUIRED)


philburk:
@rn3kk - Thanks for this patch. Since you are building the CPP binding, could you please take a look at #939, and a possible fix in #963, and make a comment.

philburk:
I am not a CMake expert. But would it be better to just build BOTH libraries using something like:

add_library(portaudiocpp_shared SHARED ${portaudiocpp-sources})
add_library(portaudiocpp_static STATIC ${portaudiocpp-sources})

RossBencina:
I am not a CMake expert, but I think we need to go back to building both static and shared libs in the main PortAudio cmakelists. It is a regression that this functionality was removed, see

#486 

and other related tickets:


https://github.com/PortAudio/portaudio/issues?q=is%3Aissue+is%3Aopen+label%3Abuild-cmake

RossBencina:
This issue needs to be fixed but we're not currently convinced that this is the correct fix. Hence assigning low priority to this PR.

It would be good if someone can create an issue to track the actual bug that this PR is supposed to fix (i.e. the specific problem building the C++ binding).

RossBencina:
Removing from milestone.  No response for 6 minths. We have no maintainers for Cmake. We have not received any help in understanding whether this is the correct fix. Happy to contine the discussion, but this can't hold up releases.

dmitrykos:
@RossBencina I can handle this question but in a different way. I already checked how C++ bindings are built and currently it is overly complicated - building C++ lib is detached from building of the main PA library. I can't build C++ bindings easily on Windows with MSVC currently as `find_package` package fails. Current implementation is probably ok for Linux but not for generic Windows building experience.

I propose to simplify building C++ bindings by making them a byproduct of building the PA library. In this way CMake interface will provide an option to build C++ buildings `PA_BUILD_CPP_BINDINGS`. This approach will greatly simplify the CMake code for building C++ bindings and it will be more portable.

If you are ok with this approach I will develop a PR for further discussion and integration. Please let me know your oppinion.

jcelerier:
I think it should be the responsibility of the user to set BUILD_SHARED_LIBS or not depending on their needs ; as a cmake user it's actually fairly frustrating when a library tries to build both; Portaudio is small and builds fast so it's ok, but some other larger libraries or frameworks basically double their minutes-long build times for something that is going to be half-discarded anyways. It also means the user needs to pollute their own cmake code with complicated logic such as:

```cmake
if(we are buildling portaudio in-tree)
  add_subdirectory(...)
  if(we want shared portaudio)
    set(portaudio_target_name portaudio_shared)
  else()
    set(portaudio_target_name portaudio_static)
  endif()
elseif(we are using prebuilt portaudio from distro)
  if(distro builds portaudio with autotools and does not ship portaudioconfig.cmake)  
    if(we want shared portaudio)
     find_library(portaudio_shared ...) # or portaudio depending on the packaging, I've seen all possible variations
    else()
     find_library(portaudio_static ...)
    endif()
    # create a target manually depending on if it's called portau
  elseif(distro ships portaudioconfig.cmake)
    find_package(portaudio)
    if(we want static portaudio and the distro provides static portaudio)
      set(portaudio_target_name PortAudio::portaudio_static)
    else if(we want shared portaudio and the distro provides shared portaudio)
      set(portaudio_target_name PortAudio::portaudio_shared)
    else(...)
      etc...
   endif()
endif()
```

@rn3kk  if you want static build in a subproject you can just do

```cmake
block()
  set(BUILD_SHARED_LIBS OFF)
  add_subdirectory(path/to/portaudio)
endblock()
```

PR #965, opened 2024-10-04
pulseaudio: Rework stopping code that it should prevent artifacts

illuusio:
Rework stopping and stream closing code as it should prevent artifacts when closing or stopping stream. Correctly handle stream termination from callback


src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
```diff
@@ -426,6 +434,13 @@ static int _PaPulseAudio_ProcessAudio(PaPulseAudio_Stream *stream,
 
         PaUtil_EndCpuLoadMeasurement( &stream->cpuLoadMeasurer,
                                       hostFrameCount );
+
+        if( ret )
+        {
+            stream->isActive = 0;
+            pulseaudioLength = length - pulseaudioOutputWritten;
+            goto outputzero;
```
philburk, line 442: I find the use of the goto to jump from inside an infinite loop to the inside of previous conditional be be a bit confusing.

Could you put the code on lines 299 to 319 into a subroutine and then call it in two places.
I think that would be easier to read.

philburk (changes requested):

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
illuusio: You are absolutely correct I'll fix that. I'm sometime too fancy with goto that I use it without noting readability.

illuusio:
This could take a look now that other one is fixed.

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
RossBencina, line 1: Github won't let me add a comment to an unchanged line, but at line 429 call to `pa_steam_write`, nothing has been copied to `bufferData` yet. Furthermore, `bufferData` memory is invalidated by the call (see https://freedesktop.org/software/pulseaudio/doxygen/stream_8h.html#a4fc69dec0cc202fcc174125dc88dada7 ) so subsequent calls to the buffer processor may crash.

I suggest moving the calls to `PaUtil_EndBufferProcessing` and `PaUtil_EndCpuLoadMeasurement` prior to the call to `pa_stream_write`.

RossBencina:
@illuusio we're awaiting your response to latest review.

illuusio:
> @illuusio we're awaiting your response to latest review.

Sorry been bit busy but now it should be fixed

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
illuusio: Hopefully I understood correctly what you meant with this but now it should be like that.

illuusio:
Ok this PR seems this one needs bit more work I separated that output thing to PR#1005. Let's get that first in so I can solve last bits in this.

RossBencina:
This PR contains a whole bunch of unrelated changes, mostly comment edits. Please reduce this to a minimal change against master that does what it claims to do: "rework stopping code to prevent artifacts."

illuusio:
It was rebased against on wrong branch now should be again master and the should not be anymore extra stuff. I don't really get why there is these extra stuff which does not belong there as they don't show in commit or anything.

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
```diff
@@ -425,7 +436,12 @@ static int _PaPulseAudio_ProcessAudio(PaPulseAudio_Stream *stream,
             size_t tmpSize = pulseaudioOutputBytes;
 
             /* Allocate memory to make it faster to output stuff */
-            pa_stream_begin_write( stream->outputStream, &bufferData, &tmpSize );
+            if( pa_stream_begin_write( stream->outputStream, &bufferData, &tmpSize ) )
+            {
+                PA_DEBUG( ("Portaudio %s: Can't output to stream!\n",
+                __FUNCTION__) )
+                return paNotInitialized;
```
RossBencina: Is there a reason to use `paNotInitialized` here but `paInsufficientMemory` in the other case of checking `pa_stream_begin_write`? I think either they should use the same error code, or there should be a comment explaining what's going on.

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
```diff
@@ -233,6 +233,37 @@ void _PaPulseAudio_Read( PaPulseAudio_Stream *stream,
...
+    size_t tmpSize = length;
+    int ret = paContinue;
+    void *bufferData = NULL;
+
+    if( length <= 0)
+    {
+        return ret;
+    }
+
+    /* Allocate memory to make it faster to output stuff */
```
```suggestion
    /* Prepare memory to make it faster to output stuff */
```

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
```diff
@@ -233,6 +233,37 @@ void _PaPulseAudio_Read( PaPulseAudio_Stream *stream,
...
+    if( pa_stream_begin_write( stream->outputStream, &bufferData, &tmpSize ) )
+    {
+        PA_DEBUG( ("Portaudio %s: Can't output to stream!\n",
+        __FUNCTION__) );
+        return paInsufficientMemory;
```
```suggestion
        return paUnanticipatedHostError;
```

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
```diff
@@ -425,7 +436,12 @@ static int _PaPulseAudio_ProcessAudio(PaPulseAudio_Stream *stream,
             size_t tmpSize = pulseaudioOutputBytes;
 
             /* Allocate memory to make it faster to output stuff */
```
```suggestion
            /* Prepare memory to make it faster to output stuff */
```

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
```suggestion
                return paUnanticipatedHostError;
```

RossBencina (changes requested):
I'll commit these suggestions now.

PR #969, opened 2024-10-19
Pulseaudio: Termination stream before operation release can lead locking

illuusio:
Cases when Pulseaudio server releases stream before operation releasing can be lead locking in pa_threaded_mainloop_wait forever.

Test if operation is running before running pa_threaded_mainloop_wait which should prevent forever waiting releasing lock.


src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
```diff
@@ -119,12 +119,15 @@ void PaPulseAudio_ReleaseOperation(PaPulseAudio_HostApiRepresentation *hostapi,
 
     while( waitOperation > 0 )
     {
-        PaPulseAudio_Lock( hostapi->mainloop );
-        pa_threaded_mainloop_wait( hostapi->mainloop );
-        PaPulseAudio_UnLock( hostapi->mainloop );
-
         localOperationState = pa_operation_get_state( localOperation );
```
kleinerm, line 122: This is still racy, because the pulseaudio server could release the stream in the time span between pa_operation_get_state() and pa_threaded_mainloop_wait() and you'd be back to square one with a hang.

Instead I'd do...

```
PaPulseAudio_Lock( hostapi->mainloop );
localOperationState = pa_operation_get_state( localOperation );
if( localOperationState == PA_OPERATION_RUNNING )
{
    pa_threaded_mainloop_wait( hostapi->mainloop );
}
 PaPulseAudio_UnLock( hostapi->mainloop );
```

... to avoid this race.

illuusio: Thank you taking look at this. This will fix this situation for good.

kleinerm: As a little optimization, I'd also call `localOperationState = pa_operation_get_state( localOperation );` again, immediately after the `pa_threaded_mainloop_wait( hostapi->mainloop );` so that the following
`if( localOperationState != PA_OPERATION_RUNNING)` will decide if it should break; out of the while loop, based on the latest localOperationsState.

This way you could skip a potentially not needed usleep(1000); and avoid a needless 1 millisecond delay when this function is called, which seems to be from various places.

illuusio: Ok I've made bit optimization. Removed delay as it sleep in semaphore in `pa_threaded_mainloop_wait` if there comes signals very quick they should come that quick they smack down CPU.

illuusio:
@kleinerm could you approve this so I can merge it to repo..

kleinerm:
@illuusio  Looks good to me, approved.

kleinerm (approved):

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
```diff
@@ -119,21 +119,23 @@ void PaPulseAudio_ReleaseOperation(PaPulseAudio_HostApiRepresentation *hostapi,
...
+        {
+            pa_threaded_mainloop_wait( hostapi->mainloop );
+        }
+        else
         {
+            // Result is DONE or CANCEL
+            PaPulseAudio_UnLock( hostapi->mainloop );
             break;
         }
+        PaPulseAudio_UnLock( hostapi->mainloop );
```
RossBencina, line 136: Suggestion: the control flow would be easier to follow if this UnLock operation was moved inside the `if(){}` branch

RossBencina: I'm not sure whether you can do anything about it, but note that the following still has a race condition:

```
        localOperationState = pa_operation_get_state( localOperation );
        if( localOperationState == PA_OPERATION_RUNNING )
        {
            pa_threaded_mainloop_wait( hostapi->mainloop );
```

illuusio: Sorry I don't see the race condition here a we are in locked state? Could you explain little bit that I can fix it or if it can't then it's just racing.

RossBencina: Sorry, I just checked the documentation for `pa_threaded_mainloop_wait` here https://www.freedesktop.org/software/pulseaudio/doxygen/thread-mainloop_8h.html#ac96d19260567d0406350f1b3a14c4bed

It is clear that the lock is released-reacquired by main-loop wait. So presumably there is no race between checking the stream state and waiting for it to change.

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
```diff
@@ -119,21 +119,23 @@ void PaPulseAudio_ReleaseOperation(PaPulseAudio_HostApiRepresentation *hostapi,
...
-        PaPulseAudio_UnLock( hostapi->mainloop );
 
+        PaPulseAudio_Lock( hostapi->mainloop );
         localOperationState = pa_operation_get_state( localOperation );
 
-        // No wait if operation have been DONE or CANCELLED
-        if( localOperationState != PA_OPERATION_RUNNING)
+        if( localOperationState == PA_OPERATION_RUNNING )
+        {
+            pa_threaded_mainloop_wait( hostapi->mainloop );
```
philburk, line 128: It looks scary to calla  wait() function under a lock because of the danger of deadlocks.
But this follows the PulseAudio example code. So it seems OK.

RossBencina, line 128: "This function may not be called inside the event loop thread. "

See docs for `pa_threaded_mainloop_wait` here: https://www.freedesktop.org/software/pulseaudio/doxygen/thread-mainloop_8h.html#ac96d19260567d0406350f1b3a14c4bed

Please could you confirm that this is not called from the main thread? `PaPulseAudio_UnLock`/`Lock` explicitly checks ` !pa_threaded_mainloop_in_thread( mainloop )`

illuusio: It shouldn't but I'll check. It's used on Portaudio signals and Pulseaudio callback but I should be check. Thank you for reading more carefully the docs :+1: . I don't know should it be replaces with just delay as this is taken from example.

illuusio: Ok I've added small commit that print PA_DEBUG print if we are on mainloop thread and it doesn't print so I can surely say that in test cases (this function is called something like 5 times in Portaudio life span)  it won't happen.

RossBencina (approved):

philburk (approved):

Issue #970, opened 2024-10-21
Allow to open stream with "current" sample rate


Spacechild1:
Sometimes we just want to open a stream with its "current" sample rate. This is particularly true for backends where the sample rate is fixed or has been set by another application. Prominent examples are Jack and ASIO.

Now, you might thank that we could simply use `PaDeviceInfo.sampleRate` for this purpose. Unfortunately, this does not always work.

While Jack gets the default sample rate directly from the Jack server, ASIO follows a rather peculiar strategy:

1. Iterate over an arbitrary list of sample rates and ask the driver if the rate is supported.
2. Pick the first rate that works.

Typically, this always results in a default sample rate of 44100 Hz, as this rate is pretty much always supported.

Now here's the problem: Just because ASIO tells you that the rate is supported does *not* mean that you can actually use it for opening a new stream!

Modern ASIO drivers typically support multiple client, but they all must use the same sample rate. If another ASIO client has already opened a stream at, say, 48000 Hz, the device has been set to that rate. If you then attempt to open a new stream using the "default sample rate" as reported by PortAudio, you would get an error!

IMO a "default sample rate" should be a sample rate that is guaranteed to work. In the case of ASIO that would be the current driver sample rate (as reported by `ASIOGetSampleRate`).

---

Originally, I thought that we could simple set `PaDeviceInfo.defaultSampleRate` to the current ASIO device sample rate and be done.

However, there's a more fundamental issue: since `PaDeviceInfo` is immutable, `defaultSampleRate` is chosen *at the time of library initialization*. However, the ASIO/Jack sample rate may change anytime!

I think the best solution would be an option that tells `Pa_OpenStream` to use the *actual* current sample rate. This could be a new stream flag like `paUseCurrentSampleRate`. After the stream has been opened, the user could query the actual sample rate with `Pa_GetStreamInfo`.

Any thoughts?

dechamps:
> However, there's a more fundamental issue: since `PaDeviceInfo` is immutable, `defaultSampleRate` is chosen _at the time of library initialization_. However, the ASIO/Jack sample rate may change anytime!

This is just a specific instance of the wider problem of the PortAudio API not having a way to report changes to device properties. The same problem applies to reporting addition and removal of devices, for example.

You could work around the issue by completely re-initializing PortAudio every time you want to "refresh" the device list, though obviously this is not exactly a cheap operation.

Spacechild1:
[...]

PR #971, opened 2024-10-21
ASIO: add `PaAsio_GetSampleRate()` function

Spacechild1:
`PaAsio_GetSampleRate()` returns the current sample rate of the specified device at the time of library initialization.
(If the current sample rate is unknown, a sample rate of 0 will be returned.)

For example, this is useful for people who want to use the current sample rate as the default sample rate (which PA's ASIO implementation does not do.)

Actually, I would rather prefer to change the definition of "defaultSampleRate" in the context of ASIO, see https://github.com/PortAudio/portaudio/issues/970, but if this is not desired, this PR would be an acceptable workaround.

### DIFF
--- a/examples/pa_devs.c
+++ b/examples/pa_devs.c
@@ -167,7 +167,11 @@ int main(void)
         {   /* Use wide char on windows, so we can show UTF-8 encoded device names */
             wchar_t wideName[MAX_PATH];
             MultiByteToWideChar(CP_UTF8, 0, deviceInfo->name, -1, wideName, MAX_PATH-1);
+        #if __USE_MINGW_ANSI_STDIO
+            wprintf( L"Name                        = %S\n", wideName );
+        #else
             wprintf( L"Name                        = %s\n", wideName );
+        #endif
         }
 #else
         printf( "Name                        = %s\n", deviceInfo->name );
@@ -186,6 +190,7 @@ int main(void)
 /* ASIO specific latency information */
         if( Pa_GetHostApiInfo( deviceInfo->hostApi )->type == paASIO ){
             long minLatency, maxLatency, preferredLatency, granularity;
+            double currentSampleRate;
 
             err = PaAsio_GetAvailableLatencyValues( i,
                     &minLatency, &maxLatency, &preferredLatency, &granularity );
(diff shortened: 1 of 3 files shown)


RossBencina:
Phil and I discussed this PR at length today. I think that an argument can be made for providing a way to query the current ASIO device sample rate irrespective of whether it's the optimal fix for #970. We already have `PaAsio_SetStreamSampleRate`.

A problem with this PR in its current form is that it does not actually query the current device sample rate. It just returns the cached value retrieved at `Pa_Initialize` time. Thus if the sample rate is changed asynchronously (e.g a change to a front panel switch, or incoming wordclock) the proposed implementation will return an incorrect value. Thus this PR is not really a perfect fix for #970.

Of course querying the current device sample rate is actually a massive pain (see `PaAsio_ShowControlPanel` for the implementation and limitations) which comes with its own caveats; and we have used the cached approach for other query functions (e.g, `PaAsio_GetInputChannelName`). So I am hesitant to request a rewrite to this patch.

Spacechild1:
> Thus this PR is not really a perfect fix for https://github.com/PortAudio/portaudio/issues/970.

I agree. In the linked issue #970 I said that I'd actually prefer to add a flag to `Pa_OpenStream` and you seemed to agree: https://github.com/PortAudio/portaudio/issues/970#issuecomment-2442903056. Do you want me to implement that?

dyfer:
@RossBencina would you consider @Spacechild1 's offer to add a flag to `Pa_OpenStream`? This would be very much a welcome addition to the ASIO backend.

dyfer:
@RossBencina ping

RossBencina:
I'm not going to approve a patch that I havn't seen, but yes, a host-api-specific flag (i.e. a flag in the host-api-specific parameters structure) to open the stream at the current ASIO device sample rate would be a reasonable candidate for a  host-API-specific extension and I can't right now think of any reason that would cause problems. If the patch is written in such a way that it is very clear that there is no change in behavior when the flag is not set, that will make it easier to make the merge decision.

Spacechild1:
Thanks for the answer! Just wanted to check if you agree with the basic idea before I make a PR.

> I'm not going to approve a patch that I havn't seen

Nobody asked for that :)

Issue #331, opened 2020-10-16
ASIO : crash in ASIO SDK function deleteDrvStruct 


yannickModulo:
Hi,

[EDIT: the following refers to the ASIOSDK file `$ASIOSDK\host\pc\asiolist.cpp`]

We have a very strange crash in debug when calling Pa_Terminate(). There is a crash in 👍 static void deleteDrvStruct (LPASIODRVSTRUCT lpdrv)
{
	IASIO	*iasio;

	if (lpdrv != 0) {
		deleteDrvStruct(lpdrv->next);
		if (lpdrv->asiodrv) {
			iasio = (IASIO *)lpdrv->asiodrv;
			iasio->Release();
		}
		delete lpdrv;   <- THIS LINE !!
	}
}

in this thread https://portaudio.music.columbia.narkive.com/5oYjtv3x/crash-on-asio-de-init, someone indicate to change delete lpdrv;  by in delete[] lpdrv; 

Do we need to do this in ASIO SDK ? strange that there is a bug.

RossBencina:
According to the followup comment https://portaudio.music.columbia.narkive.com/5oYjtv3x/crash-on-asio-de-init#post2

"""
Reinhold Hoffmann 4 years ago
[Permalink](https://narkive.com/5oYjtv3x.2)

Hi Louis-Paul,

According to

https://lists.columbia.edu/pipermail/portaudio/2016-November/000946.html

there is a regression bug in the ASIO SDK 2.3 which is open for quite some
time.

Instead

delete lpdrv; // crash here

the correct code line is

delete[] lpdrv; <====================

Hope this fixes your issue. Please let us know.

Reinhold
"""

RossBencina:
It seems that it is indeed an array, see the following line in `newDrvStruct()`:

> lpdrv = new ASIODRVSTRUCT[1];

so `delete []` is required.

umlaeute:
as of today it seems that steinberg still ships a broken version ASIO-SDK per default.
have they been informed of the issue? (i know that i could somehow file a bug myself, but i'm not exactly on good speaking terms...;-))

PR #997, opened 2025-01-24
cmake: patch asiolist.cpp as issue #331 explains

sonoro1234:
If download of ASIO api is complete, this performs the modification of "delete lpdrv" to "delete[] lpdrv".
Even in the case it was already done that would be harmless.

### DIFF
--- a/cmake/modules/FindASIO.cmake
+++ b/cmake/modules/FindASIO.cmake
@@ -66,7 +66,7 @@ if(ASIO_ROOT)
       file(READ "${ASIO_ROOT}/host/pc/asiolist.cpp" FILE_CONTENTS)
       string(REPLACE "delete lpdrv" "delete[] lpdrv" FILE_CONTENTS "${FILE_CONTENTS}")
       file(WRITE "${ASIO_ROOT}/host/pc/patched_asiolist.cpp" "${FILE_CONTENTS}")
-      message(STATUS "Done patching asiolist.cpp")
+      message(STATUS "Done patching asiolist.cpp into patched_asiolist.cpp")
     endif()
     add_library(ASIO::host INTERFACE IMPORTED)
     target_sources(ASIO::host INTERFACE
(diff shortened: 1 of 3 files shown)


RossBencina:
Thanks for this. This is a good thing to fix. It would be better to not modify the user's files in place.

As a suggestion, perhaps we could create the patched version of  `asiolist.cpp` as a separate temporary file, somewhere within the build space, as we already do for the generated `.def` file. The generated/patched file could be named `patched_asioroot_host_pc_asiolist.cpp` and this file could be used by the build step instead of `asiolist.cpp` from the sdk.

sonoro1234:
Done

cmake/modules/FindASIO.cmake
```diff
@@ -61,11 +61,18 @@ if(ASIO_ROOT)
   message(STATUS "Found ASIO SDK: ${ASIO_ROOT}")
 
   if(ASIO_FOUND AND NOT TARGET ASIO::host)
+    #patch asiolist.cpp
+    if(NOT EXISTS "${ASIO_ROOT}/host/pc/patched_asiolist.cpp")
+      file(READ "${ASIO_ROOT}/host/pc/asiolist.cpp" FILE_CONTENTS)
+      string(REPLACE "delete lpdrv" "delete[] lpdrv" FILE_CONTENTS "${FILE_CONTENTS}")
+      file(WRITE "${ASIO_ROOT}/host/pc/patched_asiolist.cpp" "${FILE_CONTENTS}")
+      message(STATUS "Done patching asiolist.cpp")
```
RossBencina, line 69:
```suggestion
      message(STATUS "Done patching asiolist.cpp into patched_asiolist.cpp")
```

philburk (approved):

RossBencina (approved):
Due to severity of the issue and the effectiveness of the fix, we're going to merge this as-is. However it could be improved further, as noted by new issue #1004

PR #1000, opened 2025-02-01
Fix bad channel count in patest_wire.c

philburk:
Now it uses the max channel count for the default devices.

For #595


dmitrykos (approved):

RossBencina (approved):
Looks fine. Please fix whitespace before merging.

PR #1005, opened 2025-02-14
pulseaudio: Correct output writing place where it's copied

illuusio:
Output copying what in incorrect place now that should be corrected. This is related: #965

### DIFF
--- a/src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
+++ b/src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
@@ -438,8 +438,8 @@ static int _PaPulseAudio_ProcessAudio(PaPulseAudio_Stream *stream,
         }
 
         hostFrameCount =
-            PaUtil_EndBufferProcessing( &stream->bufferProcessor,
-                                        &ret );
+                PaUtil_EndBufferProcessing( &stream->bufferProcessor,
+                                            &ret );
 
         PaUtil_EndCpuLoadMeasurement( &stream->cpuLoadMeasurer,
                                       hostFrameCount );
--- a/src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
+++ b/src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
@@ -309,8 +309,20 @@ static int _PaPulseAudio_ProcessAudio(PaPulseAudio_Stream *stream,
 
     if( !stream->isActive && stream->pulseaudioIsActive && stream->outputStream)
     {
-        bufferData = pulseaudioSampleBuffer;
-        memset( bufferData, 0x00, length);
+        size_t tmpSize = length;
+
+        /* Allocate memory to make it faster to output stuff */
+        pa_stream_begin_write( stream->outputStream, &bufferData, &tmpSize );
+
+        /* If bufferData is NULL then output is not ready
+         * and we have to wait for it
+         */
+        if(!bufferData)
+        {
+            return paNotInitialized;
+        }
+
+        memset( bufferData, 0x00, tmpSize);
 
         pa_stream_write( stream->outputStream,
                          bufferData,
@@ -323,8 +335,16 @@ static int _PaPulseAudio_ProcessAudio(PaPulseAudio_Stream *stream,
     }
 
 
-    while(1)
+    do
     {
+        /* Make sure that bufferData is NULL that marks there
+         * is nothing to output
+         */
+        if( isOutputCb )
+        {
+            bufferData = NULL;
+        }
+
         /*
          * If everything fail like stream vanish or mainloop
          * is in error state try to handle error
@@ -396,6 +416,9 @@ static int _PaPulseAudio_ProcessAudio(PaPulseAudio_Stream *stream,
                                        hostFramesPerBuffer );
         }
 
+        /* Is output is enable and buffer data is not NULL
+         * then crap pointer to output ringbuffer
+         */
         if( isOutputCb )
         {
 
@@ -403,14 +426,6 @@ static int _PaPulseAudio_ProcessAudio(PaPulseAudio_Stream *stream,
 
             /* Allocate memory to make it faster to output stuff */
             pa_stream_begin_write( stream->outputStream, &bufferData, &tmpSize );
-
-            /* If bufferData is NULL then output is not ready
-             * and we have to wait for it
-             */
-            if(!bufferData)
-            {
-                return paNotInitialized;
-            }
 
             PaUtil_SetInterleavedOutputChannels( &stream->bufferProcessor,
                                                  0,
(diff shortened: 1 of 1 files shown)


illuusio:
Should this been reviewed. It's mainly just moving code. @kleinerm do you have time to test this?

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
```diff
@@ -309,8 +309,20 @@ static int _PaPulseAudio_ProcessAudio(PaPulseAudio_Stream *stream,
...
+        size_t tmpSize = length;
+
+        /* Allocate memory to make it faster to output stuff */
+        pa_stream_begin_write( stream->outputStream, &bufferData, &tmpSize );
+
+        /* If bufferData is NULL then output is not ready
+         * and we have to wait for it
+         */
+        if(!bufferData)
+        {
```
RossBencina, line 321: Should you be calling `pa_stream_cancel_write` here?

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
```diff
@@ -309,8 +309,20 @@ static int _PaPulseAudio_ProcessAudio(PaPulseAudio_Stream *stream,
...
+            return paNotInitialized;
```
RossBencina, line 322: Using `paNotInitialized` here is a bit unusual but I guess it's OK if you're checking for this value in the caller.

RossBencina:
Looks ok modulo my comments. Let's wait another week for @kleinerm to show up (or not).

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
RossBencina: Never mind, we read the documentation too :)

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
```diff
@@ -420,6 +440,26 @@ static int _PaPulseAudio_ProcessAudio(PaPulseAudio_Stream *stream,
...
 
+        }
+
+        hostFrameCount =
+            PaUtil_EndBufferProcessing( &stream->bufferProcessor,
+                                        &ret );
+
+        PaUtil_EndCpuLoadMeasurement( &stream->cpuLoadMeasurer,
+                                      hostFrameCount );
+
```
RossBencina, line 451: I think you should move the `return paNotInitialized` statement here, so that you exit with the buffer processing in a known good state. Obviously this is non-trivial to do properly.

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
illuusio: What would be better. It's just error for caller but I also thought it's no good but could not figure out what should be used. Should it be just -1 or something for error.. too fancy.

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
philburk: I agree. Returning from the middle of a begin/end block seems wrong. Best to cleanup before returning.

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
RossBencina: Yes, I think -1 would be better. Or define your own internal error codes that capture the actual meaning of the error conditions (or use pulse audio error codes if that is feasible. I notice that you never check the return codes.

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
```diff
@@ -420,6 +440,26 @@ static int _PaPulseAudio_ProcessAudio(PaPulseAudio_Stream *stream,
...
+        /*
+         * Idea behind this is that is there is output we have
```
philburk, line 453: The beginning of this paragraph is unclear.

RossBencina, line 453: Not sure whether this is what you meant:

```
         * The idea behind this is that if we have allocated
         * an output buffer with pa_stream_begin_write
         * and then if PaUtil_EndBufferProcessing returns
         * paContinue we write data to the buffer, but if
         * there is no further output (ret is something
         * other than paContinue) then we have to cancel
         * the write with pa_stream_cancel_write
```

But I don't think this behavior is correct. Please check other host APIs. I think you should be calling pa_stream_write whenever hostFrameCount > 0.

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
```diff
@@ -433,15 +473,20 @@ static int _PaPulseAudio_ProcessAudio(PaPulseAudio_Stream *stream,
...
+            bufferData = NULL;
+        }
 
-        hostFrameCount =
-            PaUtil_EndBufferProcessing( &stream->bufferProcessor,
-                                        &ret );
-
-        PaUtil_EndCpuLoadMeasurement( &stream->cpuLoadMeasurer,
-                                      hostFrameCount );
+        if( ret != paContinue )
```
philburk, line 482: This break could be removed and the while loop could be use:

    do
    {
     .....
    } while( ret == paContinue );

philburk (changes requested):

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
illuusio: ok this one should be now corrected

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
illuusio: I checked ALSA and it's done the same. It does not even check `hostFrameCount` only if return in something else than `paContinue`

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
illuusio: Ok this should be now fixed

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
illuusio: This was one good idea an I incorporated that

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
illuusio: Tried to change wording to be more exact

illuusio:
Is this ok to merge?

illuusio:
Could this be updated @RossBencina  or @philburk so I can start working on other pr?

philburk (approved):

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
```diff
@@ -420,6 +435,31 @@ static int _PaPulseAudio_ProcessAudio(PaPulseAudio_Stream *stream,
             PaUtil_SetOutputFrameCount( &stream->bufferProcessor,
                                         hostFramesPerBuffer );
 
+        }
+
+        hostFrameCount =
+            PaUtil_EndBufferProcessing( &stream->bufferProcessor,
```
```suggestion
                PaUtil_EndBufferProcessing( &stream->bufferProcessor,
```

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
```diff
@@ -420,6 +435,31 @@ static int _PaPulseAudio_ProcessAudio(PaPulseAudio_Stream *stream,
...
+                                        &ret );
```
```suggestion
                                            &ret );
```

RossBencina (commented):
Fix indent. I will merge

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
```diff
@@ -396,6 +416,9 @@ static int _PaPulseAudio_ProcessAudio(PaPulseAudio_Stream *stream,
                                        hostFramesPerBuffer );
         }
 
+        /* Is output is enable and buffer data is not NULL
```
philburk, line 419: This code does not check that "buffer data is not null".

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
```diff
@@ -396,6 +416,9 @@ static int _PaPulseAudio_ProcessAudio(PaPulseAudio_Stream *stream,
...
+         * then crap pointer to output ringbuffer
```
philburk, line 420: What is a "crap pointer"?

RossBencina:
Your code comments are difficult to understand for native English speakers. Maybe try using an AI to help with grammar. We're not really happy with the comments but will merge anyway so that you can move forward.

illuusio:
> Your code comments are difficult to understand for native English speakers. Maybe try using an AI to help with grammar. We're not really happy with the comments but will merge anyway so that you can move forward.

Sorry about that. I'll use AI in future.

PR #1012, opened 2025-03-13
patest_wire.c: improvement: validate channel counts

RossBencina:
explicitly require at least one input and one output channel.


philburk (approved):
Good idea.

RossBencina:
Force pushed rebase onto master to pick up CI config fix

PR #1013, opened 2025-03-24
Increase required CMake version to 3.10

gruenich:
Newer version of CMake warn that versions older than 3.10 are deprecated. CMake 4 will be released soon and it refuses to accept a minimum required version below 3.5 without additional flags.


RossBencina:
> Set project language explicitly

`src/hostapi/asio/pa_asio.cpp` is C++ do we need to add something extra for that?

gruenich:
You are right, I missed the C++ part. I updated my second commit.

CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
-cmake_minimum_required(VERSION 3.1.0)
-project(PortAudio VERSION 19.8)
+cmake_minimum_required(VERSION 3.10)
+project(PortAudio VERSION 19.8 LANGUAGES C CXX)
```
dmitrykos, line 2: `LANGUAGES` is optional and C and CXX are enabled by default by CMake `By default C and CXX are enabled if no language options are given.` (https://cmake.org/cmake/help/latest/command/project.html), therefore specifying them explicitly is not needed.

dmitrykos (changes requested):

CMakeLists.txt
gruenich: Just for C the change made sense. Now not more. Thanks for pointing this out, I dropped the second commit.

dmitrykos (approved):

dmitrykos:
Requesting minimum CMake 3.10 will limit to Ubuntu 18.04 and thus will affect Docker builds. Currently it is CMake 3.1.0 which is supported by Ubuntu 16.04 by default.

- Ubuntu 14.04 (Trusty Tahr): CMake 2.8.12
- Ubuntu 16.04 (Xenial Xerus): CMake 3.5.1
---- 
- Ubuntu 18.04 (Bionic Beaver): CMake 3.10
- Ubuntu 20.04 (Focal Fossa): CMake 3.16
- Ubuntu 22.04 (Jammy Jellyfish): CMake 3.22
- Ubuntu 24.04 (Noble Numbat): CMake 3.28.3
- Ubuntu 24.10 (Oracular Oriole): CMake 3.30.3

More comments from other PortAudio users needed regarding this change as it will drop support for Ubuntu 16.04 builds with default tools.

gruenich:
Ready to be merge. Unless you want to reduce the minimum required version to 3.5. This will trigger a deprecation warning. It is your decision.

jpr42:
FWIW picking 3.10 as the min would close #1016 where as picking 3.5 as the minimum wouldn't

peter-toft-greve:
Great - super annoying for cmake 4. Looking forward to see this merged

martinblech:
I would humbly suggest making the much safer 3.5 min version upgrade first to fix the build in CMake >=4 without dropping support for any popular Ubuntu versions in docker builds out there (as pointed out by https://github.com/PortAudio/portaudio/pull/1013#issuecomment-2750280070). The 3.10 min version upgrade should be addressed separately as it will take longer to verify that this can be done safely.

peter-toft-greve:
Ubuntu 16.04 has EOL April 30, 2021. Given the amount of security updates since I would strongly recommend to upgrade.
Keep 16.04 support should not be a priority

gruenich:
Please either merge this or give a clear feedback that you would prefer a CMake minimum version of 3.5.

dmitrykos:
I hope @RossBencina and @philburk comment on this PR. I am personally ok with targeting 3.10.

philburk:
I am not a CMake expert. But I tend to agree with the suggestion from @martinblech to upgrade first to 3.5 to address the immediate compatibility issues. Then, maybe a week later, to upgrade to a minimum CMake of 3.10 so that we are more current.

Then if someone needs the 3.5 CMake for Ubuntu 16.04 they can use the older commit.

@RossBencina - what do you think?

gruenich:
I created !1019 to bump CMake to 3.5. Would be great to have it merged. Getting a minor release would be the cherry on top, because than packages do not have to apply the patch manually.

I will rebase this merge request once !1019 is merged.

RossBencina:
I agree to first merge #1019 (update to CMake 3.5).

It's not 100% clear to me whether we should merge the 3.10 patch before or after the PA V19.8 release (the one we are trying to close off on). Is the Ubuntu 16.04 breakage the only downside of moving to 3.10? Would this actually affect anyone that we know of?

While I am here, there are five open CMake related issues blocking the 19.8 release: https://github.com/PortAudio/portaudio/milestone/10 neither Phil nor I have the requisite experience to deal with CMake, and we have plenty to keep us busy elsewhere.  We have been asking for help with the CMake build system for literally years now. At this point you guys are taking the piss. You're happy to lobby for CMake support and jump to action when things actually break, but no one is doing the maintenance work. If the situation does not improve, sooner or later I'm going to snap and delete CMake build support from the tree. `</rant>`

dmitrykos:
> Is the Ubuntu 16.04 breakage the only downside of moving to 3.10?

Yes. For an older system, developer could alter `cmake_minimum_required(VERSION 3.5)` manually in the project. I propose to merge this PR and wait for any further issues regarding this change.

philburk:
This will have to be rebased because we merged #1019.

gruenich:
Rebased, ready to be merged. I don't mind if you make a release before merging this.

gruenich:
Any reason to not merge this?

philburk:
@gruenich - Thanks for rebasing. It's nice to know that it passes the current CI tests.

@RossBencina - what do you think about merging this now?

RossBencina:
We're going to merge this. It seems like CMake is on a constant upgrade cycle and people using CMake expect projects to stay compatible with the latest CMake. So there's no reason to lag too far behind.

PR #1018, opened 2025-04-03
Update cmake_minimum_required to 3.5

martinblech:
Fixes build error with CMake 4 and above:

```
CMake Error at CMakeLists.txt:7 (CMAKE_MINIMUM_REQUIRED):
  Compatibility with CMake < 3.5 has been removed from CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.

  Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.1.0)
+cmake_minimum_required(VERSION 3.5)
 project(PortAudio VERSION 19.8)
 
 #


martinblech:
Never mind, I see that this is superseded by the already approved https://github.com/PortAudio/portaudio/pull/1013

Issue #1016, opened 2025-03-30
Policy CMP0069 is not set


jpr42:
I currently get the following warning from portaudio

```
CMake Warning (dev):
  Policy CMP0069 is not set: INTERPROCEDURAL_OPTIMIZATION is enforced when
  enabled.  Run "cmake --help-policy CMP0069" for policy details.  Use the
  cmake_policy command to set the policy and suppress this warning.       

  INTERPROCEDURAL_OPTIMIZATION property will be ignored for target 'glfw'.
This warning is for project developers.  Use -Wno-dev to suppress it. 
```

This happens when I try and set CMAKE_INTERPROCEDURAL_OPTIMIZATION for all targets in my project including portaudio.

jpr42:
Fixed by updating minimum to CMake 3.10

PR #1019, opened 2025-04-05
[cmake] Increase minimum required version of CMake to 3.5

gruenich:
CMake 4 refuses to accept a minimum required version below 3.5 without additional flags.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.1.0)
+cmake_minimum_required(VERSION 3.5)
 project(PortAudio VERSION 19.8)
 
 #


mortie:
I'm hitting this as well, and it'll be a bigger and bigger problem as more and more systems start distributing cmake 4 by default. This MR fixes the issue.

Although the PortAudio project may want to consider whether it *actually* supports CMake 3.5, or if a higher number would be more appropriate.

philburk:
@RossBencina - this seems like a reasonable step towards the evantual 3.10 patch. What do you think?

RossBencina:
This PR (#1019) moves from CMake 3.1.0 -> 3.5, which is the minimum required version that CMake 4.0 will accept. Note that this PR still causes a deprecation warning, therefore does NOT address #1016.

With a view towards Docker images, using 3.5 maintains compatibility with Ubuntu LTS 16.04 (Xenial Xerus), which ships with CMake 3.5.1. Note that Ubuntu 14.04 shipped with 2.8.12 so we don't care about that.

For reference, PR #1013 is to require CMake 3.10 and there is more discussion there.

RossBencina (approved):
Seems like the appropriate next step.

philburk (approved):

PR #1021, opened 2025-04-29
Fix some minor Cppcheck warnings

gruenich:


test/patest_two_rates.c
```diff
@@ -82,7 +82,7 @@ static int patestCallback( const void *inputBuffer, void *outputBuffer,
     for( frameIndex=0; frameIndex<(int)framesPerBuffer; frameIndex++ )
     {
         /* Generate sine wave. */
-        float value = (float) 0.3 * sin(data->phase);
+        float value = 0.3f * sinf(data->phase);
```
RossBencina, line 85: `sinf` is C99, but we already call it in three other tests so maybe ok

philburk (approved):

test/patest_two_rates.c
gruenich: I can create a pull request to a) check in the build system that at least C99 is used and b) document C99 as the minimum required C standard.

philburk:
Ready to merge after updating the commit message.

gruenich:
@philburk What changes do you have in mind?

test/patest_two_rates.c
RossBencina: FYI: C89 is the minimum required C standard, that is not in doubt.

RossBencina:
I'm going to merge most of this with edited commit messages as @1022. Phil could you please approve it?

PR #1022, opened 2025-05-15
small fixups from PR #1021 with improved commit messages

RossBencina:
These are the commits from PR #1021 that have already been approved by Phil and I. I just updated the commit messages to be clearer.

I dropped the patch changing `sin` -> `sinf` since that is not C89. I will revert the other uses of sinf.


RossBencina:
@philburk please approve. we already reviewed this. only the commit messages have changed.

philburk (approved):

PR #1025, opened 2025-05-27
WIP fix SR handling

philburk:
Just a PR so I can see the changes easier.
DO NOT MERGE


src/hostapi/alsa/pa_linux_alsa.c
```diff
@@ -2750,17 +2757,20 @@ static PaError PaAlsaStream_Configure( PaAlsaStream *self, const PaStreamParamet
...
+        ENSURE_( GetExactSampleRate( hwParamsCapture, &realSr ), paUnanticipatedHostError );
         PA_DEBUG(( "%s: Capture period size: %lu, latency: %f\n", __FUNCTION__, self->capture.framesPerPeriod, *inputLatency ));
     }
     if( self->playback.pcm )
     {
         assert( self->playback.framesPerPeriod != 0 );
         PA_ENSURE( PaAlsaStreamComponent_FinishConfigure( &self->playback, hwParamsPlayback, outParams, self->primeBuffers, realSr,
                     outputLatency ) );
+        /* Now that we have finalized the hwParams, we can get a more accurate sample rate. */
+        ENSURE_( GetExactSampleRate( hwParamsPlayback, &realSr ), paUnanticipatedHostError );
```
philburk, line 2770: What should we do if the exact input and output sample rates are not the same?
We should log a warning if abs(diff)>tolerance.

Then use the average?

RossBencina (commented):
In function `PaAlsaStream_Configure` the variable `realSr` is used prior to calling `GetExactSampleRate`. I think `realSr` should be replaced by two variables: `approximateSampleRate` and `exactSampleRate`. What are currently uses of `realSr` prior to the `GetExactSampleRate` calls should use `approximateSampleRate`.

src/hostapi/alsa/pa_linux_alsa.c
RossBencina: I agree that we should log a warning. It can go into this PR if you like, or not.

I don't think we should change any other behavior in this PR such as using the average.

philburk:
OK, I will log a warning if the same rates do not match,
and rename the variables.

src/hostapi/alsa/pa_linux_alsa.c
```diff
@@ -2735,39 +2742,59 @@ static PaError PaAlsaStream_Configure( PaAlsaStream *self, const PaStreamParamet
         PaUtilHostBufferSizeMode* hostBufferSizeMode )
 {
     PaError result = paNoError;
-    double realSr = sampleRate;
+    double approximateSampleRate = sampleRate;
+    double exactSampleRate = sampleRate;
```
philburk: Use "precise" instead of "exact" in variable and GetExactSampleRate().

src/hostapi/alsa/pa_linux_alsa.c
```diff
@@ -2735,39 +2742,59 @@ static PaError PaAlsaStream_Configure( PaAlsaStream *self, const PaStreamParamet
...
                 hwParamsCapture, hwParamsPlayback, hostBufferSizeMode ) );
 
     if( self->capture.pcm )
     {
         assert( self->capture.framesPerPeriod != 0 );
-        PA_ENSURE( PaAlsaStreamComponent_FinishConfigure( &self->capture, hwParamsCapture, inParams, self->primeBuffers, realSr,
+        PA_ENSURE( PaAlsaStreamComponent_FinishConfigure( &self->capture, hwParamsCapture, inParams, self->primeBuffers, approximateSampleRate,
                     inputLatency ) );
+        /* Now that we have finalized the hwParams, we can get a more accurate sample rate. */
+        exactCaptureSampleRate = approximateSampleRate;
```
RossBencina: set to 0

philburk: Set to zero to make it clear that this is an output parameter in GetExactSampleRate().

src/hostapi/alsa/pa_linux_alsa.c
```diff
@@ -2735,39 +2742,59 @@ static PaError PaAlsaStream_Configure( PaAlsaStream *self, const PaStreamParamet
...
+        exactSampleRate = exactCaptureSampleRate;
     }
     if( self->playback.pcm )
     {
         assert( self->playback.framesPerPeriod != 0 );
-        PA_ENSURE( PaAlsaStreamComponent_FinishConfigure( &self->playback, hwParamsPlayback, outParams, self->primeBuffers, realSr,
+        PA_ENSURE( PaAlsaStreamComponent_FinishConfigure( &self->playback, hwParamsPlayback, outParams, self->primeBuffers, approximateSampleRate,
                     outputLatency ) );
+        /* Now that we have finalized the hwParams, we can get a more accurate sample rate. */
+        exactPlaybackSampleRate = approximateSampleRate;
```
RossBencina: set to 0

src/hostapi/alsa/pa_linux_alsa.c
```diff
@@ -2735,39 +2742,59 @@ static PaError PaAlsaStream_Configure( PaAlsaStream *self, const PaStreamParamet
...
+    double exactPlaybackSampleRate = sampleRate;
     snd_pcm_hw_params_t* hwParamsCapture, * hwParamsPlayback;
 
     alsa_snd_pcm_hw_params_alloca( &hwParamsCapture );
     alsa_snd_pcm_hw_params_alloca( &hwParamsPlayback );
 
     if( self->capture.pcm )
         PA_ENSURE( PaAlsaStreamComponent_InitialConfigure( &self->capture, inParams, self->primeBuffers, hwParamsCapture,
-                    &realSr ) );
+                    &approximateSampleRate ) );
```
philburk, line 2753: Declare approximateSampleRate as uint and pass by value to _InitialConfigure().

src/hostapi/alsa/pa_linux_alsa.c
```diff
@@ -2735,39 +2742,59 @@ static PaError PaAlsaStream_Configure( PaAlsaStream *self, const PaStreamParamet
         PaUtilHostBufferSizeMode* hostBufferSizeMode )
 {
     PaError result = paNoError;
-    double realSr = sampleRate;
+    double approximateSampleRate = sampleRate;
```
RossBencina: make `approximateSampleRate` a uint

src/hostapi/alsa/pa_linux_alsa.c
RossBencina, line 2753: pass `approximateSampleRate` by value

src/hostapi/alsa/pa_linux_alsa.c
```diff
@@ -852,9 +853,9 @@ static PaError GropeDevice( snd_pcm_t* pcm, int isPlug, StreamDirection mode, in
     unsigned int minChans = 0;
     unsigned int maxChans = 0;
     int* minChannels, * maxChannels;
-    double * defaultLowLatency, * defaultHighLatency, * defaultSampleRate =
-        &devInfo->baseDeviceInfo.defaultSampleRate;
-    double defaultSr = *defaultSampleRate;
+    double * defaultLowLatency, * defaultHighLatency;
+    double defaultSr = devInfo->baseDeviceInfo.defaultSampleRate;
+    unsigned int uintSampleRate = 0;
```
philburk: Bad name. Maybe "approximateSampleRate".

src/hostapi/alsa/pa_linux_alsa.c
```diff
@@ -2735,39 +2742,59 @@ static PaError PaAlsaStream_Configure( PaAlsaStream *self, const PaStreamParamet
...
     alsa_snd_pcm_hw_params_alloca( &hwParamsPlayback );
 
     if( self->capture.pcm )
         PA_ENSURE( PaAlsaStreamComponent_InitialConfigure( &self->capture, inParams, self->primeBuffers, hwParamsCapture,
-                    &realSr ) );
+                    &approximateSampleRate ) );
     if( self->playback.pcm )
         PA_ENSURE( PaAlsaStreamComponent_InitialConfigure( &self->playback, outParams, self->primeBuffers, hwParamsPlayback,
-                    &realSr ) );
+                    &approximateSampleRate ) );
```
RossBencina, line 2756: pass `approximateSampleRate` by value

src/hostapi/alsa/pa_linux_alsa.c
philburk: Cannot pass by value to _InitialConfigure because it is updated by calling SetApproximateSampleRate().

src/hostapi/alsa/pa_linux_alsa.c
philburk: Done

RossBencina (approved):
Approved. This definitely addresses my previous concern about making the approximate/precise sample rate distinction clear in the variable names. I'm still wondering whether the distinction between ALSA `uint` sample rates and PortAudio `double` sample rates could be made clear, since right now `sampleRate` is often used for both. But let's not further delay merging this.

PR #1027, opened 2025-05-28
loopback: check if format supported before testing

### LINKED ISSUE #1026: loopback test tries to check unsupported sample rates
The loopback test is checking formats that are not supported.

**To Reproduce**
1. Connect a Zoom AMS-22 to a Raspberry Pi 5.
1. make loopback
1. bin/paloopback

**Expected behavior**
Report that 8000, 11025 and other low sample rates are not supported.

**Actual behavior**
Lots of error messages spewing all over.

 - OS: Raspbian
 - OS Version [e.g. 22] 
 - PortAudio version: master

philburk:
Note that this PR is on top of #1024 (fix signed char)

I will rebase this when #1024 is merged.

Fixes #1026

### DIFF
--- a/qa/loopback/src/paqa.c
+++ b/qa/loopback/src/paqa.c
@@ -673,7 +673,6 @@ int PaQa_RunLoopback( LoopbackContext *loopbackContext )
     PaError err = 0;
     TestParameters *test = loopbackContext->test;
 
-
     if( test->flags & PAQA_FLAG_TWO_STREAMS )
     {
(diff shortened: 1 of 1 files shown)


RossBencina (commented):
I am ok with the change, it's obviously an improvement and if you'd like to merge as-is I'm fine with that. However I think it might be better to do the "is supported" checks separately so that the calling code can deal with unsupported formats separately from other errors.

RossBencina (approved):
Ok to merge as is. But check comment for suggestions.

Issue #112, opened 2011-02-01
require clipping for dithering sample conversion functions?


PortAudio-admin:
*Issue created by @RossBencina*

Consider whether functions which dither but don't clip should exist, V18 automatically enabled clipping whenever dithering was selected. Perhaps we should do the same.

This would have the advantage that it would eliminate many of the unimplemented converters in ticket #35

philburk:
If we dither then that could cause valid in-range data passed by the app to go out-of-range.
So, yes, I think if we dither then we should always clip.

PR #1029, opened 2025-05-28
Implement 8-bit converters with Dither

### LINKED ISSUE #168: paUInt8 not working when recording
*Issue created by @philburk*

I get scrambled data when recording with paUInt8 mode on Mac.  This was detected using the paloopback test.  Note the bad paUInt8 test result at the bottom of the attached report.

dmitrykos:
Implements converters:

- Int32_To_UInt8_Dither
- Int24_To_UInt8_Dither
- Int16_To_UInt8_Dither
- Int16_To_Int8_Dither

Implementation is validated with `patest_converters`:

```c
== flags = paClipOff ==

=== Output contains non-zero data ===
Key: . - pass, X - fail
{{{
in|  out:      f32     i32     i24     i16      i8     ui8
f32             .       .       .       .       .       .
i32             .       .       X       .       .       .
i24             .       .       .       .       .       .
i16             .       .       .       .       .       .
 i8             .       .       .       .       .       .
ui8             .       .       .       .       .       .
}}}

=== Combined dynamic range (src->dest->float32) ===
Key: Noise amplitude in dBfs, X - fail (either above failed or dest->float32 failed)
{{{
in|  out:      f32     i32     i24     i16      i8     ui8
 f32          -inf  -180.6  -138.5   -79.4   -31.9   -31.3
 i32        -144.5  -144.5     X     -85.9   -42.1   -42.1
 i24        -138.5  -138.5  -138.5   -85.5   -42.1   -42.1
 i16         -84.8   -84.8   -84.8   -84.8   -42.1   -42.1
  i8         -36.6   -36.6   -36.6   -36.6   -36.6   -36.6
 ui8         -36.4   -36.4   -36.4   -36.4   -36.4   -36.4
}}}
```

Closes #168

### DIFF
--- a/src/common/pa_converters.c
+++ b/src/common/pa_converters.c
@@ -1012,16 +1012,18 @@ static void Int32_To_UInt8_Dither(
     void *sourceBuffer, signed int sourceStride,
     unsigned int count, struct PaUtilTriangularDitherGenerator *ditherGenerator )
 {
-    /* PaInt32 *src = (PaInt32*)sourceBuffer;
-    unsigned char *dest =  (unsigned char*)destinationBuffer; */
-    (void)ditherGenerator; /* unused parameter */
+    PaInt32 *src = (PaInt32*)sourceBuffer;
+    unsigned char *dest = (unsigned char*)destinationBuffer;
+
+    PaInt32 dither;
 
     while( count-- )
     {
-        /* IMPLEMENT ME */
+        dither = PaUtil_Generate16BitTriangularDither( ditherGenerator );
+        *dest = (unsigned char) ((((src[0] >> 1) + dither) >> 23) + 128);
 
-        /* src += sourceStride;
-        dest += destinationStride; */
+        src += sourceStride;
+        dest += destinationStride;
     }
 }
 
@@ -1260,13 +1262,30 @@ static void Int24_To_UInt8_Dither(
     void *sourceBuffer, signed int sourceStride,
     unsigned int count, struct PaUtilTriangularDitherGenerator *ditherGenerator )
 {
-    (void) destinationBuffer; /* unused parameters */
-    (void) destinationStride; /* unused parameters */
-    (void) sourceBuffer; /* unused parameters */
-    (void) sourceStride; /* unused parameters */
-    (void) count; /* unused parameters */
-    (void) ditherGenerator; /* unused parameters */
-    /* IMPLEMENT ME */
+    unsigned char *src = (unsigned char*)sourceBuffer;
+    unsigned char  *dest = (unsigned char*)destinationBuffer;
+
+    PaInt32 temp, dither;
+
+    while( count-- )
+    {
+
+#if defined(PA_LITTLE_ENDIAN)
+        temp = (((PaInt32)src[0]) << 8);
+        temp = temp | (((PaInt32)src[1]) << 16);
+        temp = temp | (((PaInt32)src[2]) << 24);
+#elif defined(PA_BIG_ENDIAN)
+        temp = (((PaInt32)src[0]) << 24);
+        temp = temp | (((PaInt32)src[1]) << 16);
+        temp = temp | (((PaInt32)src[2]) << 8);
+#endif
+
+        dither = PaUtil_Generate16BitTriangularDither( ditherGenerator );
+        *dest = (unsigned char) ((((temp >> 1) + dither) >> 23) + 128);
+
+        src += sourceStride * 3;
+        dest += destinationStride;
+    }
 }
 
 /* -------------------------------------------------------------------------- */
@@ -1373,16 +1392,20 @@ static void Int16_To_Int8_Dither(
     void *sourceBuffer, signed int sourceStride,
     unsigned int count, struct PaUtilTriangularDitherGenerator *ditherGenerator )
 {
-    /* PaInt16 *src = (PaInt16*)sourceBuffer;
-    signed char *dest =  (signed char*)destinationBuffer; */
-    (void)ditherGenerator; /* unused parameter */
+    PaInt16 *src = (PaInt16*)sourceBuffer;
+    signed char *dest = (signed char*)destinationBuffer;
+
+    PaInt32 temp, dither;
 
     while( count-- )
     {
-        /* IMPLEMENT ME */
+        temp = ((PaInt32)src[0]) << 16;
 
-        /* src += sourceStride;
-        dest += destinationStride; */
+        dither = PaUtil_Generate16BitTriangularDither( ditherGenerator );
+        *dest = (signed char) (((temp >> 1) + dither) >> 23);
+
+        src += sourceStride;
+        dest += destinationStride;
     }
 }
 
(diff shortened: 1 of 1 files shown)


philburk:
Thanks for doing this. I was fixing some other issues with the loopback test on Raspberry Pi and noticed this was still broken.

src/common/pa_converters.c
```diff
@@ -1260,13 +1262,30 @@ static void Int24_To_UInt8_Dither(
...
 {
-    (void) destinationBuffer; /* unused parameters */
-    (void) destinationStride; /* unused parameters */
-    (void) sourceBuffer; /* unused parameters */
-    (void) sourceStride; /* unused parameters */
-    (void) count; /* unused parameters */
-    (void) ditherGenerator; /* unused parameters */
-    /* IMPLEMENT ME */
+    unsigned char *src = (unsigned char*)sourceBuffer;
+    unsigned char  *dest = (unsigned char*)destinationBuffer;
```
philburk, line 1266: extra space before *

src/common/pa_converters.c
```diff
@@ -1260,13 +1262,30 @@ static void Int24_To_UInt8_Dither(
...
+        temp = temp | (((PaInt32)src[1]) << 16);
+        temp = temp | (((PaInt32)src[2]) << 24);
+#elif defined(PA_BIG_ENDIAN)
+        temp = (((PaInt32)src[0]) << 24);
+        temp = temp | (((PaInt32)src[1]) << 16);
+        temp = temp | (((PaInt32)src[2]) << 8);
+#endif
+
+        dither = PaUtil_Generate16BitTriangularDither( ditherGenerator );
+        *dest = (unsigned char) ((((temp >> 1) + dither) >> 23) + 128);
```
philburk, line 1284: The dither needs to applied just below the low bit position of the destination.
So this might need to be:

    *dest = (unsigned char) ((((temp >> 9) + dither) >> 15) + 128);

src/common/pa_converters.c
```diff
@@ -1373,16 +1392,20 @@ static void Int16_To_Int8_Dither(
...
-    (void)ditherGenerator; /* unused parameter */
+    PaInt16 *src = (PaInt16*)sourceBuffer;
+    signed char *dest = (signed char*)destinationBuffer;
+
+    PaInt32 temp, dither;
 
     while( count-- )
     {
-        /* IMPLEMENT ME */
+        temp = ((PaInt32)src[0]) << 16;
```
philburk, line 1402: To align the dither correctly I think we may need:

        temp = ((PaInt32)src[0]) << 7;
        dither = PaUtil_Generate16BitTriangularDither( ditherGenerator );
        *dest = (signed char) ((temp + dither) >> 15);

src/common/pa_converters.c
```diff
@@ -1413,16 +1436,20 @@ static void Int16_To_UInt8_Dither(
...
-    (void)ditherGenerator; /* unused parameter */
+    PaInt16 *src = (PaInt16*)sourceBuffer;
+    unsigned char *dest = (unsigned char*)destinationBuffer;
+
+    PaInt32 temp, dither;
 
     while( count-- )
     {
-        /* IMPLEMENT ME */
+        temp = ((PaInt32)src[0]) << 16;
```
philburk, line 1446: Same comment as for Int8 shifting.

philburk (changes requested):
I notice that we are not clipping after we apply the dither.
But if the input is maximal then adding dither can easily lead to numeric overflow when downcasting.
Maybe we should always PA_CLIP as an integer before casting into the destination.

philburk:
@dmitrykos wrote:
> Implementation is validated with patest_converters:

How did you compile it? When I add it to the Makefile I get an error saying it cannot find PaUtil_SelectConverter!

dmitrykos:
> How did you compile it? When I add it to the Makefile I get an error saying it cannot find PaUtil_SelectConverter!

PaUtil_SelectConverter is local function of pa_converters.c. I compiled with VS configured by CMake, PA is static.

src/common/pa_converters.c
dmitrykos: so you propose to truncate first to int16 then allocate space for a dither and then truncate to int8

philburk: I am OK with:

    dither = PaUtil_Generate16BitTriangularDither( ditherGenerator ) << 8;
    *dest = (unsigned char) ((((temp >> 1) + dither) >> 23) + 128);

I think for dither to make sense it needs to be added at the 0.5 bit position.
Yes, there will be more apparent noise. But I think the idea is that the white noise
added by the dither sounds better than the quantization noise.

I will try to add a test to patest_converters.c that test dither scaling.

RossBencina:
See #112 "require clipping for dithering sample conversion functions?"

RossBencina:
Related #35 "some conversion functions not implemented"

RossBencina:
General note: I'm not sure that we have the correct scaling factor for usage of triangular noise in any of the dithering functions. It's on my mental todo list to verify that we are using the correct scaling factor.

src/common/pa_converters.c
dmitrykos: Applied changes with `<< 8` for a dither:
```c
=== Combined dynamic range (src->dest->float32) ===
Key: Noise amplitude in dBfs, X - fail (either above failed or dest->float32 failed)
{{{
in|  out:      f32     i32     i24     i16      i8     ui8
 f32          -inf  -180.6  -138.5   -79.5   -31.5   -31.5
 i32        -144.5  -144.5     X     -85.9   -37.6   -37.2
 i24        -138.5  -138.5  -138.5   -86.1   -38.2   -37.3
 i16         -84.8   -84.8   -84.8   -84.8   -37.9   -37.0
  i8         -36.6   -36.6   -36.6   -36.6   -36.6   -36.6
 ui8         -36.4   -36.4   -36.4   -36.4   -36.4   -36.4
}}}
```

I also corrected:
- Int24_To_Int8_Dither
- Int32_To_Int8_Dither

to make their implementation aligned with other 8-bit converters. 

If you check DR for i8 and ui8, it is now affected by a dither.

I am now concerned about this:
```c
(temp >> 1) + dither
```

In my implementation I used library-existing method of applying dither but it does not look ok to me. Why shall we truncate real audio data by 1 bit before adding dither value? In fact, if we do not do it and just truncate after adding the dither the DR improves by around 1.1 dB:

```c
=== Combined dynamic range (src->dest->float32) ===
Key: Noise amplitude in dBfs, X - fail (either above failed or dest->float32 failed)
{{{
in|  out:      f32     i32     i24     i16      i8     ui8
 f32          -inf  -180.6  -138.5   -79.5   -31.5   -31.5
 i32        -144.5  -144.5     X     -85.9   -37.6   -37.2
 i24        -138.5  -138.5  -138.5   -86.1   -39.4   -39.5
 i16         -84.8   -84.8   -84.8   -84.8   -39.8   -39.7
  i8         -36.6   -36.6   -36.6   -36.6   -36.6   -36.6
 ui8         -36.4   -36.4   -36.4   -36.4   -36.4   -36.4
}}}
```

Pseudocode of the change:
```c
(((temp >> 0) + dither) >> 24)
```

Shall we change to this approach by simplifying this to add and truncate as long as DR improves?

philburk: We shift right by one so that we have enough range to clip after adding the dither.

RossBencina (approved):
This looks good. Definitely an improvement. I'm still not sure that our global policy on dither scaling is correct but that's a matter for later. Phil tested with his new test code and notes that there is still a clipping problem with the dither converters, but that is covered by #112.

PR #1038, opened 2025-06-10
Avoid infinite loops in ALSA code

### LINKED ISSUE #1037: ALSA has infinite loops and hangs when "goto error" called in an error handler
There is an infinite loop in some of the error recovery in pa_linux_alsa.c that causes some apps to hang.

The infinite loop occurs in the error handling code for ALSA.
This can cause either the callback API or the blocking read/write API to hang
when an ALSA stream fails. This could, for example, occur if one unplugged
or power cycled a USB device.

**To Reproduce**
1. Connect a Behringer UMC1820 to a Raspberry Pi 4.
1. Run my Max Neuhaus app. (Don't ask. I will try to find a portable way to repro this.)
2. Power cycle the UMC1820

**Expected behavior**
The app to detect the error and recover.

**Actual behavior**
The app hangs and I get 1000's of error messages per second saying something like:

    'self->capture.ready || self->playback.ready' failed...

 - OS: Raspbian
 - PortAudio version: current and older versions

**Additional context**
Here is an excerpt of the code from pa_linux_alsa.c at line 3999:
```
error:
    [snip]
            /* If we're reporting frames eligible for processing, one of the handles better be ready */
            PA_UNLESS( self->capture.ready || self->playback.ready, paInternalError );
```

Note that PA_UNLESS calls "goto error" so this results in an infinite loop!

We need to review all of pa_linux_alsa.c for this pattern.

philburk:
Do not use the error macros that call "goto error" after an "error:" label.

Fixes #1037

### DIFF
--- a/src/hostapi/alsa/pa_linux_alsa.c
+++ b/src/hostapi/alsa/pa_linux_alsa.c
@@ -565,7 +565,7 @@ static void PaAlsa_CloseLibrary()
 }
 
 /* Check return value of ALSA function, and map it to PaError */
-#define ENSURE_(expr, code) \
+#define ENSURE_ON_ERROR_(expr, code, _on_error) \
     do { \
         int __pa_unsure_error_id;\
         if( UNLIKELY( (__pa_unsure_error_id = (expr)) < 0 ) ) \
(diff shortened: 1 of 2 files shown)


src/hostapi/alsa/pa_linux_alsa.c
```diff
@@ -3291,7 +3297,7 @@ static PaError AlsaRestart( PaAlsaStream *stream )
     PA_DEBUG(( "%s: Restarted audio\n", __FUNCTION__ ));
 
 error:
-    PA_ENSURE( PaUnixMutex_Unlock( &stream->stateMtx ) );
+    PA_ENSURE_SAFE( PaUnixMutex_Unlock( &stream->stateMtx ) );
```
philburk: Better name is PA_ENSURE_NO_GOTO_

RossBencina:
I'm OK for you to go ahead merge this after you fix the macro name.

philburk:
I tested this with the Max Neuhaus code and now it recovers gracefully from power cycling the USB device.

PR #1043, opened 2025-06-13
patest_converters.c: fix printed table label: "dynamic range" -> "max abs diff"

RossBencina:
The test computes maximum absolute differences over a non-exhaustive test signal. It does not really establish the dynamic range of the conversion. Update the labels on the printed output accordingly.


philburk (approved):

philburk:
@RossBencina - feel free to merge

PR #1046, opened 2025-06-25
Pulseaudio: Update pulseaudio comments

illuusio:
This pull request addresses concerns regarding the comments in the file by updating several comments within the PulseAudio files.

### DIFF
--- a/src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
+++ b/src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
@@ -248,18 +248,19 @@ static int _PaPulseAudio_ProcessAudio(PaPulseAudio_Stream *stream,
     void *bufferData = NULL;
     size_t pulseaudioOutputWritten = 0;
 
-    /* If there is no specified per host buffer then
-     * just generate one or but correct one in place
+    /* If a per-host buffer is not specified, a new buffer should
+     * be generated or the existing one should be corrected in place.
      */
     if( hostFramesPerBuffer == paFramesPerBufferUnspecified )
     {
         if( !stream->framesPerHostCallback )
         {
-            /* This just good enough and most
-             * Pulseaudio server and ALSA can handle it
+            /*
+             * This configuration is deemed sufficient, as it is
+             * compatible with most PulseAudio servers and ALSA.
              *
-             * We should never get here but this is ultimate
-             * backup.
+             * Although reaching this point is not expected, it serves
+             * as a final backup measure.
              */
             hostFramesPerBuffer = PAPULSEAUDIO_FRAMESPERBUFFERUNSPEC;
 
@@ -273,7 +274,7 @@ static int _PaPulseAudio_ProcessAudio(PaPulseAudio_Stream *stream,
 
     if( stream->outputStream )
     {
-        /* Calculate how many bytes goes to one frame */
+        /* The number of bytes allocated to a single frame should be calculated. */
         pulseaudioInputBytes = pulseaudioOutputBytes = (hostFramesPerBuffer * stream->outputFrameSize);
 
         if( stream->bufferProcessor.streamCallback )
@@ -282,8 +283,8 @@ static int _PaPulseAudio_ProcessAudio(PaPulseAudio_Stream *stream,
         }
     }
 
-    /* If we just want to have input but not output (Not Duplex)
-     * Use this calculation
+    /* If input is desired without output (non-duplex operation),
+     * the following calculation should be utilized.
      */
     if( stream->inputStream )
     {
@@ -295,10 +296,10 @@ static int _PaPulseAudio_ProcessAudio(PaPulseAudio_Stream *stream,
         }
     }
 
-    /* If we have input which is mono and
-     * output which is stereo. We have to copy
-     * mono to monomono which is stereo.
-     * Then just read half and copy
+    /* In the case where the input is mono and the output is stereo,
+     * the mono input must be copied to a mono-mono format, which
+     * corresponds to stereo. Subsequently, only half of the data
+     * should be read and copied.
      */
     if( isOutputCb &&
         stream->outputSampleSpec.channels == 2 &&
@@ -314,8 +315,8 @@ static int _PaPulseAudio_ProcessAudio(PaPulseAudio_Stream *stream,
         /* Allocate memory to make it faster to output stuff */
         pa_stream_begin_write( stream->outputStream, &bufferData, &tmpSize );
 
-        /* If bufferData is NULL then output is not ready
-         * and we have to wait for it
+        /* If bufferData is NULL, it indicates that the output is
+         * not ready, necessitating a wait for its availability.
          */
         if(!bufferData)
         {
@@ -337,8 +338,8 @@ static int _PaPulseAudio_ProcessAudio(PaPulseAudio_Stream *stream,
 
     do
     {
-        /* Make sure that bufferData is NULL that marks there
-         * is nothing to output
+        /* It must be ensured that bufferData is NULL, indicating that
+         * there is no data available for output.
          */
         if( isOutputCb )
         {
@@ -346,14 +347,18 @@ static int _PaPulseAudio_ProcessAudio(PaPulseAudio_Stream *stream,
         }
 
         /*
-         * If everything fail like stream vanish or mainloop
-         * is in error state try to handle error
+         * In the event of failures, such as the stream becoming
+         * unavailable or the main loop entering an error state,
+         * an attempt should be made to handle the error.
          */
         PA_PULSEAUDIO_IS_ERROR( stream, paStreamIsStopped )
 
-        /* There is only Record stream so
-         * see if we have enough stuff to feed record stream
-         * If not then bail out.
+        /* Since only a record stream is present, it must be verified
+         * whether sufficient data is available to support the record
+         * stream.
+         *
+         * If adequate data is not available, the operation should be
+         * terminated.
          */
         if( isInputCb &&
             PaUtil_GetRingBufferReadAvailable(&stream->inputRing) < pulseaudioInputBytes )
@@ -390,11 +395,14 @@ static int _PaPulseAudio_ProcessAudio(PaPulseAudio_Stream *stream,
 
         PaUtil_BeginCpuLoadMeasurement( &stream->cpuLoadMeasurer );
 
-        /* When doing Portaudio Duplex one has to write and read same amount of data
-         * if not done that way Portaudio will go boo boo and nothing works.
-         * This is why this is done as it's done
+        /* In the context of PortAudio duplex operation, it is required
+         * that the same amount of data be written and read. Failure to
+         * adhere to this requirement may result in malfunctioning
+         * behavior of PortAudio.
          *
-         * Pulseaudio does not care and this is something that needs small adjusting
+         * This necessity dictates the implementation approach.
+         * PulseAudio operates independently of this constraint,
+         * necessitating minor adjustments.
          */
         PaUtil_BeginBufferProcessing( &stream->bufferProcessor,
                                       &timeInfo,
@@ -416,8 +424,8 @@ static int _PaPulseAudio_ProcessAudio(PaPulseAudio_Stream *stream,
                                        hostFramesPerBuffer );
         }
 
-        /* Is output is enable and buffer data is not NULL
-         * then crap pointer to output ringbuffer
+        /* If output is enabled and the buffer data is not NULL,
+         *  a pointer to the output ring buffer is assigned.
          */
         if( isOutputCb )
         {
@@ -445,18 +453,20 @@ static int _PaPulseAudio_ProcessAudio(PaPulseAudio_Stream *stream,
                                       hostFrameCount );
 
         /*
-         * pa_stream_begin_write gives pointer to ring
-         * buffer in Pulseaudio. WhenPaUtil_EndBufferProcessing
-         * returns paContinue then is ok to write output.
+         * A pointer to the ring buffer in PulseAudio is provided by
+         * the function pa_stream_begin_write.
+         * When PaUtil_EndBufferProcessing yields a return value of
+         * paContinue, it is indicated that writing output is
+         * permissible.
          *
-         * When it's something else than paContinue
-         * then we have to cancel writing with
-         * with pa_stream_cancel_write.
+         * In instances where a return value other than paContinue is
+         * received, the writing process must be terminated using
+         * pa_stream_cancel_write.
          *
-         * If there is no space in output ringbuffer or for
-         * example USB device has dissapiered and pointer
-         * is NULL then we just get out of loop as there is not much
-         * we can't do
+         * If insufficient space is detected in the output ring buffer,
+         * or if the USB device is no longer present and the pointer is
+         * NULL, the loop is exited, as further actions are rendered
+         * ineffective.
          */
         if( ret == paContinue && isOutputCb && bufferData )
         {
--- a/src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
+++ b/src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
@@ -113,10 +113,10 @@ void PaPulseAudio_ReleaseOperation(PaPulseAudio_HostApiRepresentation *hostapi,
     pa_operation *localOperation = (*operation);
     pa_operation_state_t localOperationState = PA_OPERATION_RUNNING;
 
-    // As mainly operation is done when running locally
-    // done after 1-3 then 1000 is enough to wait if
-    // something goes wrong
-
+    /* Since the primary operations are conducted locally, a wait time
+     * of 1 to 3 seconds, followed by an additional 1000 milliseconds,
+     * is deemed sufficient in the event of an error.
+     */
     while( waitOperation > 0 )
     {
 
@@ -185,11 +185,13 @@ void _PaPulseAudio_WriteRingBuffer( PaUtilRingBuffer *ringbuffer,
                                     size_t length )
 {
     /*
-     * If there is not enough room. Read from ringbuffer to make
-     * sure that if not full and audio will just underrun
+     * If insufficient space is available, data should be read from
+     * the ring buffer to ensure that it is not full and to prevent
+     * audio underrun.
      *
-     * If you try to read too much and there is no room then this
-     * will fail. But I don't know how to get into that?
+     * Attempting to read an excessive amount of data when there is
+     * insufficient space will result in failure; however, the
+     * conditions leading to this situation are not clearly understood.
      */
     if( PaUtil_GetRingBufferWriteAvailable( ringbuffer ) < length )
     {
@@ -211,9 +213,10 @@ void _PaPulseAudio_Read( PaPulseAudio_Stream *stream,
     const void *pulseaudioData = NULL;
 
     /*
-     * Idea behind this is that we fill ringbuffer with data
-     * that comes from input device. When it's available
-     * we'll fill it to callback or blocking reading
+     * The concept underlying this implementation is to populate the
+     * ring buffer with data obtained from the input device. Once the
+     * data is available, it will be utilized for either callback
+     * processing or blocking reads.
      */
     if( pa_stream_peek( stream->inputStream,
                         &pulseaudioData,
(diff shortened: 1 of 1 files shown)


src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
```diff
@@ -584,7 +597,9 @@ PaError PaPulseAudio_CloseStreamCb( PaStream * s )
         && PA_STREAM_IS_GOOD( pa_stream_get_state( stream->outputStream ) ) )
     {
         PaPulseAudio_Lock(stream->mainloop);
-        /* Pause stream so it stops faster */
+        /* The stream should be paused to facilitate a quicker cessation
```
philburk: The original comment was more concise and more clear. How were these comments generated?

illuusio: Mostly with GPT and then altered by hand

illuusio:
I've made more small adjustments to comments. Now they should be more shorter and precises.

RossBencina:
@illuusio I have pushed some edits (see my latest commit). Could you please review them? I have some additional remarks which I'll put into a review now.

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
```diff
@@ -109,36 +109,39 @@ int PaPulseAudio_updateTimeInfo( pa_stream * s,
...
+    unsigned int wait = 1000;
     pa_operation *localOperation = (*operation);
     pa_operation_state_t localOperationState = PA_OPERATION_RUNNING;
 
-    // As mainly operation is done when running locally
-    // done after 1-3 then 1000 is enough to wait if
-    // something goes wrong
-
-    while( waitOperation > 0 )
+    /* Since the primary operations are conducted locally, a wait time
```
RossBencina: What does this mean? what does "mainly operation is done when running locally" or "conducted locally" mean?

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
```diff
@@ -109,36 +109,39 @@ int PaPulseAudio_updateTimeInfo( pa_stream * s,
...
         {
-            // Result is DONE or CANCEL
+            /* Result is DONE or CANCEL */
             PaPulseAudio_UnLock( hostapi->mainloop );
             break;
         }
         PaPulseAudio_UnLock( hostapi->mainloop );
 
-        waitOperation --;
+        wait --;
```
RossBencina: ?BUG: if the earlier comment is correct, and the retry loop runs for seconds/milliseconds, shouldn't there be a call to `usleep()` here at the end of the loop body?

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
```diff
@@ -185,11 +188,13 @@ void _PaPulseAudio_WriteRingBuffer( PaUtilRingBuffer *ringbuffer,
                                     size_t length )
 {
     /*
-     * If there is not enough room. Read from ringbuffer to make
-     * sure that if not full and audio will just underrun
+     * If insufficient space is available, read data from
+     * the ring buffer to ensure that it is not full and to prevent
+     * audio underrun.
```
RossBencina: If the buffer is full, and data has to be discarded, that would be an "overrun" not an "underrun". Does that sound correct to you?

```suggestion
     * audio overrun.
```

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
```diff
@@ -282,8 +286,8 @@ static int _PaPulseAudio_ProcessAudio(PaPulseAudio_Stream *stream,
         }
     }
 
-    /* If we just want to have input but not output (Not Duplex)
-     * Use this calculation
+    /* If input is desired without output (non-duplex operation),
```
RossBencina: It is not clear to me whether there is a problem with the comment or the code here.

Are you sure that this logic is for "input without output" aka "input but not output" (also known as input-only or half-duplex input)? It seems to me that this logic is for all cases where there is input (i.e. half-duplex input only, AND full-duplex input+output), but perhaps I am mistaken? The logic in many places has `if ( stream->outputStream)` and `if( stream->inputStream )`, never `else if` so my reading is that `stream->inputStream` evaluates to true for both input-only and input+output PortAudio streams.

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
```diff
@@ -295,10 +299,10 @@ static int _PaPulseAudio_ProcessAudio(PaPulseAudio_Stream *stream,
         }
     }
 
-    /* If we have input which is mono and
-     * output which is stereo. We have to copy
-     * mono to monomono which is stereo.
-     * Then just read half and copy
+    /* In the case where the input is mono and the output is stereo,
+     * the mono input must be copied to a mono-mono format, which
+     * corresponds to stereo. Subsequently, only half of the data
```
```suggestion
     * corresponds to stereo. Subsequently, only half as much data
```

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
```diff
@@ -337,23 +341,23 @@ static int _PaPulseAudio_ProcessAudio(PaPulseAudio_Stream *stream,
...
-         * is in error state try to handle error
+        /* Handle failures, such as the stream becoming unavailable or the main loop
+         * entering an error state.
          */
         PA_PULSEAUDIO_IS_ERROR( stream, paStreamIsStopped )
 
-        /* There is only Record stream so
-         * see if we have enough stuff to feed record stream
-         * If not then bail out.
+        /* Since only a record stream is present, it must be verified
```
RossBencina: Once again the logic seems inconsistent with the comment here, there is no check for "only a record stream is present," `isInputCb` can be set above even when there is an output stream (i.e. in the full duplex case). Maybe I am wrong?

In case the comment is correct, here's a suggested improvement: "In the input-only case, check whether we have enough data available to feed the buffer processor, if not bail out."

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
```diff
@@ -390,17 +394,22 @@ static int _PaPulseAudio_ProcessAudio(PaPulseAudio_Stream *stream,
...
-         * This is why this is done as it's done
+        /* In the context of PortAudio duplex operation, it is required
+         * that the same amount of data be written and read. Failure to
+         * adhere to this requirement may result in malfunctioning
+         * behavior of PortAudio.
          *
-         * Pulseaudio does not care and this is something that needs small adjusting
+         * This necessity dictates the implementation approach.
+         * PulseAudio operates independently of this constraint,
+         * necessitating minor adjustments.
```
RossBencina: what does "minor adjustments" mean? adjustments to what? which adjustments?

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
```diff
@@ -390,17 +394,22 @@ static int _PaPulseAudio_ProcessAudio(PaPulseAudio_Stream *stream,
 
         PaUtil_BeginCpuLoadMeasurement( &stream->cpuLoadMeasurer );
 
-        /* When doing Portaudio Duplex one has to write and read same amount of data
-         * if not done that way Portaudio will go boo boo and nothing works.
-         * This is why this is done as it's done
+        /* In the context of PortAudio duplex operation, it is required
```
RossBencina: I'm not sure how to do a multi-line suggestion in a github review, but here's a DRAFT of what I'm thinking, subject to resolving what "minor adjustments" is about (see below).

```
        /* NOTE: In the full duplex case PortAudio requires us to read and write the same amount
         * of data, otherwise it doesn't work. This dictates our implementation approach.
         * Pulseaudio has no such constraint, necessitating minor adjustments.
         */
```

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
```diff
@@ -314,8 +318,8 @@ static int _PaPulseAudio_ProcessAudio(PaPulseAudio_Stream *stream,
         /* Allocate memory to make it faster to output stuff */
         pa_stream_begin_write( stream->outputStream, &bufferData, &tmpSize );
 
-        /* If bufferData is NULL then output is not ready
-         * and we have to wait for it
+        /* If bufferData is NULL, it indicates that the output is
+         * not ready, necessitating a wait for its availability.
```
```suggestion
         * not ready -- we have to wait for it to become available.
```

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
```diff
@@ -390,17 +394,22 @@ static int _PaPulseAudio_ProcessAudio(PaPulseAudio_Stream *stream,
...
+         * This necessity dictates the implementation approach.
+         * PulseAudio operates independently of this constraint,
+         * necessitating minor adjustments.
          */
         PaUtil_BeginBufferProcessing( &stream->bufferProcessor,
                                       &timeInfo,
                                       0 );
 
-        /* Read of ther is something to read */
+        /* Read from input buffer if there is enough bytes or add
```
```suggestion
        /* Read from input buffer if there are enough bytes or add
```

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
```diff
@@ -390,17 +394,22 @@ static int _PaPulseAudio_ProcessAudio(PaPulseAudio_Stream *stream,
...
+         * zero if there is not
```
```suggestion
         * zero if there are not
```

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
```diff
@@ -827,16 +832,19 @@ PaError PaPulseAudio_StartStreamCb( PaStream * s )
...
     if( stream->outputStream )
     {
-        /* tlength does almost the same as fragsize in record.
-         * See reasoning up there in comments.
+        /* The tlength parameter functions similarly to fragsize in
+         * the recording process, as explained in the preceding comments.
          *
-         * In future this should we tuned when things changed
-         * this just 'good' default
+         * In the future, this should be adjusted as necessary when
```
RossBencina: What does "in the future" mean? is this a `TODO` or `FIXME`? what does "conditions change" mean? what conditions?

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
```diff
@@ -416,15 +425,15 @@ static int _PaPulseAudio_ProcessAudio(PaPulseAudio_Stream *stream,
...
+        /* If output is enabled and the buffer data is not NULL,
+         *  a pointer to the output ring buffer is assigned.
          */
         if( isOutputCb )
         {
 
             size_t tmpSize = pulseaudioOutputBytes;
 
-            /* Allocate memory to make it faster to output stuff */
+            /* Pre allocate memory from output buffer which will make output faster */
```
```suggestion
            /* Pre allocate memory from output buffer to make output faster */
```

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
```diff
@@ -926,24 +934,24 @@ PaError PaPulseAudio_StartStreamCb( PaStream * s )
...
+    /* Now that everything is set up, start the callback */
     if( stream->inputStream )
     {
         pa_stream_set_read_callback( stream->inputStream,
                                      PaPulseAudio_StreamRecordCb,
                                      stream );
     }
 
-    /* Allways unlock.. so we don't get locked */
+    /* It is essential to always unlock to prevent being locked out. */
```
RossBencina: What does "being locked out" mean? And why is this comment here? Is this comment more important than any other case where we unlock? maybe this comment is not needed at all?
```suggestion
```
 Or maybe you mean something like:

```
IMPORTANT REMINDER: Before returning from this function, be sure to call PaPulseAudio_UnLock() on all code paths to avoid being locked out.
```

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
```diff
@@ -273,7 +277,7 @@ static int _PaPulseAudio_ProcessAudio(PaPulseAudio_Stream *stream,
 
     if( stream->outputStream )
     {
-        /* Calculate how many bytes goes to one frame */
+        /* Calculate how many bytes go into to one frame */
```
RossBencina: This comment is not consistent with the code. I think that the code is calculating the size of the host buffer in bytes.
```suggestion
        /* Calculate the host buffer size in bytes */
```

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
```diff
@@ -981,7 +989,7 @@ static PaError RequestStop( PaPulseAudio_Stream * stream,
 
     stream->missedBytes = 0;
 
-    /* Test if there is something that we can play */
+    /* Check whether there is any data available for playback. */
```
```suggestion
    /* Check whether there is any data to play */
```

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
```diff
@@ -416,15 +425,15 @@ static int _PaPulseAudio_ProcessAudio(PaPulseAudio_Stream *stream,
                                        hostFramesPerBuffer );
         }
 
-        /* Is output is enable and buffer data is not NULL
-         * then crap pointer to output ringbuffer
+        /* If output is enabled and the buffer data is not NULL,
+         *  a pointer to the output ring buffer is assigned.
```
RossBencina, line 408: Does this capture the intended meaning? It reads very strangely.

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
```diff
@@ -926,24 +934,24 @@ PaError PaPulseAudio_StartStreamCb( PaStream * s )
         goto startstreamcb_error;
     }
 
-    /* Make sure we pass no error on intialize */
+    /* It must be ensured that no errors are encountered during
+     * initialization.
+     */
```
RossBencina: This comment (and the original version) is confusing. I think better would be:

```
    /* Success! */
```

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
```diff
@@ -109,36 +109,39 @@ int PaPulseAudio_updateTimeInfo( pa_stream * s,
...
     pa_operation_state_t localOperationState = PA_OPERATION_RUNNING;
 
-    // As mainly operation is done when running locally
-    // done after 1-3 then 1000 is enough to wait if
-    // something goes wrong
-
-    while( waitOperation > 0 )
+    /* Since the primary operations are conducted locally, a wait time
+     * of 1 to 3 seconds, followed by an additional 1000 milliseconds,
+     * is deemed sufficient to to detect successful completion or to detect an error.
```
philburk: "to to"

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
illuusio: Yes there should. Good catch.

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
illuusio: Ok I've update this comment logic to be more sane (I think and hope)

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
illuusio: Fixed this one also

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
illuusio: It's bit shady logic but works. I've updated comment

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
illuusio: I've removed that part as I couldn't iron out my idea..

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
illuusio: Seems to be updated already.

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
illuusio: Removed altogether as it didn't add nothing to comments.

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
illuusio: Update this one.

illuusio:
Is this ready to merge?

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
RossBencina: Each PR needs to address a specific issue. Changing code for an unrelated purpose is not appropriate in the current PR. Please factor all of the changes to `PaPulseAudio_ReleaseOperation` (including both the code and  the comments) into a separate PR. The comments in this function may need to be clarified but we can discuss that in the new PR.

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
RossBencina: Awaiting your response of my review for doc comment at line 402 ("In the context of PortAudio duplex operation...") in the current version.

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
RossBencina: Nothing has been updated. The comment still makes little sense. Unless you can explain why a comment is needed there I suggest that you delete the entire comment "If output is enabled and the buffer data is not NULL,  a pointer to the output ring buffer is assigned."

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
RossBencina: Awaiting your response to my review for doc comment at line 360 ("Since only a record stream is present...")

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
RossBencina: Awaiting your response of my review for doc comment at line 402 ("In the context of PortAudio duplex operation...") in the current version.

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
RossBencina: Please factor to separate PR as described in other review comment.

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
RossBencina: Ok, I think the comment should be deleted altogether. I will do that.

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
RossBencina: I'm just going to delete it.

RossBencina:
Thanks for the updates. I have deleted two comments that were low-quality and redundant.

Remaining issues:

- Please factor code and comment changes to `PaPulseAudio_ReleaseOperation ()` into a separate PR as detailed in my review comments today.

- Please respond to my three unaddressed review comments (for lines 360 and 402). You may need to click the "X hidden conversations. Load more..." button to see them.

illuusio:
> Thanks for the updates. I have deleted two comments that were low-quality and redundant.
> 
> Remaining issues:
> 
>     * Please factor code and comment changes to `PaPulseAudio_ReleaseOperation ()` into a separate PR as detailed in my review comments today.
> 
>     * Please respond to my three unaddressed review comments (for lines 360 and 402). You may need to click the "X hidden conversations. Load more..." button to see them.

I've decided to remove comment 402 as it does belong there and it's not correct place to comment that. On line 360 I've updated commenting and I've formed new PR (actually 2 to also remove that missedBytes which is not used anywhere and should not be there anymore)

philburk:
The two PRs that were separated out are #1103 and #1104.

Issue #1042, opened 2025-06-13
Document Sample Format Conversion Policies


RossBencina:
Make a doxygen doc addressing:

- Linearity about 0
- DC offset
- Scale of 1.0
- Transparancy of round-trip conversions (e.g. int8 -> float -> int8)
- Which modes allow integer wrap-around (e.g. does dither imply clipping)

The purpose of this document is to establish policies that can be tested and implemented.

RossBencina:
I was heavily involved in mailing list discussions when we hashed this out. I believe that some things were clearly established, and in hindsight other things were overlooked or not understood by me. My view is that the main policy decisions that were established and (hopefully) implemented long ago are:

- zero maps to zero
-  when converting from float to int, +1.0 and -1.0 should map into integer formats without clipping (e.g. for float -> int8, 1.0 -> 127, -1.0 -> -127).
- when converting from int to float, the output must always be in the range [1.0, -1.0] inclusive. (e.g. for int8 -> float, 128 -> 1.0)
- when clipping is enabled, out of range values should saturate, not wrap around.

Although linearity around 0 and 0 DC bias are obviously desirable, I didn't fully understand this at the time. I think we may still have problems here.

RossBencina:
Correct scaling of dither noise is a related issue, I'm not sure whether it is a policy or an implementation detail.

dmitrykos:
> * when converting from float to int, +1.0 and -1.0 should map into integer formats without clipping (e.g. for float -> int8, 1.0 -> 127, -1.0 -> -127).

To my view the scaling, if we take example of float -> int8 needs to be 128 (0x80). With this approach -1.0 will reach min -128, and +1.0 will have to be truncated/clamped to 127 during the conversion. 0x80 will give full dynamic range when float is converted to int8 because -1.0 values will not be missed from the conversion. Integers are asymmetrical by their nature therefore from the perspective of dynamic range (and thus audio quality) the best scaling value is -128 (0x80).

It is easy to see the effect of 0x7F or 0x80 scalers on DR with `patest_converters` and my #1041 PR `Float32_To_Int8_Clip`:

This code gives -42.2 dB DR that is 7.03 bits of valid data:
```c
/* convert to signed 8-bit */
samp = (PaInt32)(*src * 0x80);

/* clip output */
PA_CLIP_( samp, -0x80, 0x7F );

/* convert to unsigned */
*dest = (signed char) samp;
```

[...]

PR #1051, opened 2025-07-24
Fix format converter issues found by paqa_dither

philburk:
Use floatf() to fix non-linearity at 0.0.
Always clip when doing dithering to prevent numeric wrap-around.
Add extra half bit to compensate for bias when truncating.

### DIFF
--- a/src/common/pa_converters.c
+++ b/src/common/pa_converters.c
@@ -58,12 +58,14 @@
  Int32_To_Int16_Dither, Int32_To_Int8_Dither, Int16_To_Int32
 */
 
+#include <math.h>
 
 #include "pa_converters.h"
 #include "pa_dither.h"
 #include "pa_endianness.h"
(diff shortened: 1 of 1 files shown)


src/common/pa_converters.c
```diff
@@ -574,7 +581,7 @@ static void Float32_To_Int16(
 
     while( count-- )
     {
-        short samp = (short) (*src * (32767.0f));
+        short samp = (short) floorf(*src * (32767.0f));
```
philburk, line 584: Maybe use PA_FLOAT_TO_INT32 here.

philburk:
TODO:
* Check the CPU performance difference for these changes.
* Check whether using floorf() can cause numeric wraparound for values just under -1.0.
* Write mathematical definition of the required behavior. Then test the behavior. #1042

PR #1052, opened 2025-07-29
initialize variable-sized arrays explicitly

smoofra:
Initialize variable-sized arrays to zero explicitly, to avoid the following error:

```
qa/paqa_dither.c:155:5: error: variable-sized object may not be initialized
  155 |     int histogram[kNumBins] = {};
      |     ^~~

$ gcc --version
gcc (Raspbian 10.2.1-6+rpi1) 10.2.1 20210110
```

### DIFF
--- a/qa/paqa_dither.c
+++ b/qa/paqa_dither.c
@@ -152,8 +152,11 @@ int ShowDitherDistribution( void )
     const int kNumSamples = 24 * 1024;
     PaInt32 minDither = 0x01000000;
     PaInt32 maxDither = -0x01000000;
-    int histogram[kNumBins] = {};
+    int histogram[kNumBins];
     int maxCount = 0;
+
+    for (int i = 0; i < kNumBins; i++)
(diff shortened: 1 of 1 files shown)


qa/paqa_dither.c
```diff
@@ -273,14 +276,19 @@ static int TestDitherLinearity(PaSampleFormat sourceFormat, PaSampleFormat desti
...
-    double expected[kNumSteps] = {};
+    double averages[kNumSteps];
+    double expected[kNumSteps];
     double stride = (maxValue - minValue) / (kNumSteps - 1);
     char supported = 0;
     double slope = 0.0;
     double bias = 0.0;
     double rSquared;
 
+    for (int i = 0; i < kNumSteps; i++) {
```
philburk: We can remove this block because all the values are set at lines 297,298.

smoofra: good point, fixed.

philburk (approved):

PR #1059, opened 2025-08-21
alsa: fix goto related errors reported by clang

### LINKED ISSUE #516: ALSA push/pop hack breaks Clang 10
error: cannot jump from this goto statement to its label
    goto end;

jump bypasses initialization of variable with __attribute__((cleanup))
    pthread_cleanup_push( &OnExit, stream );

/usr/include/pthread.h:599:36: note: expanded from macro 'pthread_cleanup_push'
    struct __pthread_cleanup_frame __clframe

philburk:
Move labels around to prevent the errors.

Fixes #516

Thanks to @ongyx on Github for the patch.

### DIFF
--- a/src/hostapi/alsa/pa_linux_alsa.c
+++ b/src/hostapi/alsa/pa_linux_alsa.c
@@ -4458,17 +4458,17 @@ static void *CallbackThreadFunc( void *userData )
         }
     }
 
-end:
-    ; /* Hack to fix "label at end of compound statement" error caused by pthread_cleanup_pop(1) macro. */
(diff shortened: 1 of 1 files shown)


philburk:
The CMAKE build errors in CI for MSVC are unrelated because this only changes Linux builds.

RossBencina (approved):
Slight concern about what happens if someone modifies the `while (1)` loop in the future to fall through. We considered additional mitigations (an extra `/* DOES NOT FALL THROUGH */` comment, or gating the `PA_DEBUG` call on non-zero error), but nothing seemed an obvious improvement.

PR #1068, opened 2025-08-29
ALSA: allow more channels because some plugins have 10000

### LINKED ISSUE #797: Rejecting ALSA devices with high max number of channels is wrong
**Describe the bug**
I have a very normal audio setup without fancy new stuff like  PulseAudio or PipeWire. Just plain ALSA. There is no special configuration. I use the default configs installed with alsa-lib. The default config makes the default audio device a softvol+dmix ALSA plugin combination that mixes the audio from all applications in software to be output by the HDA on board soundchip.

The plugins don't limit the number of channels they can accept. libasound [limits the number of channels to 10000](https://github.com/alsa-project/alsa-lib/blob/master/src/pcm/pcm_plug.c#L705) when the plugin/hardware does not impose another limit.

Since b3f637f55e74c7b8fd05ebdfb04a2c8547e316ba PortAudio rejects all devices that allow more than 1024 channels, which breaks sound for me.

**To Reproduce**
1. Install Linux without PulseAudio or PipeWire
2. Try to play sound 

**Expected behavior**
The default ALSA device is used

```
FillInDevInfo: Filling device info for: default
GropeDevice: collecting info ..
GropeDevice: Limiting number of plugin channels to 128
GropeDevice: collecting info ..
GropeDevice: Limiting number of plugin channels to 128
Default input device: default
Default output device: default
FillInDevInfo: Adding device default: 13
```

**Actual behavior**
The default ALSA device is rejected

```
FillInDevInfo: Filling device info for: default
GropeDevice: collecting info ..
GropeDevice: maxChans = 10000, which is unreasonably high
FillInDevInfo: Failed groping default for capture
```

 - OS: Linux
 - OS Version: 5.15, alsa-lib 1.2.8
 - PortAudio version: master branch

philburk:
The old code was rejecting OK devices.

Also fixed some broken PA_DEBUG statements.

Fixes #797

### DIFF
--- a/src/hostapi/alsa/pa_linux_alsa.c
+++ b/src/hostapi/alsa/pa_linux_alsa.c
@@ -912,12 +912,22 @@ static PaError GropeDevice( snd_pcm_t* pcm, int isPlug, StreamDirection mode, in
 
     ENSURE_( alsa_snd_pcm_hw_params_get_channels_min( hwParams, &minChans ), paUnanticipatedHostError );
     ENSURE_( alsa_snd_pcm_hw_params_get_channels_max( hwParams, &maxChans ), paUnanticipatedHostError );
+    /* ALSA may set maxChans to UINT_MAX if something is wrong.
+     * Some plugins may set it to 10000, in which case we will clip to a lower value.
+     */
+    const unsigned int kBadMaxChannels = 20000;
     const unsigned int kReasonableMaxChannels = 1024;
-    if( maxChans > kReasonableMaxChannels )
+    if( maxChans > kBadMaxChannels )
     {
-        PA_DEBUG(( "%s: maxChans = %u, which is unreasonably high\n", __FUNCTION__, maxChans ));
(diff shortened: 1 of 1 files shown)


src/hostapi/alsa/pa_linux_alsa.c
```diff
@@ -912,13 +912,23 @@ static PaError GropeDevice( snd_pcm_t* pcm, int isPlug, StreamDirection mode, in
 
     ENSURE_( alsa_snd_pcm_hw_params_get_channels_min( hwParams, &minChans ), paUnanticipatedHostError );
     ENSURE_( alsa_snd_pcm_hw_params_get_channels_max( hwParams, &maxChans ), paUnanticipatedHostError );
+    /* ALSA may set maxChans to UINT_MAX if something is wrong.
+     * Some plugins may set it to 10000, in which case we will clip to a lower value.
```
RossBencina:
```suggestion
     * Our understanding is that if a plugin does not set maxChans then ALSA sets it to 10000,
       in which case we will clip to a lower value.
```

PR #1070, opened 2025-08-29
ALSA: set buffer time when testing formats

### LINKED ISSUE #526: ALSA alsa_snd_pcm_hw_params() on Ubuntu 20.04 started returning an unexpected -ENOMEM
Theo Veenker reported a bug on the PA list.
We may need a workaround in PortAudio.

------------------------------

Under Ubuntu 20.04 (clean out of the box) I'm encountering a problem with
Pa_IsFormatSupported(). The attached test program demonstrates the problem. Under Linux
Mint 18 (which is based on Ubuntu 16.04) the program works fine, with both
v190600_20161030 and v190700_20210307. Under Ubuntu 20.04 however it reports a failure at
hostapi/alsa/pa_linux_alsa.c line 1818 (or a few lines up for v190600_20161030).

The problem also occurs in Audacity according to its debug output.

The problematic part is this (in pa_linux_alsa.c):

         int ret = 0;
         if( ( ret = alsa_snd_pcm_hw_params( pcm, hwParams ) ) < 0 )
         {
             if( -EINVAL == ret )
             {
                 /* Don't know what to return here */
                 result = paBadIODeviceCombination;
                 goto error;
             }
             else if( -EBUSY == ret )
             {
                 result = paDeviceUnavailable;
                 PA_DEBUG(( "%s: Device is busy\n", __FUNCTION__ ));
             }
             else
             {
                 result = paUnanticipatedHostError;     line 1815
             }

             ENSURE_( ret, result );                    line 1818
         }

Apparently alsa_snd_pcm_hw_params() returns -12 (which is -ENOMEM). No memory seems
unlikely here. If I comment out lines 1815 and 1818 the program runs fine. So the device
can play audio just fine, but it is just not possible to check whether a particular
format/config is valid.

philburk:
This prevents an ENOMEM error on some devices when the buffer size is unspecified when testing formats.

Fixes #526

### DIFF
--- a/src/hostapi/alsa/pa_linux_alsa.c
+++ b/src/hostapi/alsa/pa_linux_alsa.c
@@ -1828,7 +1828,7 @@ static PaError TestParameters( const PaUtilHostApiRepresentation *hostApi, const
     /*
      * Intel HDA driver doesn't set PCM rule to limit maximum size of buffer.
      * This can result in a request for too large a buffer size.
-     * That can cause memory allocation error in ALSA PCM core, at least Linux kernel 5.8.
+     * That can cause a memory allocation error in ALSA PCM core, at least it does in Linux kernel 5.8.
      * As a workaround, limit buffer size to a reasonable value.
      */
     {


src/hostapi/alsa/pa_linux_alsa.c
```diff
@@ -1825,6 +1825,19 @@ static PaError TestParameters( const PaUtilHostApiRepresentation *hostApi, const
     /* Some specific hardware (reported: Audio8 DJ) can fail with assertion during this step. */
     ENSURE_( alsa_snd_pcm_hw_params_set_format( pcm, hwParams, Pa2AlsaFormat( hostFormat ) ), paUnanticipatedHostError );
 
+    /*
+     * Intel HDA driver doesn't set PCM rule to limit maximum size of buffer.
+     * This can result in a request for too large a buffer size.
+     * That can cause memory allocation error in ALSA PCM core, at least Linux kernel 5.8.
```
RossBencina:
```suggestion
     * That can cause a memory allocation error in ALSA PCM core, at least it does in Linux kernel 5.8.
```

PR #1072, opened 2025-09-18
pulseaudio: free input ring buffer on close and open error (fixes #968)

### LINKED ISSUE #968: PulseAudio OpenStream Memory Leak
LeakSanitizer has helped me find a memory leak in the PaPulseAudio_BlockingInitRingBuffer function that is called by OpenStream. This function calls malloc( size ) that is not freed in CloseStream.

**To Reproduce**
1. Call Pa_OpenStream
2. Call Pa_CloseStream
3. Build program using -fsanitize=address,undefined to point out the memory leak
4. Run program on Ubuntu 22
5. When program ends LeakSanitizer states there is leak in pa_linux_pulseaudio.c:960

**Expected behavior**
There is no leak after program ends.

**Actual behavior**
LeakSanitizer reports a leak. See attached screenshot.
image

 - OS: Ubuntu 22
 - PortAudio version: stable, nightly snapshot (which?), current (please give date and/or Git hash):
 Current, git hash - 57aa393109ec996799d3a5846c9ecb0a65b64644
 - If Windows or Linux, which Host API (e.g. WASAPI):
 PulseAudio

**Additional context**

Note: PortAudio is a community supported project. If you have a solution, please create a Pull Request for us to consider.
I strongly suspect this could be fixed by freeing this memory at the end of PaPulseAudio_CloseStreamCb within pa_linux_pulseaudio_cb.c. If I hear no objection, I'll free this malloc'd memory at this location and test my project. If this works for me I can create a Pull Request with this change. Unfortunately I'm not that familiar with git so this may take some time.

jaredhagel2:
Problem
-------
A 262,144-byte leak is reported by ASAN when opening/closing a PulseAudio input stream. The leak originates from memory allocated in `PaPulseAudio_BlockingInitRingBuffer` and not freed on stream close or when `OpenStream` fails.

Root cause
----------
`PaPulseAudio_BlockingInitRingBuffer` allocates a raw buffer for the input ring. The buffer wasn’t freed in `PaPulseAudio_CloseStreamCb` or in the `openstream_error` cleanup path, leading to a persistent allocation per stream.

Fix
---
- In `PaPulseAudio_CloseStreamCb`, free `stream->inputRing.buffer` after stream termination and reset ring fields.
- In `OpenStream` error cleanup, free `stream->inputRing.buffer` if initialized and reset ring fields.

Impact and validation
---------------------
- Eliminates the ASAN-reported direct leak of 262,144 bytes.
- Tested by running a simple open/start/stop/close flow using the PulseAudio host on Linux with ASAN/LSAN enabled; leak count returns to zero for this allocation.

Notes
-----
- No changes are required in user code; this strictly fixes host API cleanup.
- Fixes #968.

### DIFF
--- a/src/hostapi/pulseaudio/pa_linux_pulseaudio.c
+++ b/src/hostapi/pulseaudio/pa_linux_pulseaudio.c
@@ -1367,7 +1367,7 @@ PaError OpenStream( struct PaUtilHostApiRepresentation *hostApi,
 
     if( stream )
     {
-        /* If the blocking input ring buffer was initialized, release it. */
+        /* If the blocking input ring buffer was allocated, release it. */
         if( stream->inputRing.buffer )
         {
             free( stream->inputRing.buffer );
--- a/src/hostapi/pulseaudio/pa_linux_pulseaudio.c
+++ b/src/hostapi/pulseaudio/pa_linux_pulseaudio.c
@@ -1367,6 +1367,19 @@ PaError OpenStream( struct PaUtilHostApiRepresentation *hostApi,
 
     if( stream )
     {
+        /* If the blocking input ring buffer was initialized, release it. */
+        if( stream->inputRing.buffer )
+        {
+            free( stream->inputRing.buffer );
+            stream->inputRing.buffer = NULL;
+            stream->inputRing.bufferSize = 0;
+            stream->inputRing.bigMask = 0;
+            stream->inputRing.smallMask = 0;
+            stream->inputRing.elementSizeBytes = 0;
+            stream->inputRing.readIndex = 0;
+            stream->inputRing.writeIndex = 0;
+        }
+
         PaUtil_FreeMemory( stream->inputStreamName );
         PaUtil_FreeMemory( stream->outputStreamName );
         PaUtil_FreeMemory( stream );
--- a/src/hostapi/pulseaudio/pa_linux_pulseaudio.c
+++ b/src/hostapi/pulseaudio/pa_linux_pulseaudio.c
@@ -1372,12 +1372,6 @@ PaError OpenStream( struct PaUtilHostApiRepresentation *hostApi,
         {
             free( stream->inputRing.buffer );
             stream->inputRing.buffer = NULL;
-            stream->inputRing.bufferSize = 0;
-            stream->inputRing.bigMask = 0;
-            stream->inputRing.smallMask = 0;
-            stream->inputRing.elementSizeBytes = 0;
-            stream->inputRing.readIndex = 0;
-            stream->inputRing.writeIndex = 0;
         }
 
         PaUtil_FreeMemory( stream->inputStreamName );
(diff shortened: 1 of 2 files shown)


src/hostapi/pulseaudio/pa_linux_pulseaudio.c
```diff
@@ -1367,6 +1367,19 @@ PaError OpenStream( struct PaUtilHostApiRepresentation *hostApi,
 
     if( stream )
     {
+        /* If the blocking input ring buffer was initialized, release it. */
+        if( stream->inputRing.buffer )
+        {
+            free( stream->inputRing.buffer );
+            stream->inputRing.buffer = NULL;
+            stream->inputRing.bufferSize = 0;
```
philburk: This code touches a lots of internal members of the ring buffer.
We free the whole thing on line 1385 so this is probably unnecessary.
Plus it is a bit brittle.
After this leak is fixed, we should review the ring buffer code and maybe manage this better.

To clarify, I am referring to setting "bufferSize = 0", etc.

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
```diff
@@ -659,6 +659,21 @@ PaError PaPulseAudio_CloseStreamCb( PaStream * s )
...
     PaUtil_TerminateBufferProcessor( &stream->bufferProcessor );
     PaUtil_TerminateStreamRepresentation( &stream->streamRepresentation );
+    /* Free any memory allocated for the blocking input ring buffer. */
+    if( stream->inputRing.buffer )
+    {
+        /* At this point input/output streams have been disconnected and unref\'d,
+         * so no other thread should be accessing the ring buffer. */
+        free( stream->inputRing.buffer );
+        stream->inputRing.buffer = NULL;
+        stream->inputRing.bufferSize = 0;
```
philburk: See other comment about removing this code.

jaredhagel2:
Thanks, Phil — updated per your feedback. I removed the internal ring buffer field resets and now only free(stream->inputRing.buffer) and set it to NULL in both the close and open error paths. Please let me know if youd prefer we also drop the NULL assignment or adjust anything else.

src/hostapi/pulseaudio/pa_linux_pulseaudio.c
```diff
@@ -1367,6 +1367,13 @@ PaError OpenStream( struct PaUtilHostApiRepresentation *hostApi,
 
     if( stream )
     {
+        /* If the blocking input ring buffer was initialized, release it. */
```
RossBencina:
```suggestion
        /* If the blocking input ring buffer was allocated, release it. */
```

RossBencina (approved):
Awesome. Approved with slight comment edit.

philburk:
This was great. Setting the pointer to NULL after the free() is necessary and sufficient.
Thanks for the well organized PR.

Issue #1074, opened 2025-09-19
Callbacks sometimes hang after a few hours


philburk:
This usually happens with ALSA running on embedded devices such as Raspberry Pi or ODroid.

PR #1075, opened 2025-09-19
Test long recording for stalls

philburk:
You may want to use an explicit device instead of the default.

This will print the number of seconds every 10 seconds. If it stalls it will probably print an error.

For #930 and #1074


philburk:
Some CMake builds were failing because the function declarations were missing named parameters.

PR #1081, opened 2025-10-03
Fix paqa_devs non-interleaved buffers

### LINKED ISSUE #923: paqa_dev on Raspberry Pi crashes with "segmentation fault" for blocking non-interleaved data
**To Reproduce**
I was running on Raspbian 64-bit OS Lite. But the full OS may also crash.

    sudo apt-get install libasound2-dev
    ./configure && make
    bin/paqa_devs

**Expected behavior**
No crash. 

**Actual behavior**
Lots of errors from paqa_devs.c:395 about "data->framesleft" being some big number.
Then a segmentation fault as soon as it tries to test "BLOCKING, NON-INT".

 - OS: Pi OS Lite V
 - OS Version: 6.6.20
 - PortAudio version: master

philburk:
The buffer needs to be an array of pointers
to each non-interleaved sample array.

Fixes #923

### DIFF
--- a/qa/paqa_devs.c
+++ b/qa/paqa_devs.c
@@ -181,15 +181,38 @@ static void PaQaSetupData(PaQaData *myData,
         int numChannels = (parameters->mode == MODE_OUTPUT)
                 ? parameters->numOutputChannels
                 : parameters->numInputChannels;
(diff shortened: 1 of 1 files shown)


RossBencina (approved):
Looks good. Just add a comment to indicate that audioBuffer can sometimes be a pointer-to-pointers.

PR #840, opened 2023-09-19
Portaudio with Oboe Implementation

hopefulGiupplo:
As I said in [this issue](https://github.com/PortAudio/portaudio/issues/826), I finally can say that this PaOboe implementation seems to be working.
Instruction to how to build PortAudio with Oboe are detailed in src/hostapi/oboe/Readme.md

PR #1084, opened 2025-10-16
feat: add oboe host API

acolombier:
This superseded #840

Note that this patch was previously on top of b0cc303e95fdb7c6c953337051378071c9043e88 ([ref](https://github.com/mixxxdj/vcpkg/blob/befc898b6c2f88c9563503075d517e365d043ca2/overlay/ports/portaudio/portfile.cmake#L4)) but will need to be re-tested on top of on master, once we have finish upgraded our vcpkg to the latest versions.


src/hostapi/oboe/pa_oboe.cpp
```diff
@@ -0,0 +1,1814 @@
...
+    //     if (paOboeStream->hasInput)
+    //         PaUtil_FreeMemory(paOboeStream->inputBuffers[i]);
+    // }
+
+    // if (paOboeStream->hasOutput)
+    //     PaUtil_FreeMemory(paOboeStream->outputBuffers);
+    // if (paOboeStream->hasInput)
+    //     PaUtil_FreeMemory(paOboeStream->inputBuffers);
+
+    // PaUtil_FreeMemory(paOboeStream);
```
acolombier: This remains a blocking issue - Perhaps @philburk you would have an idea of why this approach doesn't currently work and leads to a segfault?

philburk:
@hopefulGiupplo and @acolombier - thank you both for working on this.

We are getting close to finishing 19.8. So I can start looking at this PR in more detail.
Hopefully we can get it merged in 19.9

src/hostapi/oboe/pa_oboe.cpp
philburk: @acolombier wrote:
> is the stream still active shortly after closing?

Ideally no. But these were some system bugs that could cause this.
These have been fixed in the newer versions of Android. But if you are running on older versions of Android then
you may hit these problems. Luckily you can avoid the problems with some careful coding.

This Oboe page has lots of information on this issue:
https://github.com/google/oboe/wiki/TechNote_HowToAvoidCrashes

There is also this related issue:
https://github.com/google/oboe/wiki/CrashUAF_ObtainReleaseBuffer

philburk:
@acolombier - Were any changes made to the code between the end of #840 and the squashing into the 1st commit of this PR?

src/hostapi/oboe/README.md
```diff
@@ -0,0 +1,42 @@
...
+----  
+To build portaudio with Oboe, there are some necessary steps:
+1) An android NDK is needed to crosscompile it. I used the version 25.1.8937393, which I found at
+   https://developer.android.com/ndk/downloads.
+2) Clone the Oboe repository - just follow the steps detailed here:
+   https://github.com/google/oboe/blob/main/docs/GettingStarted.md.
+   Make sure to correctly link the NDK path in the Oboe build. If you instead prefer to use the
+   prebuilt libraries, you can skip this step.
+3) Set the CMake variable OBOE_DIRECTORY (used in cmake/modules/FindOboe.cmake) to the path of the
+   cloned Oboe repository, and build the Oboe libraries (you can use "build_all_android.sh").
```
philburk: Mention that "oboe/build_all_android.sh" is in the Oboe repo.

src/hostapi/oboe/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+   If you instead used the prebuilt libraries, do the following:
+   - set OBOE_DIRECTORY to TRUE;
+   - set OBOE_LIBRARY_DIRS path_to_Oboe_libraries_folder/${ANDROID_ABI}), the code will search the
+     prebuilt library of the chosen ABI in that folder.
+
+4) Build PaOboe (you can use "build_all_PaOboe.sh").
```
philburk: build_all_PaOboe.sh in in #840 but not in this repo!

Should it be?

README.md
```diff
@@ -40,6 +40,7 @@ Please feel free to join. See http://www.portaudio.com for details.
     src/hostapi/coreaudio   = Macintosh Core Audio for OS X
     src/hostapi/dsound      = Windows Direct Sound
     src/hostapi/jack        = JACK Audio Connection Kit
+    src/hostapi/oboe        = Oboe Library for Android (see src/hostapi/oboe/README.md)
```
philburk: This can be a link:

    see [src/hostapi/oboe/README.md](src/hostapi/oboe/README.md)

acolombier:
Hi @philburk , yes, here is the summary
- Refactored the CMake rule to allow more flexibility with existing oboe library (instead of expecting a local clone)
- Refactor oboe-specific port audio methods to allow native device detection and request specific modes (performance, media type, exclusivity, ...) for each open request (instead of global value to drive those
- Remove any code style change (space to tab conversion mainly)
Let me know if you would prefer me to keep the initial commit history.

src/hostapi/oboe/README.md
```diff
@@ -0,0 +1,42 @@
...
+5) Don't forget to add liboboe.so and libportaudio.so in your jniLibs folder.
+
```
philburk, line 21: It would be nice to have instructions for adding this to an Android Studio project.
I think most Android developers are using that.

It might help to add some build.gradle files, which might get imported into Android Studio automatically.

include/pa_oboe.h
```diff
@@ -0,0 +1,154 @@
...
+
+/**
+ * Since Obee doesn't have the ability to detect devices, this helper functions allows the user to register devices, 
+ * previously detected thanks to the JNI interface, since Google is committed to only allow this option to detect devices.
+ * @param name The device name
+ * @param direction The device direction
+ * @param channelCounts The device channel count
+ * @param sampleRate The device default sampleRate
+ */
+void PaOboe_RegisterDevice(const char* name, PaOboe_Direction direction, int channelCounts, int sampleRate);
```
philburk: I notice that PaOboe_RegisterDevice() was added since #840.

Ideally a squashed #840 would have been added as the first commit in its original form.
Then the second commit (or more) would be all the changes made since #840.
Otherwise we have a gap in the commit history.
This is not a critical issue. I am just being picky.

include/pa_oboe.h
```diff
@@ -0,0 +1,154 @@
...
+void PaOboe_SetNumberOfBuffers(unsigned numberOfBuffers);
+
+/**
+ * Since Obee doesn't have the ability to detect devices, this helper functions allows the user to register devices, 
+ * previously detected thanks to the JNI interface, since Google is committed to only allow this option to detect devices.
+ * @param name The device name
+ * @param direction The device direction
+ * @param channelCounts The device channel count
+ * @param sampleRate The device default sampleRate
+ */
```
philburk, line 154: It seems like we would also need a deviceId so that we can select the device using builder.setDeviceId(id). Maybe I am missing something.

src/hostapi/oboe/pa_oboe.cpp
```diff
@@ -0,0 +1,1814 @@
...
+#include <string.h>
+#include "oboe/Oboe.h"
+
+#include <android/log.h>
+#include <android/api-level.h>
+
+#include "pa_oboe.h"
+
+//FIXME: if your project needs a specific PaFormat, modify this value
+#define paOboeDefaultFormat paFloat32
```
philburk, line 76: Good default.

src/hostapi/oboe/pa_oboe.cpp
```diff
@@ -0,0 +1,1814 @@
...
+using namespace oboe;
+
+//Useful global variables
+struct RegisteredDevice {
+    char* name;
+    PaOboe_Direction direction;
+    int channelCount;
+    int sampleRate;
+};
+RegisteredDevice* paOboe_registeredDevices = NULL;
```
philburk: This could be a vector<RegisteredDevice*>.

src/hostapi/oboe/pa_oboe.cpp
```diff
@@ -0,0 +1,1814 @@
...
+
+    //getter and setter of mOboeEngine and mOboeCallbackStream
+    OboeEngine *getEngine() { return mOboeEngine; }
+    void setEngine(OboeEngine *oboeEngine) { mOboeEngine = oboeEngine; }
+
+    PaOboeStream *getStreamAddress() { return mOboeCallbackStream; }
+    void setCallbackStream(PaOboeStream *paOboeStream) { mOboeCallbackStream = paOboeStream; }
+
+    //The only instances of output and input streams that will be used, and their builders
+    std::shared_ptr <AudioStream> mOutputStream;
```
philburk, line 217: Good! Using shared_ptr.

src/hostapi/oboe/pa_oboe.cpp
```diff
@@ -0,0 +1,1814 @@
...
+    //callback utils
+    PaOboeStream *mOboeCallbackStream;
+    unsigned long mFramesProcessed{};
+    PaStreamCallbackTimeInfo mTimeInfo{};
+    struct timespec mTimeSpec{};
+};
+
+
+/**
+ * Stream engine of the host API - Oboe. We allocate only one instance of the engine per PaOboeStream, and
```
philburk: If this is a singleton then you could use the Singleton getInstance() pattern from:
https://github.com/google/oboe/blob/3b02710c2ca69da5358d7d4dd37d46bd7351a474/src/common/Trace.h#L30

philburk (changes requested):
Some general thoughts...

Oboe will support almost any configuration that you request. So you may not need to probe for them. Just advertise, for example, mono, stereo, {8000, 16000, 32000, 44100, 48000}, {float, int16, int24}

You can decide whether to use PerformanceMode=LowLatency  if the suggested latency from PortAudio is < 80(?) msec.

To avoid crashes, do not delete anything right after close.
Just hang onto the memory in a unique_ptr and only free it if you allocate a new one later.

If the app specifies a sample rate, use the Oboe sample rate converter.
You will get much better latency.
    builder->setSampleRateConversionQuality(oboe::SampleRateConversionQuality::Medium)

philburk:
> Remove any code style change (space to tab conversion mainly)

Did you convert spaces to tabs OR tabs to spaces?! I hope the latter.

> Let me know if you would prefer me to keep the initial commit history.

That is up to you. I don't know how hard it would be. And we would lose this partial review if we started a new PR.
I am OK to continue with the PR as is. But you will not get credit for your changes.

I am fine either way. I can just focus on the current state of the code and move forward.

acolombier:
Thanks for all this input and the initial review. I have pushed a new head as #, with 3 commits:
- [0aa7bde](https://github.com/PortAudio/portaudio/pull/1084/commits/0aa7bdee77577d93f61eea0e78f4d3f176705664) - The 100+ commits from [NetResultsIT:Rebase](https://github.com/NetResultsIT/portaudio_oboe/tree/Rebase), on the same head as #840, squashed. ([null diff](https://github.com/NetResultsIT/portaudio_oboe/compare/Rebase..0aa7bde))
- [3527ce3](https://github.com/PortAudio/portaudio/pull/1084/commits/3527ce3a76cb5278a6b3a8a871d082ec3c863a85) -    The merge commit to the latest head I was working from (96dedbd860631f83155141030a07efc3813cfbd5)
- [2e3aad8](https://github.com/PortAudio/portaudio/pull/1084/commits/2e3aad82a30f0c0aa9ef26097d46f84aa41a253b) - My changes, including some of the cleanup related to the first version (thus a fixup, which we could squash) ([null diff](https://github.com/PortAudio/portaudio/compare/2e3aad8..3002310))

Let me know if this is better to review, I can revert otherwise.

> Did you convert spaces to tabs OR tabs to spaces?! I hope the latter.

I just reverted to what was used, in order to limit the diff - initial version add 310 line changes, including many that were unrelated to oboe and seem to be random style updates (capitalisation number of spaces/tabs, ..). Regarding the used style, I try to imitate as much what I could see in other places of PortAudio, but feel free to point inconsistency and I will update.

I will start addressing your feedback asap!

src/hostapi/oboe/README.md
acolombier: Ah no indeed! It's part of the elements I have removed. Now, we have a more standard CMake that should simply leverage existing oboe header and library from the configured prefix

src/hostapi/oboe/README.md
acolombier: Indeed, I will update these instructions

src/hostapi/oboe/README.md
acolombier: Unfortunately, I don't use Android Studio and don't have any experience with it - I use PortAudio+oboe in Qt, using a CMake setup. 
WOuld you be able to share some basic build.gradle?

acolombier:
I have pushed the first batch of changes, including some DRY cleanup and better concurrency/allocation handling.
I have changed the stream close free to happen in the destructor and leveraging a shared_ptr to delay the freeing of resources.

I still need to test that!

acolombier:
Tested and could not reproduce any crash! 
Though I couldn't confirm if it also gets deallocated correctly.

src/hostapi/oboe/pa_oboe.cpp
```diff
@@ -0,0 +1,1783 @@
...
+
+    PaUtil_ResetBufferProcessor(&paOboeStream->bufferProcessor);
+
+    //Checking if the stream(s) are already active.
+    //TODO: check if it's working as expected (extensive testing needed, no problem spotted with situational tests)
+    if (paOboeStream->isActive) {
+        LOGW("[PaOboe - StartStream]\t Stream was already active, stopping...");
+        StopStream(paStream);
+        LOGW("[PaOboe - StartStream]\t Restarting...");
+        StartStream(paStream);
```
wax-m, line 1524: Maybe i am missing something, but this seems to start the stream twice?

acolombier: The stream should be stopped just above but perhaps it failed to stop it? Could you share the logs?

wax-m: Hey 
i did not actually use it yet, i was just reading through the code to figure out how it is structured, and in this place in the code i got confused, because it appears to me, that in the case that the stream is already running, it will be stopped, then StartStream will be called, then the rest of the original StartStream will be executed.

in my next step i will try to get it to run using android-studio, maybe i will be able to supply a gradle file afterwards

wax-m:
Hey, I got this PR to run in an android-studio project, and i was wondering, if there are any reasons not to include the JNI workflow to get the audiodevices directly in the library?
I did it with plain JNI, not the JNI wrapped by QT for Android

acolombier:
JNI sounded like a major dependency to add to portaudio at first, but I guess it could later be added as optional dependency?

PR #1085, opened 2025-10-17
cmake: change logic for PA_BUILD_SHARED_LIBS to default ON

### LINKED ISSUE #918: CMake regression?: default for PA_BUILD_SHARED_LIBS flipped from ON to OFF
Regarding PR #765 ("Remove usage of global BUILD_SHARED_LIBS variable" by @bheisch, merged April 1, 2023), @dechamps wrote:

> FYI it looks like this patch caused the default for PA_BUILD_SHARED_LIBS to flip from ON to OFF, which surprised me - I ended up accidentally releasing a version of my project with PortAudio statically linked instead of bundling it as a separate DLL. (See https://github.com/dechamps/FlexASIO/commit/42e683b14b1e6398d4ceb5d4558b0d6da03ad73e.) Thankfully that's benign in my case, but it might not be for others. Was the change of default value intended?

philburk:
If PA_BUILD_SHARED_LIBS is already defined then keep the original value. If not defined, and if BUILD_SHARED_LIBS is defined then use that value.
    Otherwise set PA_BUILD_SHARED_LIBS=ON.

The previous default was OFF.
In PortAudio v19.7 the default was ON.

Fixes #918

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,6 +5,10 @@ project(PortAudio VERSION 19.8)
 # General PortAudio stuff
 #
 
+# Policy: By default, build PortAudio as a shared library.
+# There are two ways to override the default behavior:
+# 1. User can select shared or static PortAudio library build by explicitly defining PA_BUILD_SHARED_LIBS to ON or OFF (e.g. on the command line or an importing project)
+# 2. If PA_BUILD_SHARED_LIBS is not set we respect the standard CMake BUILD_SHARED_LIBS variable if it is set.
 if (NOT DEFINED PA_BUILD_SHARED_LIBS)
     if (DEFINED BUILD_SHARED_LIBS)
         set(PA_BUILD_SHARED_LIBS ${BUILD_SHARED_LIBS})


philburk:
Hmm the Windows CMake builds are failing with:
```
C:/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/14.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: D:/a/portaudio/portaudio/qa/paqa_dither.c:164:(.text+0xb2f): undefined reference to `PaUtil_Generate16BitTriangularDither'
```

The paqa_dither test uses internal functions like PaUtil_InitializeTriangularDitherState () and PaUtil_Generate16BitTriangularDither(). The autoconf build has special targets to handle that.
But the CMake build fails on Windows when using shared (dynamic) libraries. (Thank you CI.)

RossBencina:
I believe that the new logic in this PR achieves the desired behavior. However, I'm not sure whether it is correct CMake usage to declare an option conditionally. If there is a way to do it while keeping the option declaration global that would be preferable. CMake experts: what do you think?

RossBencina:
How about this:
```
# Default building shared PortAudio lib to ON, but allow
# both BUILD_SHARED_LIBS and PA_BUILD_SHARED_LIBS to override if explicitly defined,
# with PA_BUILD_SHARED_LIBS taking precedence
set(_DEFAULT_PA_BUILD_SHARED_LIBS ON)
if (DEFINED BUILD_SHARED_LIBS)
    set(_DEFAULT_PA_BUILD_SHARED_LIBS ${BUILD_SHARED_LIBS})
endif()
option(PA_BUILD_SHARED_LIBS "Build shared library" _DEFAULT_PA_BUILD_SHARED_LIBS )
```

generic-pers0n:
> I believe that the new logic in this PR achieves the desired behavior. However, I'm not sure whether it is correct CMake usage to declare an option conditionally. If there is a way to do it while keeping the option declaration global that would be preferable. CMake experts: what do you think?

On my end, I don't find there to be any issues with declaring CMake options conditionally. At least, I see this done in Tenacity's CMakeLists.txt [here](https://codeberg.org/tenacityteam/tenacity/src/commit/d3a9aa39d1c78836dbd5d6f84c8a88aea5058ca6/CMakeLists.txt#L723) and I don't think there's any issue with it.

For future reference, and not necessarily relevant for this issue, you may also be interested in `cmake_dependent_option()`, which automatically forces a value for a build option if a specified condition evaluates to false.

generic-pers0n (commented):
Aside from the aforementioned build issues, this looks good to me!

philburk:
I think this is OK to merge after #1099 gets merged.

CMakeLists.txt
```diff
@@ -5,14 +5,23 @@ project(PortAudio VERSION 19.8)
 # General PortAudio stuff
 #
 
-option(PA_BUILD_SHARED_LIBS "Build dynamic library" ${BUILD_SHARED_LIBS})
+if (NOT DEFINED PA_BUILD_SHARED_LIBS)
```
RossBencina, line 12:
```suggestion
# Policy: By default, build PortAudio as a shared library.
# There are two ways to override the default behavior:
# 1. User can select shared or static PortAudio library build by explicitly defining PA_BUILD_SHARED_LIBS to ON or OFF (e.g. on the command line or an importing project).
# 2. If PA_BUILD_SHARED_LIBS is not set we respect the standard CMake BUILD_SHARED_LIBS variable if it is set.
if (NOT DEFINED PA_BUILD_SHARED_LIBS)
```

PR #1087, opened 2025-10-19
Localize declaration of for loop variable

gruenich:
Declaration at same place at use.
Minimize variable scope.
Prevent unintended use outside for loop.

### DIFF
--- a/src/hostapi/jack/pa_jack.c
+++ b/src/hostapi/jack/pa_jack.c
@@ -568,7 +568,7 @@ static PaError BuildDeviceList( PaJackHostApiRepresentation *jackApi )
         tmp_client_name[match_info.rm_eo - match_info.rm_so] = '\0';
 
         /* do we know about this port's client yet? */
-        for( unsigned int i = 0; i < numClients; i++ )
+        for( unsigned long i = 0; i < numClients; i++ )
         {
             if( strcmp( tmp_client_name, client_names[i] ) == 0 )
                 client_seen = TRUE;


gruenich:
There are still cases left where `i` is declared outside the `for` loop. If you are interested, I can complete this pull request before merging. Further, I can create additional pull requests for other variables besides `i` and other localizations of variables.

dmitrykos:
This change will give a compilation error if C compiler is configured strictly for C89/C90. It will work though with C99 and up.

gruenich:
In reality, portaudio cannot be compiled in strict C89 mode.
* src/common/pa_debugprint.c:58:1: C++-style (`//`) comments are not allowed
* src/common/pa_trace.h:106:39: variadic macros are a C99 feature
* src/hostapi/pulseaudio/pa_linux_pulseaudio.c:82:24: mixing declarations and code is a GNU89 / C99 extension
* the included headers from pulse use `inline` which is not known in C89.

Before this change, it can be compiled with `-std=gnu89`. After the change, it requires `c99` or `gnu99`.

I guess we have to wait with this issue to have resolved #606 first. The current milestone is v20.

RossBencina:
Our current policy is C89 for maximum portability, so this PR is not appropriate at this time. We do intend to review the policy after the v19.8 release. Most likely we will still require the core common infrastructure to be C89.

illuusio:
> * src/hostapi/pulseaudio/pa_linux_pulseaudio.c:82:24: mixing declarations and code is a GNU89 / C99 extension

As this can be easily fixed but it seems that one can't compile Pulseaudio code with `-std=c89` as headers are GNU89 / ANSI C compatible. Should this be raised to ANSI C (as known as C90?). It's still very compatible backwards..

gruenich:
@illuusio C89 and C90 is the same standard, one by ANSI the other by ISO. The next increment is C99 with errors corrected C11.

RossBencina:
@illuusio in the host API code, it is OK to use whatever C version you need for PulseAudio to work.

RossBencina:
@gruenich and pa_asio.cpp is C++ because ASIO is C++.

The basic policy is that the code should be as portable as possible and that means C89 where-ever possible (most significantly this is important for MSVC which historically never supported C99 and up, now maybe they do, but I still don't want to break backwards compatibility.)

> src/common/pa_debugprint.c:58:1: C++-style (//) comments are not allowed

Ideally that should be fixed.

> src/common/pa_trace.h:106:39: variadic macros are a C99 feature

Presumably MSVC always supported that.

In any case if you want to discuss it further I think we should move discussion to #606

RossBencina:
> src/common/pa_trace.h:106:39: variadic macros are a C99 feature

Proposed fix: #1132

gruenich:
Portaudio will remain compatible to C89, so this won't be merged.

RossBencina:
Reopening. My view is that that /hostapi portion of this PR should still be considered for merge. The work has been done and we shouldn't throw it away. Pending a final policy decision I think we're going to end up allowing language features, C99, and , in particular, declarations in for loop head for host APIs, where that language feature is supported by all compilers within the support window. I'm pretty sure MacOS and Linux would have supported C99 for all versions of coreaudio, ALSA, JACK that we support (maybe not ALSA and JACK), and I also think old versions of MSVC supported declaration of variable in for loop head.

RossBencina:
Phil will merge the host api changes. We will keep /common and /qa C89

gruenich:
I adjusted the changes accordingly, should be fine for merging.

philburk:
@RossBencina - feel free to merge this after you approve.

src/hostapi/jack/pa_jack.c
```diff
@@ -568,7 +568,7 @@ static PaError BuildDeviceList( PaJackHostApiRepresentation *jackApi )
         tmp_client_name[match_info.rm_eo - match_info.rm_so] = '\0';
 
         /* do we know about this port's client yet? */
-        for( i = 0; i < numClients; i++ )
+        for( unsigned int i = 0; i < numClients; i++ )
```
RossBencina:
```suggestion
        for( unsigned long i = 0; i < numClients; i++ )
```

PR #1089, opened 2025-10-23
Remove non-functional SGI/Irix host-API references

RossBencina:
SGI/Irix was never ported to the v19 API, but there was still space for it in the host API configuration. This PR removes the remaining references to SGI/Irix from the codebase.


src/os/unix/pa_unix_hostapis.c
```diff
@@ -66,10 +64,6 @@ PaUtilHostApiInitializer *paHostApiInitializers[] =
         PaAlsa_Initialize,
 #endif
 
-#ifdef PA_USE_SNDIO
-        PaSndio_Initialize,
```
philburk, line 70: @RossBencina - did you mean to remove sndio? The declaration is still on line 48.

src/os/unix/pa_unix_hostapis.c
```diff
@@ -66,10 +64,6 @@ PaUtilHostApiInitializer *paHostApiInitializers[] =
         PaAlsa_Initialize,
 #endif
 
-#ifdef PA_USE_SNDIO
```
RossBencina, line 69: This shouldn't have been deleted. Will fix.

RossBencina:
@philburk this is ready to merge now. It changes configure.ac so you may want to merge it prior to running autoreconf (or maybe not).

src/os/unix/pa_unix_hostapis.c
philburk: Thanks for restoring this.

PR #1090, opened 2025-10-24
PulseAudio API: Swap name and description fields

### LINKED ISSUE #1088: PulseAudio backend: sink info uses pulseaudio description instead of name.
**Describe the bug**
when use any python wrapper  for portaudio and enumerate devices.
field "name", actually contains "description".

It makes impossible to pass this field to external tools, like paplay, pw-play and other.

It looks like this lines must not be used:
https://github.com/PortAudio/portaudio/blob/master/src/hostapi/pulseaudio/pa_linux_pulseaudio.c#L393

**To Reproduce**
1.  use any python wrapper. For example "sounddevice"

```
import sounddevice
print(sounddevice.queiry_devices())
```

output:
```
  0 Default Sink, PulseAudio (0 in, 32 out)
>  1 Default Source, PulseAudio (32 in, 0 out)
   2 This is description, PulseAudio (0 in, 6 out)
   3 JieLi BR21 Analog Stereo, PulseAudio (0 in, 2 out)
```

check in `pactl list sinks`

```
Sink #39
        State: SUSPENDED
        Name: snd_room_1
        Description: This is description
        Driver: PipeWire
```

For sink 39, field Description "This is description" is shown as "name" in portaudio.

**Expected behavior**
"Name" field contains "name" from pulseaudio, not 'description'
 
**Actual behavior**
"Name" field contains "description" from pulseaudio, not 'name'

 - OS: debian 13 trixie
 - OS 13
 - PortAudio version: stable, from official debian repository.
 - API: PulseAudio API

dibalavs:
This patch fixes #1088 issue when
'name' field contains description message instead of real name.

### DIFF
--- a/src/hostapi/pulseaudio/pa_linux_pulseaudio.c
+++ b/src/hostapi/pulseaudio/pa_linux_pulseaudio.c
@@ -371,7 +371,7 @@ void PaPulseAudio_SinkListCb( pa_context * c,
 {
     PaPulseAudio_HostApiRepresentation *pulseaudioHostApi =
         (PaPulseAudio_HostApiRepresentation *) userdata;
-    const char *pulseaudioDeviceName = NULL;
+    const char *pulseaudioDeviceDescription = NULL;
 
 
     /* If this is null we have big problems and we probably are out of memory */
@@ -388,16 +388,16 @@ void PaPulseAudio_SinkListCb( pa_context * c,
         goto error;
     }
 
-    pulseaudioDeviceName = l->name;
+    pulseaudioDeviceDescription = l->name;
 
     if( l->description != NULL )
     {
-        pulseaudioDeviceName = l->description;
+        pulseaudioDeviceDescription = l->description;
     }
 
     if( _PaPulseAudio_AddAudioDevice( pulseaudioHostApi,
-                                      pulseaudioDeviceName,
                                       l->name,
+                                      pulseaudioDeviceDescription,
                                       0,
                                       l->sample_spec.channels,
                                       0,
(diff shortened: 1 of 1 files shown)


illuusio:
As I was starting to remember why I did it in first place. Actually I was thinking users. As name `alsa_output.pci-0000_04_00.6.analog-stereo` in name list says nothing for example Mixxx user but `Family 17h/19h/1ah HD Audio Controller Analoginen stereo`  says at least something.
Which is better device name of course is always correct and should be used but it's not so convenient? PR itself is okay by me I just want to discuss this one here.

dibalavs:
> s name alsa_output.pci-0000_04_00.6.analog-stereo in name list says nothing for example Mixxx user but Family 17h/19h/1ah HD Audio Controller Analoginen stereo says at least something.

I do not agree with it. Name is not designed to be seen by humans.
Name must be:
1) Unique. If we have multiple sinks, name must point to concrete instance of concrete device
2) Name must be used as parameter to external sound tools (`paplay --device=alsa_output.pci-0000_04_00.6.analog-stereo`  - will work, `paplay --device="Family 17h/19h/1ah HD Audio Controller Analoginen stereo"` will not work)

Humans must see 'Description' field instead.

The reason, because I file a bug, is I have 3 absolutely the same USB audio cards And all of them have description field is:

`Description: JieLi BR21 Analog Stereo`

I have no any way to find concrete instance if card. 
Also pyaudio (wrapper over portaduio) does not have access to `description` field. So in buggy case, When I enumerate audio cards, I receive absolutely the same names of all cards, And I am unable to use this name in external tools, like paplay.

RossBencina:
@dibalavs please update this PR now as-per Phil's request:

> we should change pulseaudioDeviceName to
> pulseaudioDeviceDescription so that the code is more clear.

I do not want to merge a half-fix.

RossBencina:
Thank you!

Issue #1094, opened 2025-11-21
Windows CMake builds are failing for paqa_dither with shared libraries


philburk:
Windows CMake builds are failing with:
```
C:/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/14.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: D:/a/portaudio/portaudio/qa/paqa_dither.c:164:(.text+0xb2f): undefined reference to `PaUtil_Generate16BitTriangularDither'
```

The paqa_dither test uses internal functions like PaUtil_InitializeTriangularDitherState () and PaUtil_Generate16BitTriangularDither(). The autoconf build has special targets to handle that.
But the CMake build fails on Windows when using shared (dynamic) libraries. (Thank you CI.)

_Originally posted by @philburk in https://github.com/PortAudio/portaudio/issues/1085#issuecomment-3413423064_

philburk:
I notice that CMakeLists.txt has:
```
if(LINK_PRIVATE_SYMBOLS)
  add_test(patest_converters)
endif()
add_test(patest_dither)
```

PR #1095, opened 2025-11-22
Add CI that builds both STATIC and SHARED libraries

### LINKED ISSUE #1093: CMake: CI does not test building shared libs
apparently building shared lib is broken, but we did not know because CI doesn't build it.

see recent discussion on #1085
phil to make a ticket about the breakage

philburk:
Clean the build after the STATIC build.

Fixes #1093

### DIFF
--- a/qa/CMakeLists.txt
+++ b/qa/CMakeLists.txt
@@ -1,8 +1,6 @@
 add_test(paqa_errs)
 add_test(paqa_devs)
-if(LINK_PRIVATE_SYMBOLS)
-  add_test(paqa_dither)
-endif()
+add_test(paqa_dither)
 add_test(paqa_latency)
 
 subdirs(loopback)


philburk:
The MinGW build for STATIC is passing:
    
    [ 78%] Linking CXX executable paqa_dither.exe

But the SHARED build is failing with:
```
[ 78%] Linking C executable paqa_dither.exe
[ 78%] Built target patest_enumerate_default_latency
[ 79%] Building C object qa/CMakeFiles/paqa_latency.dir/paqa_latency.c.obj
[ 80%] Building C object qa/loopback/CMakeFiles/paloopback.dir/src/audio_analyzer.c.obj
C:/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/15.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: CMakeFiles\paqa_dither.dir/objects.a(paqa_dither.c.obj): in function `MeasureAverageConversion':
D:/a/portaudio/portaudio/qa/paqa_dither.c:204:(.text+0x8e): undefined reference to `PaUtil_InitializeTriangularDitherState'
```

So it seems the CI is building with both STATIC and SHARED and the SHARED build is failing. Yay!

philburk:
I tested the fix that is now in #1096 and the CI passed.
So the CI change seems good and the fix seems good.
I then removed the fix so that reviewers can see that the CI is failing, and thus catching the bug in #1094.

RossBencina:
I think you could do this with a single build step with the matrix like this:

```
build:
  strategy:
    matrix:
      PA_BUILD_SHARED_LIBS: [ON, OFF]
      
...

    - name: Configure PortAudio library
      run: cmake

...

      -DPA_BUILD_SHARED_LIBS=${{matrix.PA_BUILD_SHARED_LIBS}}
```

philburk:
Use #1099 instead.

PR #1097, opened 2025-11-22
PulseAudio: Fix device name handling for sinks and sources

illuusio:
PR #1090, which changed how device names are handled, broke the process of selecting the correct device when connecting to playback sinks or recording sources.

This commit addresses this issue by ensuring proper device selection during connection.


RossBencina:
We're going to merge this PR, but it looks like maybe `pulseaudioHostApi->pulseaudioDeviceNames` could be removed completely and just use `deviceInfoArray[stream->inputDevice].name` everywhere.

RossBencina (approved):

philburk (approved):

PR #1099, opened 2025-12-04
Use matrix in CMake CI to build shared libs

### LINKED ISSUE #1093: CMake: CI does not test building shared libs

philburk:
Fixes #1093

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -7,7 +7,6 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # os: [ubuntu-latest, windows-latest, macOS-latest]
         name: [Ubuntu-GCC, Ubuntu-MinGW, Windows-MSVC, Windows-MinGW, macOS-Clang, macOS-Clang-framework]
         shared_libs: [ON, OFF]
         include:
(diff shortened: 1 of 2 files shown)


philburk:
When I add the line:

    pa_shared_libs: [ON, OFF]

then I lose the CMake Windows MINGW and MSVC build and the Ubuntu builds.
The number of tests goes from 24 to 16!

philburk:
Moving the pa_build_shared declaration after the include: block did not help.
Moving the MacOS blocks to the top made no difference.

philburk:
Copilot figured it out. I had to declare each OS option at the top of the matrix.

philburk:
I renabled the paqa_dither build for shared libs and it failed:

C:/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/15.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: CMakeFiles\paqa_dither.dir/objects.a(paqa_dither.c.obj): in function `MeasureAverageConversion':
D:/a/portaudio/portaudio/qa/paqa_dither.c:204:(.text+0x8e): undefined reference to `PaUtil_InitializeTriangularDitherState'

So we are solving the issue in #1093

philburk:
Why do we only have two failed builds here but 4 failed builds on #1095.
In #1095 we also failed on Windows MSVC.
This PR is missing:
* CMake Windows MSVC [pull_request, push]
* CMake Ubuntu GGCC [pull_request, push]

RossBencina (approved):

PR #1104, opened 2025-12-14
pulseaudio: Remove missedBytes variable as it's not used anywhere

illuusio:
In Pulseaudio there is formal callback when buffer under runs. Rely on callback and remove not needed missedBytes variable.


bdrung:
This commit caused a regression for me: https://github.com/PortAudio/portaudio/issues/1119

PR #1107, opened 2025-12-18
[Windows] Fix Windows ARM64 compilation

### LINKED ISSUE #1101: Issue compiling for Windows ARM64
**Describe the bug**
Cannot compile on Windows ARM64.

**To Reproduce**
1. Compile with an ARM64 toolchain for Windows.

**Expected behavior**
It compiles.

**Actual behavior**
The issue lies in the *MemoryBarrier macros. 

```
lld-link: error: undefined symbol: _ReadBarrier
>>> referenced by /vcpkg/buildtrees/portaudio/src/dfffbcfbb6-442ea863eb.clean/src/hostapi/wdmks/pa_win_wdmks.c:682
>>>               portaudio.lib(pa_win_wdmks.c.obj):(MemoryBarrierRead)
>>> referenced by /vcpkg/buildtrees/portaudio/src/dfffbcfbb6-442ea863eb.clean/src/common/pa_ringbuffer.c:181
>>>               portaudio.lib(pa_ringbuffer.c.obj):(PaUtil_GetRingBufferReadRegions)
>>> referenced by /vcpkg/buildtrees/portaudio/src/dfffbcfbb6-442ea863eb.clean/src/common/pa_ringbuffer.c:224
>>>               portaudio.lib(pa_ringbuffer.c.obj):(PaUtil_ReadRingBuffer)

lld-link: error: undefined symbol: _WriteBarrier
>>> referenced by /vcpkg/buildtrees/portaudio/src/dfffbcfbb6-442ea863eb.clean/src/hostapi/wdmks/pa_win_wdmks.c:687
>>>               portaudio.lib(pa_win_wdmks.c.obj):(MemoryBarrierWrite)
>>> referenced by /vcpkg/buildtrees/portaudio/src/dfffbcfbb6-442ea863eb.clean/src/common/pa_ringbuffer.c:144
>>>               portaudio.lib(pa_ringbuffer.c.obj):(PaUtil_AdvanceRingBufferWriteIndex)
>>> referenced by /vcpkg/buildtrees/portaudio/src/dfffbcfbb6-442ea863eb.clean/src/common/pa_ringbuffer.c:214
>>>               portaudio.lib(pa_ringbuffer.c.obj):(PaUtil_WriteRingBuffer)
```

 - OS: Windows 10
 - OS Version 19045.6456
 - PortAudio version: 147dd722548358763a8b649b3e4b41dfffbcfbb6 (vcpkg)

**Additional context**
I'm just preparing some pipeline for Windows ARM64 support, note that I can't debug, test or use it, but at least, the issue has already been seen for future reference.

Here is a potential fix: https://github.com/PortAudio/portaudio/blob/master/src/common/pa_memorybarrier.h#L112-L119
```
#   include <intrin.h>
#   if defined(_M_IX86) || defined(_M_X64)
#       pragma intrinsic(_ReadWriteBarrier)
#       pragma intrinsic(_ReadBarrier)
#       pragma intrinsic(_WriteBarrier)
#       define PaUtil_FullMemoryBarrier()  _ReadWriteBarrier()
#       define PaUtil_ReadMemoryBarrier()  _ReadBarrier()
#       define PaUtil_WriteMemoryBarrier() _WriteBarrier()
#   elif defined(_M_ARM64)
#       define PaUtil_FullMemoryBarrier()  __dmb(_ARM64_BARRIER_ISH)
#       define PaUtil_ReadMemoryBarrier()  __dmb(_ARM64_BARRIER_ISHLD)
#       define PaUtil_WriteMemoryBarrier() __dmb(_ARM64_BARRIER_ISHST)
#   else
#       error Unsupported MSVC architecture
#   endif
```

Microsoft reference:
https://learn.microsoft.com/en-us/cpp/intrinsics/readbarrier?view=msvc-170
https://learn.microsoft.com/en-us/cpp/intrinsics/arm64-intrinsics?view=msvc-170

Nemirtingas:
Add ARM64 memory barrier definitions for MSVC.

Fixes #1101

### DIFF
--- a/src/common/pa_memorybarrier.h
+++ b/src/common/pa_memorybarrier.h
@@ -101,7 +101,12 @@
 #elif (_MSC_VER >= 1400) && !defined(_WIN32_WCE)
 #   include <intrin.h>
 /* note that MSVC intrinsics _ReadWriteBarrier(), _ReadBarrier(), _WriteBarrier() are just compiler barriers *not* memory barriers */
-#   if defined(_M_IX86) || defined(_M_X64)
+#   if defined(_M_ARM64)
+        /* https://learn.microsoft.com/en-us/cpp/intrinsics/arm64-intrinsics?view=msvc-170 */
+#       define PaUtil_FullMemoryBarrier()  __dmb(_ARM64_BARRIER_ISH)
+#       define PaUtil_ReadMemoryBarrier()  __dmb(_ARM64_BARRIER_ISHLD)
+#       define PaUtil_WriteMemoryBarrier() __dmb(_ARM64_BARRIER_ISHST)
+#   else
 #       pragma intrinsic(_ReadWriteBarrier)
 #       pragma intrinsic(_ReadBarrier)
 #       pragma intrinsic(_WriteBarrier)
@@ -109,13 +114,6 @@
 #       define PaUtil_FullMemoryBarrier()  _ReadWriteBarrier()
 #       define PaUtil_ReadMemoryBarrier()  _ReadBarrier()
 #       define PaUtil_WriteMemoryBarrier() _WriteBarrier()
-#   elif defined(_M_ARM64)
-        /* https://learn.microsoft.com/en-us/cpp/intrinsics/arm64-intrinsics?view=msvc-170 */
-#       define PaUtil_FullMemoryBarrier()  __dmb(_ARM64_BARRIER_ISH)
-#       define PaUtil_ReadMemoryBarrier()  __dmb(_ARM64_BARRIER_ISHLD)
-#       define PaUtil_WriteMemoryBarrier() __dmb(_ARM64_BARRIER_ISHST)
-#   else
-#       error Unsupported MSVC architecture
 #   endif
 #elif defined(_WIN32_WCE)
 #   define PaUtil_FullMemoryBarrier()


src/common/pa_memorybarrier.h
```diff
@@ -100,6 +100,7 @@
 #   endif
 #elif (_MSC_VER >= 1400) && !defined(_WIN32_WCE)
 #   include <intrin.h>
+/* note that MSVC intrinsics _ReadWriteBarrier(), _ReadBarrier(), _WriteBarrier() are just compiler barriers *not* memory barriers */
```
RossBencina, line 103: Right, so does your proposal also behave as a compiler barrier? That is important.

src/common/pa_memorybarrier.h
```diff
@@ -100,13 +100,22 @@
...
 #elif (_MSC_VER >= 1400) && !defined(_WIN32_WCE)
 #   include <intrin.h>
-#   pragma intrinsic(_ReadWriteBarrier)
-#   pragma intrinsic(_ReadBarrier)
-#   pragma intrinsic(_WriteBarrier)
-/* note that MSVC intrinsics _ReadWriteBarrier(), _ReadBarrier(), _WriteBarrier() are just compiler barriers *not* memory barriers */
-#   define PaUtil_FullMemoryBarrier()  _ReadWriteBarrier()
-#   define PaUtil_ReadMemoryBarrier()  _ReadBarrier()
-#   define PaUtil_WriteMemoryBarrier() _WriteBarrier()
+#   if defined(_M_IX86) || defined(_M_X64)
```
philburk: I am worried about the availability of _M_I_X86 on very old platforms, pre-2008.
That might cause a build failure regression.

I think a safer approach would be:

```
#    if defined(_M_ARM64)
         ... new ARM code ...
#    else
        ... old X86 code ...
 #    endif
```

philburk (changes requested):
Thanks for the suggestion and the quick response.

src/common/pa_memorybarrier.h
Nemirtingas: According to MSVC documentation, __dmb stands for Data Memory Barrier.

```
The intrinsic functions
__dmb (data memory barrier)
__dsb (data synchronization barrier)
__isb (instruction synchronization barrier)

use the following predefined values to specify the memory barrier restriction in terms of the sharing domain and the kind of access that are affected by the operation.
```

I know very little on the why and how thoses barriers are used or needed. But according to LLMs, on ARM64 (which seems to reorder aggressively). Here's its answer if its of some use:

```
On ARM64, a compiler barrier is not sufficient because the CPU itself may reorder memory accesses.
Therefore we need a hardware memory barrier.

__dmb() (Data Memory Barrier) is the correct choice because it enforces ordering of memory accesses between cores, without unnecessarily stalling the pipeline.

DMB guarantees that memory operations before the barrier are observed before those after it.

It is designed for thread synchronization and lock-free code.

It is significantly lighter than DSB, which waits for all memory operations to fully complete and is intended for MMIO or kernel-level operations.

ISB is unrelated to memory ordering and only synchronizes instruction execution.

Using __dmb(ISH / ISHLD / ISHST) provides the minimum required ordering guarantees on ARM while preserving performance, making it the correct equivalent when moving from MSVC compiler barriers to ARM64.
```

src/common/pa_memorybarrier.h
Nemirtingas: Done

src/common/pa_memorybarrier.h
RossBencina: My question stands and I have not been able to find an answer, but given that the intrinsics are intended for use as memory barriers I am going to assume that they are implicitly compiler barriers.

RossBencina (approved):

RossBencina:
Thank you!

PR #1112, opened 2025-12-26
pa_asio.cpp: Declare loop variable in for loop

gruenich:
C++ allows this.
Limits the scope of the loop variable.


RossBencina:
Thank you.

PR #1114, opened 2026-01-01
WDM-KS: fix some inconsistent logs

LastExceed:
Hi, I'm new here. Found some inconsistencies in the logging while debugging. The commits should be pretty self-explanatory.

(Also maiden C/C++ PR for me, yay)

### DIFF
--- a/src/hostapi/wdmks/pa_win_wdmks.c
+++ b/src/hostapi/wdmks/pa_win_wdmks.c
@@ -3301,6 +3301,7 @@ PaWinWdmFilter** BuildFilterList( int* pFilterCount, int* pNoOfPaDevices, PaErro
     *pFilterCount = filterCount;
     *pNoOfPaDevices = noOfPaDevices;
 
+    PA_LOGL_;
     return ppFilters;
 }
 


philburk (approved):

RossBencina:
@LastExceed thank you and welcome! It looks like there are similar issues with other logs in that file if you feel like fixing them (missing `\n`, missing `PA_LOGL_` before `return`, e.g. line 2938), no pressure :). If you're going to work with PA/WDM-KS I recommend looking at open issues tagged with `src-wdmks` there is at least one known bug that hasn't been fixed because no one has tested the fix. https://github.com/PortAudio/portaudio/issues?q=is%3Aissue%20state%3Aopen%20label%3Asrc-wdmks

LastExceed:
I've been using this file as a learning reference (information on ks audio seems to be extremely scarce, this is literally the only open source code i could find about it) and I'm basically at the hello-world stage where I got a sinewave going in a scratch project, with all settings hardcoded and zero error handling. so not sure if I'll be of much help 😅 but I'll take a look, I can probably find those other logging bugs at least

Issue #1119, opened 2026-01-07
Recording audio with Audacity using PulseAudio does not work


bdrung:
**Describe the bug**
The audio signal is 100% and Audacity stops reacting when trying to record audio.

**To Reproduce**
Steps to reproduce the behavior. Include code if applicable.
1. Open Audacity
2. Set PulseAudio as input
3. Press record

**Expected behavior**
Normal recording

**Actual behavior**
The audio signal is 100% and Audacity stops reacting when trying to record audio.

**Desktop (please complete the following information):**
 - OS: Ubuntu
 - OS Version 24.10
 - PortAudio version: current, commit adbf810825252d0c23c2223626b327425d05be3b

**Additional context**
I bisected this issue to commit 7966aacfab8c878a8fd968f67145c4d319e9f69e.

illuusio:
Ok. Seems bit strange should not do this. I'll test this.

illuusio:
Ok could not verify this on Audacity currently but could you test this pull request PortAudio/portaudio#1120? I'll examine
 more on this is that is not fixing situation and thanks for reporting this.

PR #1120, opened 2026-01-07
pulseaudio: Rework enough bytes written bytes or not enough bytes to read

illuusio:
Rework we have written enough bytes to output or we don't have enough bytes to read in pulseaudio. This should fix issue #1119

### DIFF
--- a/src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
+++ b/src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
@@ -365,6 +365,17 @@ static int _PaPulseAudio_ProcessAudio(PaPulseAudio_Stream *stream,
          */
         PA_PULSEAUDIO_IS_ERROR( stream, paStreamIsStopped )
 
+        /* If we have written enough bytes (If output is enabled)
+         * or there is no enough bytes to be for input
+         * break out from loop
+         */
+        if( (isInputCb &&
+            PaUtil_GetRingBufferReadAvailable(&stream->inputRing) < pulseaudioInputBytes) ||
+            pulseaudioOutputWritten >= length )
+        {
+            break;
+        }
+
         if(  stream->outputStream )
         {
             PaPulseAudio_updateTimeInfo( stream->outputStream,


bdrung:
Tested and that fixes the issue #1119 for me.

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
```diff
@@ -365,6 +365,17 @@ static int _PaPulseAudio_ProcessAudio(PaPulseAudio_Stream *stream,
          */
         PA_PULSEAUDIO_IS_ERROR( stream, paStreamIsStopped )
 
+        /* If we have written enough bytes (If output if enabled)
```
bdrung:
```suggestion
        /* If we have written enough bytes (If output is enabled)
```

src/hostapi/pulseaudio/pa_linux_pulseaudio_cb.c
```diff
@@ -365,6 +365,17 @@ static int _PaPulseAudio_ProcessAudio(PaPulseAudio_Stream *stream,
...
+         * or there is no enouhg bytes to be for input
```
bdrung:
```suggestion
         * or there is no enough bytes to be for input
```

illuusio:
Thank you for looking into this. I've rebased your changes to one commit.

bdrung:
You can squash the typo fixup commit into your initial commit and there is no need for mentioning me as co-author for such trivial change.

illuusio:
If it's ok to you I've just updated PR

RossBencina:
@illuusio wrote:
> If it's ok to you I've just updated PR

Adding to a PR is fine but please do not force-push to non-draft PRs.

RossBencina (approved):
We are going to merge this as an emergency fix that reverts to previous known-better behavior of 7966aacfab8c878a8fd968f67145c4d319e9f69e. However, under normal circumstances:

- The comment is unacceptably bad. It is unclear whether it adds anything. We seriously considered deleting it but for the sake of keeping history we have retained it.

- The logic does not look correct. We have opened issue #1128 in which we will shortly add more detail.

PR #1133, opened 2026-02-20
convert C++ comments to C comments in pa_debugprint.c

RossBencina:
(/common sources and headers should be C89)


philburk:
Addresses comment in https://github.com/PortAudio/portaudio/pull/1087#issuecomment-3419531187

Issue #1138, opened 2026-03-27
Streamline structure of PortAudio CMake build docs


RossBencina:
Now that #1044 has been merged, the CMake build docs `doc\src\tutorial\compile_cmake.dox` have been updated to reflect the current (v19.8) cmake build system. As far as we can determine the docs are now correct.

The implied structure of the document is now:

- Tutorial:
    - Steps to build PortAudio using CMake
- Platform-specific details/nuances/tweaks/hints/tricks
- Supported CMake build options reference
 
The following issues remain:
- The heading structure is flat enough that the above structure is not very clear. 
- The "Linux, OS X, and Others" section is mostly, if not entirely, a duplicate of the initial tutorial sections

Suggestions from Ross:

- Clarify the Tutorial/Platform-specific Details/Build Options structure using heading levels and a top index that states the contents of the document
- Only use discursive tutorial style text in the tutorial. For the later sections treat these as "user guide" and "reference manual" material.
- Reframe the platform-specific sections as additional details only, lift any generic information into the tutorial. If build steps/commands need to be made explicit in the platform-specific sections to provide clear context for the discussion, just put all of the build commands in a single code block. Don't use discursive tutorial style text.

RossBencina:
@generic-pers0n your input is welcome if you are still around. @philburk will be looking at this.

generic-pers0n:
> [@generic-pers0n](https://github.com/generic-pers0n) your input is welcome if you are still around. [@philburk](https://github.com/philburk) will be looking at this.

Sure thing! Lately I've not had too much time due to work and personal issues, but I'll try to stick around and give my feedback on things, if not making contributions myself! 😄

philburk:
I installed doxygen on my Mac using:  brew install doxygen

When I ran:    doxygen Doxyfile
I got:
```
Preprocessing /Users/phil/Work/portaudio/pagit/doc/src/tutorial/compile_cmake.dox...
/Users/phil/Work/portaudio/pagit/doc/src/tutorial/compile_cmake.dox:335: warning: Reached end of file while still searching closing '`' of a verbatim block starting at line 332
/Users/phil/Work/portaudio/pagit/doc/src/tutorial/compile_cmake.dox:335: warning: Reached end of file while still inside a (nested) comment. Nesting level 1 (possible line reference(s): 1)
Parsing file /Users/phil/Work/portaudio/pagit/doc/src/tutorial/compile_cmake.dox...
[...]

PR #1140, opened 2026-04-09
Remove back-tick that was breaking CMake doxygen docs

philburk:
Related to #1138


philburk:
Change comment to say "backtick" instead of "tilda".

RossBencina (approved):

PR #1146, opened 2026-04-21
Add Windows ARM64 support in CI

Harishmcw:
This PR adds Windows ARM64 support to the MSBuild and CMake CI workflows.

Changes:
- CMake
    - Added `Windows-MSVC-ARM64` and `Windows-MinGW-ARM64` jobs.
    - Configured appropriate vcpkg triplets and generators.
    - Added conditional `-A` handling for MSVC ARM64 builds.
- MSBuild
  - Added ARM64 platform configuration to `portaudio.vcproj` and `portaudio.sln`
  - Added ARM64 to the CI build matrix.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -7,7 +7,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        name: [Ubuntu-GCC, Ubuntu-MinGW, Windows-MSVC, Windows-MinGW, macOS-Clang, macOS-Clang-framework]
+        name: [Ubuntu-GCC, Ubuntu-MinGW, Windows-MSVC, Windows-MinGW, Windows-MSVC-ARM64, Windows-MinGW-ARM64, macOS-Clang, macOS-Clang-framework]
         shared_libs: [ON, OFF]
         include:
           - name: Ubuntu-GCC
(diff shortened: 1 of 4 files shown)


.github/workflows/cmake.yml
```diff
@@ -49,6 +49,29 @@ jobs:
...
+            # ASIO_SDK_ZIP_PATH needs to be quoted or CMake will save the download to
+            # asiosdk instead of asiosdk.zip.
+            asio_sdk_cache_path: "asiosdk.zip"
+            # Somehow CMake fails to find the toolchain file if a relative path is used on Windows.
+            cmake_options:
+              -DPA_USE_ASIO=ON
+              -DASIO_SDK_ZIP_PATH="asiosdk.zip"
+          - name: Windows-MinGW-ARM64
+            os: windows-11-arm
+            vcpkg_triplet: arm64-mingw-static
```
RossBencina, line 66: We notice that there is no `cmake_arch` assignment here (or for most of the other targets). Could you add comments indicating why this is not needed (maybe once at the use-site saying "if cmake_arch is not defined use the toolchain default" and a comment here: "arch is implicit by using the arm64-.. toolchain".

Harishmcw: Thanks for the feedback!

Added the comments as requested.

RossBencina (changes requested):
To check that this wasn't going to create backward compatibility problems I opened the `.sln` solution in Visual Studio 2005. There were no obvious problems introduced by the presence of the ARM64 configuration (it would not build of course). The Win32 target seemed to build (modulo some unrelated errors that we should probably fix). 

Could I get you to fix one more thing please? : When I opened the solution `ARM64` was already the selected configuration. Could you make open with the 64-bit intel config please?

Harishmcw:
Hi @RossBencina,

After some investigation I found that on Visual Studio 2005, the default configuration is picked alphabetically, since ARM64 sorts before x64, it ends up selected by default.

RossBencina:
@Harishmcw thanks for looking into it. We're intending to merge this as-is but could you please provide us with a short explanation of how you made the changes to the `.vcproj` file? Have you tested it locally? I don't have the ARM toolchain installed here. I can confirm that the `.vcproj` does load and convert in VS2022 with a warning about no ARM toolchain (see below) not a problem, just commenting for completeness. Interestingly VS2022 then selects the x64 configuration non-alphabetically even though there is a Win32 configuration.

image

RossBencina (approved):
Approved, but let's finish the conversation first. I would like to know how you generated the `.vcproj` changes, including which VisualStudio version you used, and whether you tested it locally. We may include some of these details in the commit message.

Harishmcw:
The `.vcproj` changes were made manually by cloning the existing `x64` configuration blocks, adding `_ARM64` to preprocessor definitions, and updating output filenames. The updated `.vcproj` was upgraded to `.vcxproj` using Visual Studio 2022 and built successfully for ARM64 locally, producing `portaudio_arm64.dll`.

Harishmcw:
Hi @RossBencina,

Just following up on this. Could you please review the MR when you get a chance?

PR #1151, opened 2026-05-06
Correct and improve CMake documentation

### LINKED ISSUE #745: Documentation: CMake needs `CMAKE_PREFIX_PATH` not `CMAKE_MODULE_PATH`
Closed the previous bug as it resulted from confusion between the release and the code from Git.

For the release, the only problem with the documentation (http://files.portaudio.com/docs/v19-doxydocs/compile_cmake.html) is that you must set `CMAKE_PREFIX_PATH` to the `CMAKE_INSTALL_PATH` where you installed PortAudio, not `CMAKE_MODULE_PATH`.

For the Git snapshot, the documentation is not at all valid, as the name of the package and the way it is used have totally changed.

dhdaines:
A number of issues are fixed:

- `--config` doesn't at all do what the doc said, as it is not accepted at configuration (generation) time and only works at build time on multi-target builds.  Best not to mention it at all.
- no need to pass `-G Unix Makefiles` as that is the default, but maybe you want to use Ninja
- everything about `CMAKE_MODULE_PATH` was wrong, add the correct way to point to an installed PortAudio

Fixes: #745

### DIFF
--- a/doc/src/tutorial/compile_cmake.dox
+++ b/doc/src/tutorial/compile_cmake.dox
@@ -127,7 +127,7 @@ As usual, you can alternatively invoke the build tool CMake has generated build 
 
 @subsection compile_cmake_general_running_tests_examples Running Tests and Examples
 
-As stated earlier, tests and examples are located within the `test` and `examples` folders, respectively. Likewise, Windows has an exception where both tests and examples are under an additional subdirectory named after the current configuration in their respective directories. (This also happens with multi-config generators too). In this example, we are going to assume that you have built these tests with the default `Debug` configuration. If your configuration is different, then adjust the examples accordingly. For a list of all possible build configurations, please see @ref compile_cmake_options_general.
+As stated earlier, tests and examples are located within the `test` and `examples` folders, respectively. Likewise, Windows has an exception where both tests and examples are under an additional subdirectory named after the current configuration in their respective directories. (This also happens with multi-config generators on other platforms). In this example, we are going to assume that you have built these tests with the default `Debug` configuration. If your configuration is different, then adjust the examples accordingly. For a list of all possible build configurations, please see @ref compile_cmake_options_general.
 
 To list tests, run the following command in the same terminal or command prompt window used earlier (substitute `ls` with `dir` at the Windows Command Prompt):
 
--- a/doc/src/tutorial/compile_cmake.dox
+++ b/doc/src/tutorial/compile_cmake.dox
@@ -85,15 +85,15 @@ To build _both_ examples _and_ tests alongside a normal PortAudio build, run the
 
     cmake -DPA_BUILD_TESTS=ON -DPA_BUILD_EXAMPLES=ON ../portaudio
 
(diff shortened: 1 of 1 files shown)


philburk:
@generic-pers0n - Hello Avery, can you please take a look at this?

generic-pers0n (approved):
Changes look good! I approve! 😄

RossBencina:
With respect to single configuration/multi-configuration generators, I think your suggestion of switching to describe single-cofiguration generators (e.g. Makefile generator) is certainly the simplest path, however both Xcode and Visual Studio generators are going to be common and both of those are multi-config (as is Ninja multi-config, which is what I usually use). So if you want to describe only single-config generators that at least needs to be mentioned I think (it also makes sense to be explicit about `-G Makefile` to avoid confusion.) An alternative is to add separate sections for single-config and multi-config generators, or just a single sentence about multi-config so that people at least know that it's a thing.

generic-pers0n:
> With respect to single configuration/multi-configuration generators, I think your suggestion of switching to describe single-cofiguration generators (e.g. Makefile generator) is certainly the simplest path, however both Xcode and Visual Studio generators are going to be common and both of those are multi-config (as is Ninja multi-config, which is what I usually use). So if you want to describe only single-config generators that at least needs to be mentioned I think (it also makes sense to be explicit about `-G Makefile` to avoid confusion.) An alternative is to add separate sections for single-config and multi-config generators, or just a single sentence about multi-config so that people at least know that it's a thing.

I agree in that switching to describing single-config generators is an easier route to take. I also do find it important to briefly discuss single-config and multi-config generators too, mentioning that both Visual Studio and Xcode are multi-config while Makefiles are single config, due to notable differences between the two types. I would also be okay with a simple and brief mention of the difference too, but such a difference may be good to bring up.

(I just saw @RossBencina's comment after I approved, but even without these changes I think things would at least be okay. It's still an improvement over before after all! :smile:)

dhdaines:
> I agree in that switching to describing single-config generators is an easier route to take. I also do find it important to briefly discuss single-config and multi-config generators too, mentioning that both Visual Studio and Xcode are multi-config while Makefiles are single config, due to notable differences between the two types. I would also be okay with a simple and brief mention of the difference too, but such a difference may be good to bring up.

Makes sense!  I will put back something like the previous mention of the use of `--config` when building.

I have to admit that I was unaware of Ninja multi-config, and will probably use it all the time now that I know that it exists.

RossBencina:
> Makes sense! I will put back something like the previous mention of the use of --config when building.

Great, thanks! It will be good to correct the docs.

dhdaines:
> > Makes sense! I will put back something like the previous mention of the use of --config when building.
> 
> Great, thanks! It will be good to correct the docs.

(back from long weekend) I will try to find some time to do this today...

dhdaines:
Should be good now!

generic-pers0n (approved):

RossBencina (approved):
Looks good, thanks. We're going to merge it. The only thing I noticed was that the paragraph on line 70 could mention Xcode, and I'm not sure whether the multiconfig artefacts are always placed directly in subdirs of the build directory for all generators (maybe they are, maybe not).

> When using Visual Studio to build on Windows, or the "Ninja Multi-Config" generator on Unix, the library files will be in a folder named after your configuration (e.g., `Debug/`, `Release/`, `RelWithDebInfo/`, etc.). Otherwise, they will simply end up in the root of your build directory. Now you are ready to begin the actual build.

philburk (approved):

PR #1156, opened 2026-06-23
Fix Float32_To_UInt8 undefined behavior on negative samples

### LINKED ISSUE #1155: Float32_To_UInt8 corrupts negative samples — undefined behavior casting a negative float to unsigned char
## Summary

In `src/common/pa_converters.c`, the non-clipping `Float32_To_UInt8` converter casts a possibly-negative `float` directly to `unsigned char`, which is undefined behavior (C11 §6.3.1.4). It happens to produce correct output on x86/MSVC by coincidence, but on conforming platforms — notably **AArch64/clang (Apple Silicon, Linux arm64, Raspberry Pi)** — every negative sample is mapped to `128`, destroying the entire negative half of the signal.

## Affected code

`Float32_To_UInt8` (~line 762):

```c
unsigned char samp = (unsigned char)(128 + ((unsigned char) (*src * (127.0f))));
```

The inner `(unsigned char)(*src * 127.0f)` converts a value in `[-127, 127]` to `unsigned char`. For any negative sample the truncated value cannot be represented in `unsigned char`, so the conversion is undefined behavior.

## Behavior by platform

Example with `*src = -1.0f` (expected output: `1`):

| Platform | `float`→`unsigned char` of `-127.0f` | Result | |
|---|---|---|---|
| x86 / MSVC (`cvttss2si` then truncate low byte) | `-127` → `0x81` = `129` | `128 + 129` → `1` | accidentally correct |
| AArch64 / clang (`fcvtzu`, saturates) | `0` | `128 + 0` → `128` | **wrong** |

On AArch64 *all* negative inputs saturate to `0`, so the output is a constant `128` for the whole negative half-cycle (severe distortion / DC-biased garbage).

## Why the sibling converters are unaffected

`Float32_To_UInt8_Clip`, `Float32_To_UInt8_Dither`, and `Float32_To_UInt8_DitherClip` all route through `PaInt32` before applying the `+128` bias, so only the plain non-clip variant exhibits the bug.

## Proposed fix

Go through a signed integer first (matching the `_Clip` variant):

```c
unsigned char samp = (unsigned char)(128 + (int)(*src * (127.0f)));
```

This yields the correct result on every platform: `(int)(-127.0f)` = `-127`, then `128 + (-127)` = `1`. It is bit-identical to the previous x86 output for all in-range samples.

## Notes

- Longstanding latent bug, increasingly relevant as PortAudio targets ARM (macOS arm64, Linux arm64).
- Reproduces with any `paUInt8` output stream without clipping carrying negative samples.

electro-logic:
Fixes #1155.

`Float32_To_UInt8` casts a possibly-negative `float` to `unsigned char` before adding the 128 bias, which is undefined behavior (C11 §6.3.1.4):

```c
unsigned char samp = (unsigned char)(128 + ((unsigned char) (*src * (127.0f))));
```

It happens to give the correct result on x86/MSVC (`cvttss2si` + low-byte truncation), but on **AArch64/clang** `fcvtzu` saturates negatives to `0`, so every negative sample becomes `128` — the entire negative half of the signal collapses to mid-scale. The `_Clip` / `_Dither` / `_DitherClip` UInt8 variants are unaffected because they route through `PaInt32`.

The fix casts through `int` first (matching those variants):

```c
unsigned char samp = (unsigned char)(128 + (int)(*src * (127.0f)));
```

Bit-identical to the previous x86 output for in-range samples, and correct on every platform. Full analysis in #1155.

### DIFF
--- a/src/common/pa_converters.c
+++ b/src/common/pa_converters.c
@@ -759,7 +759,10 @@ static void Float32_To_UInt8(
 
     while( count-- )
     {
-        unsigned char samp = (unsigned char)(128 + ((unsigned char) (*src * (127.0f))));
+        /* Cast to int before biasing: converting a negative float directly to unsigned char is
+           undefined behavior (C11 6.3.1.4) and saturates to 0 on some targets (e.g. AArch64),
+           collapsing the negative half of the signal to 128. */
+        unsigned char samp = (unsigned char)(128 + (int)(*src * (127.0f)));
         *dest = samp;
 
         src += sourceStride;


philburk:
Nice simple fix.

I see the the clipped and dithered versions of Float>UInt8 do not have the bug and do not need the fix.

philburk (approved):
Code looks good.

The CI failures seem unrelated. Maybe a change in the CI infrastructure.

RossBencina (approved):

PR #1159, opened 2026-06-26
CI: update cmake.yml github action workflow to use new GitHub default…

RossBencina:
… toolchain: Visual Studio 18 2026


philburk (approved):
Simple and it works.

PR #1160, opened 2026-07-02
Uniquely identify sound devices on Windows

bear101:
If a user unplugs a USB headset and plugs it into another USB port then the sound device may get a new PaDeviceIndex, i.e. we cannot use PaDeviceIndex to identify a sound device.

With this change a sound device (PaDeviceInfo) can be uniquely identified after a Pa_Terminate()/Pa_Initialize().

So far only Windows is supported (using the sound device's GUID).

### DIFF
--- a/include/portaudio.h
+++ b/include/portaudio.h
@@ -523,6 +523,7 @@ typedef struct PaDeviceInfo
     PaTime defaultHighOutputLatency;
 
     double defaultSampleRate;
+    const char* uniqueID;
 } PaDeviceInfo;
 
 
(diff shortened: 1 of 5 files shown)


include/portaudio.h
```diff
@@ -523,6 +523,7 @@ typedef struct PaDeviceInfo
     PaTime defaultHighOutputLatency;
 
     double defaultSampleRate;
+    const char* uniqueID;
```
philburk, line 526: This is an API change so it cannot go into V19.
We can discuss this for V20.

Changing structure sizes can break code, even with the structure version. Adding an additional function might be safer.

philburk:
CI is failing with:

D:/a/_temp/msys64/ucrt64/bin/../lib/gcc/x86_64-w64-mingw32/16.1.0/../../../../x86_64-w64-mingw32/bin/ld.exe: src/hostapi/wmme/.libs/pa_win_wmme.o:pa_win_wmme.c:(.rdata$.refptr.GUID_NULL[.refptr.GUID_NULL]+0x0): undefined reference to `GUID_NULL'

I recommend creating an Issue. We need to discuss the problem before we can consider the possible solutions.

PR #1164, opened 2026-07-13
docs: note that PA_ALSA_DYNAMIC defaults to OFF

### LINKED ISSUE #1163: CMake: PA_ALSA_DYNAMIC defaults to OFF, docs says defaults to ON
`CMakeLists.txt` line 353:
```
option(PA_ALSA_DYNAMIC "Enable dynamically loading libasound with dlopen using PaAlsa_SetLibraryPathName" OFF)
```

`compile_cmake.dox` lines 278-281:
```
@note All options here default to `ON` if available except for the OSS, ASIO, and skeleton backends.

- **`-DPA_USE_ALSA=ON|OFF`**: (Linux/Unix-like/etc. Only). Use the ASLA host API. Requires ALSA.
    - **`-DPA_ALSA_DYNAMIC=ON|OFF`**: Dynamically load libasound with dlopen using @ref PaAlsa_SetLibraryPathName.
 ```

Probably be better to tell the user how to run CMake to get the default values rather than making promises in the docs.

TheBoardRoomMarketing:
`CMakeLists.txt` sets `PA_ALSA_DYNAMIC` to `OFF`, but the note above the backend options list states everything defaults to `ON` except the OSS, ASIO, and skeleton backends — `PA_ALSA_DYNAMIC` isn't listed among those exceptions, so a reader following the note would expect it to default to `ON`.

This adds a footnote (matching the existing style used for the OSS exception) clarifying the actual default, rather than changing the build default itself.

Fixes #1163

### DIFF
--- a/doc/src/tutorial/compile_cmake.dox
+++ b/doc/src/tutorial/compile_cmake.dox
@@ -278,7 +278,7 @@ This section contains options that configure your build. Some options are specif
 @note All options here default to `ON` if available except for the OSS, ASIO, and skeleton backends.
 
 - **`-DPA_USE_ALSA=ON|OFF`**: (Linux/Unix-like/etc. Only). Use the ASLA host API. Requires ALSA.
-    - **`-DPA_ALSA_DYNAMIC=ON|OFF`**: Dynamically load libasound with dlopen using @ref PaAlsa_SetLibraryPathName.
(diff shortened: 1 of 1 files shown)


philburk (approved):
Thanks. I can see that it is set to OFF in CMakeLists.txt
      option(PA_ALSA_DYNAMIC "Enable dynamically loading libasound with dlopen using PaAlsa_SetLibraryPathName" OFF)

RossBencina (approved):

PR #1172, opened 2026-08-21
fix monotonic latency errors on MacOS

### LINKED ISSUE #1171: paqa_latency_monotonic failing on Mac OS due to small deltas
The patest_latency_monotonic.c test gradually increases the suggested latency while opening and closing a stream.
It checks the resulting final latency. The final latency should be equal to or greater than the
suggested latency until it reaches the maximum possible latency.
Once the maximum final latency is reached the final latency should never increase,
even if the suggested latency increases.
But sometimes the final latency continues to rise as the suggested latency rises.

**To Reproduce**
You need PR #1111 to get paqa_latency_monotonic.c
On Mac OS Tahoe 26.5.2
./configure && make
bin/paqa_latency_monotonic

**Expected behavior**
All tests to pass.

**Actual behavior**
Failures on most devices, for example:
```
=============== Using device #3: 'Maono PD200W Mic USB' (Core Audio) ==================
------------------------ paqaCheckMultipleSuggested - OUTPUT ------------
  lowLatency   = 0.00341667
  highLatency  = 0.01275
  numChannels  = 2
  sampleRate   = 48000 Hz
  samplePeriod = 2.083333e-05 seconds
  suggested[ 0] = 0.000000, final = 0.002396, (final - suggested) = 2.395833e-03 sec = 105.66 fr
  suggested[ 1] = 0.001275, final = 0.002396, (final - suggested) = 1.120833e-03 sec = 49.43 fr
  suggested[ 2] = 0.002550, final = 0.002542, (final - suggested) = -8.333333e-06 sec = -0.37 fr
     detectedMaximumLatency = 0.002542
  suggested[ 3] = 0.003825, final = 0.003812, (final - suggested) = -1.250000e-05 sec = -0.55 fr
/Users/phil/Work/portaudio/pagit/qa/pa_latency_monotonic.c:176 - ERROR - Latency should be == detectedMaximumLatency
OUTPUT CHECK FAILED !!! #3: 'Maono PD200W Mic USB'
```

philburk:
Rounding errors were causing the new paqa_latency_monotonic.c test to fail.

Fixes #1171

### DIFF
--- a/src/hostapi/coreaudio/pa_mac_core.c
+++ b/src/hostapi/coreaudio/pa_mac_core.c
@@ -68,9 +68,15 @@
 #include <math.h>
 #include <libkern/OSAtomic.h>
 
+#include <fenv.h>
+#pragma STDC FENV_ACCESS ON
+
 #include "pa_mac_core.h"
 #include "pa_mac_core_utilities.h"
 #include "pa_mac_core_blocking.h"
+
+#include <fenv.h>
+#pragma STDC FENV_ACCESS ON
 
 #ifdef __cplusplus
 extern "C"
@@ -82,6 +88,17 @@ extern "C"
 
 /* prototypes for functions declared in this file */
 PaError PaMacCore_Initialize( PaUtilHostApiRepresentation **hostApi, PaHostApiIndex index );
+
+static double rounded_divide(double numerator, double denominator, int round)
+{
+    int old_mode = fegetround();
+    if (fesetround(FE_UPWARD) == 0) {
+            double result = numerator / denominator;
+            fesetround(old_mode);
+            return result;
+    }
+    return numerator / denominator;
+}
 
 /*
  * Function declared in pa_mac_core.h. Sets up a PaMacCoreStreamInfoStruct
@@ -1720,12 +1737,6 @@ static PaError OpenStream( struct PaUtilHostApiRepresentation *hostApi,
     PaSampleFormat hostInputSampleFormat, hostOutputSampleFormat;
     UInt32 fixedInputLatency = 0;
     UInt32 fixedOutputLatency = 0;
-    /* Most of the tests pass with just the ceil() call. But there was one failure that
-     * needed this epsilon fix. The reportedLatency was low by 1e-19.
-     * This value value was chosen because it is much smaller than the human scale and
-     * much larger than the scale of the rounding errors.
-     */
-    const double kLatencyEpsilon = 1e-10; /* Rounding error in frames. */
     // Accumulate contributions to latency in these variables.
     UInt32 inputLatencyFrames = 0;
     UInt32 outputLatencyFrames = 0;
@@ -2050,13 +2061,8 @@ static PaError OpenStream( struct PaUtilHostApiRepresentation *hostApi,
     if( inputParameters )
     {
         inputLatencyFrames += PaUtil_GetBufferProcessorInputLatencyFrames(&stream->bufferProcessor);
-        double latency = inputLatencyFrames / sampleRate;
-        if( latency < inputParameters->suggestedLatency &&
-            (inputParameters->suggestedLatency - latency) * sampleRate < kLatencyEpsilon )
-        {
-            latency = inputParameters->suggestedLatency;
-        }
-        stream->streamRepresentation.streamInfo.inputLatency = latency;
+        /* round upward so that reported latency is not less than actual latency */
+        stream->streamRepresentation.streamInfo.inputLatency = rounded_divide(inputLatencyFrames, sampleRate, FE_UPWARD);
     }
     else
     {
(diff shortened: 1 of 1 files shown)


src/hostapi/coreaudio/pa_mac_core.c
```diff
@@ -1719,6 +1720,7 @@ static PaError OpenStream( struct PaUtilHostApiRepresentation *hostApi,
     PaSampleFormat hostInputSampleFormat, hostOutputSampleFormat;
     UInt32 fixedInputLatency = 0;
     UInt32 fixedOutputLatency = 0;
+    const double kLatencyEpsilon = 1e-10; /* Rounding error in frames. */
```
RossBencina: Is the epsilon rounding required to pass the test? Is there any rationale behind 1e-10?

src/hostapi/coreaudio/pa_mac_core.c
```diff
@@ -2043,7 +2045,13 @@ static PaError OpenStream( struct PaUtilHostApiRepresentation *hostApi,
...
     {
         inputLatencyFrames += PaUtil_GetBufferProcessorInputLatencyFrames(&stream->bufferProcessor);
-        stream->streamRepresentation.streamInfo.inputLatency = inputLatencyFrames / sampleRate;
+        double latency = inputLatencyFrames / sampleRate;
+        if( latency < inputParameters->suggestedLatency &&
+            (inputParameters->suggestedLatency - latency) * sampleRate < kLatencyEpsilon )
+        {
+            latency = inputParameters->suggestedLatency;
+        }
+        stream->streamRepresentation.streamInfo.inputLatency = latency;
```
RossBencina: Note that this is not actually correct. streamInfo.*Latency should include the rest of the latency stack not just the buffer processor latency. On Mac we have access to a lot of latency information and I'm pretty sure we are using it elsewhere. None of that should stop this patch being merged of course.

src/hostapi/coreaudio/pa_mac_core.c
philburk: Most of the tests pass with just the ceil() call. But there was one failure that needed this epsilon fix.
The reportedLatency was low by 1e-19.

So 1e-10 seems much smaller than the human scale and much larger than the scale of the rounding errors.

src/hostapi/coreaudio/pa_mac_core.c
philburk: True. A more accurate latency estimation from the host would help us choose a better buffer size and also provide a more accurate latency value. That would help recording apps compensate for latency offsets.
Improving latency estimates would be a nice enhancement.
We can measure the accuracy of the estimate by doing the round-trip loopback measurements.

src/hostapi/coreaudio/pa_mac_core.c
```diff
@@ -2043,7 +2050,13 @@ static PaError OpenStream( struct PaUtilHostApiRepresentation *hostApi,
     if( inputParameters )
     {
         inputLatencyFrames += PaUtil_GetBufferProcessorInputLatencyFrames(&stream->bufferProcessor);
-        stream->streamRepresentation.streamInfo.inputLatency = inputLatencyFrames / sampleRate;
+        double latency = inputLatencyFrames / sampleRate;
```
RossBencina: Try this approach for both input and output latency and see whether it fixes the test.

Add this to top of file:

```
#include <fenv.h>
#pragma STDC FENV_ACCESS ON
```

```suggestion
        /* round upwards so that reported latency is not less than actual latency */
        double latency;
        int old_mode = fegetround();
        if (fesetround(FE_UPWARD) == 0) {
                latency = inputLatencyFrames / sampleRate;
                fesetround(old_mode);
        } else {
                latency = inputLatencyFrames / sampleRate;
        }        
```

src/hostapi/coreaudio/pa_mac_core.c
```diff
@@ -2043,7 +2050,13 @@ static PaError OpenStream( struct PaUtilHostApiRepresentation *hostApi,
...
+        if( latency < inputParameters->suggestedLatency &&
+            (inputParameters->suggestedLatency - latency) * sampleRate < kLatencyEpsilon )
+        {
+            latency = inputParameters->suggestedLatency;
+        }
+        stream->streamRepresentation.streamInfo.inputLatency = latency;
```
RossBencina: Try this approach for both input and output latency and see whether it fixes the failing test.

Add this to top of file:

```
#include <fenv.h>
#pragma STDC FENV_ACCESS ON

double rounded_divide(double numerator, double denominator, int round)
{
    int old_mode = fegetround();
    if (fesetround(FE_UPWARD) == 0) {
            double result = numerator / denominator;
            fesetround(old_mode);
            return result;
    }
    return numerator / denominator;      
}
```

[Although note that it is safest to place the pragma and the rounded_divide() function definition in its own separate .c file in order to ensure that the `FENV_ACCESS` pragma applies to that one function. If it solves this problem I could imagine that we will need the function elsewhere in PortAudio.]

```suggestion
        /* round upward so that reported latency is not less than actual latency */
        stream->streamRepresentation.streamInfo.inputLatency = rounded_divide(inputLatencyFrames, sampleRate, FE_UPWARD);
```

src/hostapi/coreaudio/pa_mac_core.c
```diff
@@ -2053,7 +2066,13 @@ static PaError OpenStream( struct PaUtilHostApiRepresentation *hostApi,
...
     {
         outputLatencyFrames += PaUtil_GetBufferProcessorOutputLatencyFrames(&stream->bufferProcessor);
-        stream->streamRepresentation.streamInfo.outputLatency = outputLatencyFrames / sampleRate;
+        double latency = outputLatencyFrames / sampleRate;
+        if( latency < outputParameters->suggestedLatency &&
+            (outputParameters->suggestedLatency - latency) * sampleRate < kLatencyEpsilon )
+        {
+            latency = outputParameters->suggestedLatency;
+        }
+        stream->streamRepresentation.streamInfo.outputLatency = latency;
```
RossBencina:
```suggestion
         /* round upward so that reported latency is not less than actual latency */
        stream->streamRepresentation.streamInfo.outputLatency = rounded_divide(outputLatencyFrames, sampleRate, FE_UPWARD);
```

RossBencina (approved):

Issue #518, opened 2021-02-23
CMake expected ASIOSDK location is undocumented


EvanBalster:
When using CMake to generate portaudio with ASIO support, the ASIO SDK directory must be a sibling of the portaudio repository.  This is not documented in the ASIO readme:
https://github.com/PortAudio/portaudio/blob/master/src/hostapi/asio/ASIO-README.txt.

I had to investigate this CMake file to suss it out:
https://github.com/portaudio/portaudio/blob/master/cmake_support/FindASIOSDK.cmake

**To Reproduce**

# Use CMake to configure PortAudio for building in Visual Studio (or any Windows compiler).
# Select `PA_USE_ASIO` in CMake's configuration options.
# Attempt to generate the project.

**Expected behavior**

CMake build instructions should be documented at least briefly in https://github.com/PortAudio/portaudio/blob/master/src/hostapi/asio/ASIO-README.txt.

Recommend CMake should support one or more of the following:

- A configuration variable for ASIOSDK_ROOT_DIR
- Also scanning for the ASIO SDK in `src/hostapi/asio`

**Actual behavior**

CMake reports missing values of ASIOSDK_ROOT_DIR and ASIOSDK_INCLUDE_DIR.  Defining these as CMake environment variables has no effect.  The CMake configuration does not provide a a way to provide these values.

**Desktop:**
 - OS: Windows
 - OS Version: 10
 - PortAudio version: main branch
 - ASIO host API

Be-ing:
#461 searches the portaudio repository root and the directory above it for the SDK ZIP file. Alternatively the path can be explicitly specified with the CMake option `ASIO_SDK_ZIP_PATH`.

EvanBalster:
Historical methods for building Portaudio with ASIO support involved placing the ASIO SDK in `src/hostapi/asio`.  Although it's a strange location, it is still indicated by the documentation and so I would recommend adding it to the search directories.

RossBencina:
I agree with your "Expected behavior" @EvanBalster, the supported location(s) should be consistent between the docs, the legacy MSVC build file and CMake.  On the other hand, I agree with @Be-ing about the logical locations for the SDK.

It's not clear from @Be-ing's comment whether the SDK is expected to be zipped or unzipped. But certainly an unzipped option should be supported since it is common to patch bugs in the SDK.

Be-ing:
[...]

PR #1173, opened 2026-08-21
CMake: FindASIO.cmake: improve code for patching ASIO SDK bug

RossBencina:
The pre-existing code fixes a bug in the ASIO SDK where `lpdrv` is allocated using array new: `lpdrv = new ASIODRVSTRUCT[1];` but deleted using scalar delete: `delete lpdrv;` by replacing the deletion with `delete [] lpdrv;`.

This commit improves the patch logic as follows:

- Only apply the patch if required, i.e. if `lpdrv = new <ident>[<int>]` appears in the source code.
- Use whitespace tolerant regex for the replacement match string.
- Write the patched file to the build directory (${CMAKE_CURRENT_BINARY_DIR}/patched_asiolist.cpp) not the ASIO SDK directory.
- Use configure_file() so that the output timestamp is unchanged when the content is unchanged, so consumers don't rebuild on every configure.
- Add a dependency on the original unpatched file so the patch is regenerated if the SDK is swapped or the file is edited.
- STATUS messages now distinguish patched / already-fixed / skipped.

Note: an existing prepatched patched_asiolist.cpp is still honored.

Work towards #518

### DIFF
--- a/cmake/modules/FindASIO.cmake
+++ b/cmake/modules/FindASIO.cmake
@@ -61,18 +61,54 @@ if(ASIO_ROOT)
   message(STATUS "Found ASIO SDK: ${ASIO_ROOT}")
 
   if(ASIO_FOUND AND NOT TARGET ASIO::host)
-    #patch asiolist.cpp
-    if(NOT EXISTS "${ASIO_ROOT}/host/pc/patched_asiolist.cpp")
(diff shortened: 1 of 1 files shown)


philburk (approved):
I agree with the intent and the code seems OK with me.
But I am not a CMake expert. So it would be good to have a more expert review.

RossBencina:
@dmitrykos could you review this please?

Issue #77, opened 2008-02-19
CoreAudio implementation doesn't support non-interleaved ReadStream and WriteStream


PortAudio-admin:
*Issue created by @RossBencina*

There is no test for this case either.

See mailing list discussion thread 19/2/2008:
Re: [Portaudio] ASIO Blocking I/O Patch Implementation (a ring buffercontaining frames, not bytes)

> So I took a look at pa_mac_core_blocking.c, what I noticed is:
>
> - It doesn't seem to use the buffer processor
>
> - It doesn't appear to handle non-interleaved calls to ReadStream/ 
> WriteStream (shall I create a ticket for that?). This particular  
> case is one of the reasons why Sven is using the buffer processor on  
> the ReadStream()/WriteStream() side of the queue.
>

Yes, since that sounds like a major issue and I won't be able to  
address that anytime soon. Is there a test for it?

[nope]

There is a ticket to implement the test: #83

RossBencina:
Note that PA/ASIO implements ReadStream() and WriteStream() using the PaBufferProcessor, which should be able to handle the interleaved and non-interleaved cases. This code code could be copied/adapted.

ReadStream:
https://github.com/PortAudio/portaudio/blob/f217b421cf7336a5e7cc03181289df6c8974966e/src/hostapi/asio/pa_asio.cpp#L3547

WriteStream:
https://github.com/PortAudio/portaudio/blob/f217b421cf7336a5e7cc03181289df6c8974966e/src/hostapi/asio/pa_asio.cpp#L3718

PR #1175, opened 2026-08-25
Fix/1174 coreaudio startstop deadlock

### LINKED ISSUE #1174: macos: deadlock stopping audio
**Describe the bug**
There is a race condition on macos where stopping audio can lead to deadlock.

**To Reproduce**
1. Extract the c file and CMakeLists.txt in the attached zip:
[MacOSPortAudioDeadlock.zip](https://github.com/user-attachments/files/31388123/MacOSPortAudioDeadlock.zip)

2. standard CMake build: 
```
mkdir build
cd build
cmake .. -G Ninja
ninja
```

3. Run the test program. Will output to stdout every 100 start/stop cycles. Will report on deadlock. 
```
% ./pa_start_stop_deadlock 
pid 40056: hammering open/start/stop/close cycles, duplex (100000 iterations max)...
iteration 0

DEADLOCK: no progress for 10 seconds at iteration 53 (pid 40056)
```

**Expected behavior**
Expect it not to deadlock.

**Actual behavior**
The program typically will deadlock within a few hundred start/stop iterations.

 - OS: macOS
 - OS Version 26.4
 - PortAudio version: 19.7.0

**Additional context**
I somehow saw this twice during local testing, so investigated and created the attached minimal test program to reproduce. Details on the deadlock are in comments at the top of the pa_start_stop_deadlock.c.

I have a potential solution, will create a PR.

robsussman:
Fixes #1174 

## Problem

`startStopCallback`, the `kAudioOutputUnitProperty_IsRunning` property listener in `pa_mac_core.c`, calls `AudioUnitGetProperty`. CoreAudio can deliver that notification synchronously on the HAL IO thread while holding the HAL IOContext mutex, and `AudioUnitGetProperty` needs the AudioUnit instance mutex — which a concurrent `Pa_StopStream` already holds inside `AudioOutputUnitStop` while it waits for that same IOContext mutex. This AB-BA inversion permanently wedges the thread calling `Pa_StopStream`. See #1174 for the full analysis, stacks, and a standalone stress repro (typically deadlocks within a few hundred open/start/stop/close cycles of a full-duplex stream on distinct input/output devices).

## Fix (two commits)

1. **Break the lock cycle.** The listener infers the stop transition from `stream->state` instead of calling back into the AudioUnit: `StopStream`/`AbortStream` set `STOPPING` before stopping the units and `StartStream` sets `ACTIVE` before starting them, so the state filters the same transitions the `IsRunning` query did.

2. **Restore the stream-finished-callback delivery the minimal fix dropped** (`paComplete`/`paAbort` self-stops and spontaneous device stops). The listener now does no work at all: it pokes a per-stream GCD dispatch source (`dispatch_source_merge_data` is lock- and allocation-free, safe in the locked context the notification is delivered in). The source's handler runs on a per-stream serial queue where no CoreAudio locks are held, so it can safely query `IsRunning` to distinguish starts from stops, and fires the finished callback for stops that happen while `ACTIVE` or `CALLBACK_STOPPED`. Normal `StopStream`/`AbortStream` no longer depend on the notification: `FinishStoppingStream` fires the callback once the units are confirmed stopped — matching how other host APIs (skeleton, ALSA) invoke it from a PortAudio-owned non-realtime context rather than from an OS callback. This also stops user callback code from running inside CoreAudio's locked notification context, which was its own re-entrancy hazard. An atomic once-per-start flag keeps delivery exactly-once under races between spontaneous stops, callback-requested stops, and concurrent `Pa_StopStream`. `CloseStream` cancels the source and drains the queue before disposing the units the handler may query. 

Behavior preserved from the old code: a spontaneous stop leaves the stream state `ACTIVE` (the app still calls `Pa_StopStream`), and in two-unit duplex streams the output unit is the one consulted.

## Verification

- The stress repro attached to #1174: unpatched deadlocks within 121–299 cycles (4/4 runs); patched survives 3000 cycles. The repro's CMakeLists can be pointed at any checkout via `FETCHCONTENT_SOURCE_DIR_PORTAUDIO` to reproduce/verify.
- Verified unplug of USB device fires notification properly

### DIFF
--- a/src/hostapi/coreaudio/pa_mac_core.c
+++ b/src/hostapi/coreaudio/pa_mac_core.c
@@ -338,6 +338,54 @@ static PaError OpenAndSetupOneAudioUnit( const PaMacCoreStream *stream,
     PaUtil_SetLastHostErrorInfo( paCoreAudio, errorCode, errorText )
 
 /*
+ * Fire the user's stream finished callback, at most once per StartStream.
+ * May be called concurrently from FinishStoppingStream and from the
+ * startStopQueue handler; the atomic exchange makes sure only one fires.
+ */
+static void fireStreamFinishedOnce( PaMacCoreStream *stream )
+{
+    if( OSAtomicCompareAndSwap32Barrier( 1, 0, &stream->streamFinishedArmed ) )
+    {
+        PaStreamFinishedCallback *sfc = stream->streamRepresentation.streamFinishedCallback;
+        if( sfc )
+            sfc( stream->streamRepresentation.userData );
+    }
+}
+
+/*
+ * Runs on stream->startStopQueue in response to startStopCallback. Unlike in
+ * the listener itself, it is safe to call AudioUnitGetProperty here: nothing
+ * on this queue holds CoreAudio locks, so at worst the query blocks until a
+ * concurrent start/stop completes. Detects the two ways the units can stop
+ * without Pa_StopStream/Pa_AbortStream driving it: the user callback
+ * requested a stop (CALLBACK_STOPPED), or the device stopped spontaneously
+ * while ACTIVE (unplugged, etc.).
+ */
+static void handleStartStopNotification( void *inRefCon )
+{
+    PaMacCoreStream *stream = (PaMacCoreStream *) inRefCon;
+    if( stream->state == ACTIVE || stream->state == CALLBACK_STOPPED )
+    {
+        /* This may be a start or a stop notification (they coalesce in the
+           dispatch source); query the unit to find out. Mirror
+           FinishStoppingStream's choice of unit and element. */
+        AudioUnit unit = stream->outputUnit ? stream->outputUnit : stream->inputUnit;
+        AudioUnitElement element = stream->outputUnit ? 0 : 1;
+        UInt32 isRunning = 1;
+        UInt32 size = sizeof( isRunning );
+        OSStatus err = AudioUnitGetProperty( unit, kAudioOutputUnitProperty_IsRunning,
+                                             kAudioUnitScope_Global, element, &isRunning, &size );
+        /* on error, do nothing rather than misreport a stop */
+        if( !err && !isRunning )
+            fireStreamFinishedOnce( stream );
+        /* As before, a spontaneous stop leaves state ACTIVE: the user must
+           still call Pa_StopStream. */
+    }
+    /* STOPPING/STOPPED: FinishStoppingStream owns completion and fires the
+       stream finished callback itself. */
+}
+
+/*
  * Callback called when starting or stopping a stream.
  */
 static void startStopCallback(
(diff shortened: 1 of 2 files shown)


philburk:
@robsussman - I notice that your test runs against the release version of PortAudio 19.7.0.
We are working on the head of the GitHub repo.
Are you seeing the same hang on the current top of tree?

When I ran my test on an M4 Mac Mini.
I ran 13700 iterations and saw no deadlock.
Maybe that is because my input device was a USB mic.

What device are you running on?
A Macbook built-in Mic may have very different timing.

robsussman:
@philburk, good point, I just retested against master and it happened there too with both Debug and Release builds, both times in less than 300 iterations.

I'm running Tahoe 26.4 on an M5 Macbook Pro. 

I tried changing the system default input and output devices to a USB audio interface (Scarlett 4i4 USB). It took longer to deadlock but still did at iteration 769. 

Tested on another machine: MacBook Pro M1 running macOS Ventura 13.7.8 with built in mic and speakers. So far 4,100 iterations and no deadlock. So maybe this requires a newer macOS version to reproduce?

philburk:
I was able to repro the deadlock on my M4 Mac Mini running Mac OS 26.6.2.
I am using PortAudio top-of-tree.

With the changes in this PR #1175, I cannot reproduce the deadlock. Yay! **Thank you so much** for the clear repro test program and the fix.

Our next step is to fully review the change to see if there are any unintended changes in behavior, which can cause a regression. I will run our small QA suite and the test programs.

nat3Github:
just to add a new datapoint, i am on a macbook m1 pro and encountered deadlocks when closing my programm pretty often (especially closing when busy). i traced the hang down to a race in portaudio. 

File and function: src/hostapi/coreaudio/pa_mac_core.c, startStopCallback
  - Thread 1: Pa_StopStream → FinishStoppingStream → AudioOutputUnitStop. It holds the AudioUnit mutex and waits for the mutex inside HALC_ProxyIOContext::StopIOProc.
  - Thread 2: HALC_ProxyIOContext::IOWorkLoop (holding the HAL mutex) → the IsRunning listener startStopCallback → AudioUnitGetProperty, which waits for the AudioUnit mutex.

philburk:
@nat3Github - Thanks for the report. Do you think you are seeing the same bug? It certainly sounds similar.

Does this PR solve the problem for you?  You can checkout the fix by doing:

```
git checkout master
git fetch origin pull/1175/head:pr_1175
git checkout pr_1175
```

philburk:
I am testing the fix with various QA tests:

bin/paqa_errs - PASS

bin/paqa_devs hung
------ Test: OUTPUT, device = 3, #ch = 4, rate = 48000, format = 1, CALLBACK, INTER
------ Test: OUTPUT, device = 3, #ch = 4, rate = 88200, format = 1, CALLBACK, INTER
------ Test: OUTPUT, device = 3, #ch = 4, rate = 96000, format = 1, CALLBACK, INTER
                           HUNG HERE !!!

reran bin/paqa_devs and it PASSed except for BLOCKING, NON-INT. Known issue. See #77
reran bin/paqa_devs and it PASSED. - cannot reproduce the first hang...

bin/paqa_latency - PASS
bin/paqa_latency_monotonic - PASS

The following were run and also sounded fine: PASS
bin/paex_sine
bin/paex_write_sine
bin/paex_record
bin/patest_multi_sine 
bin/patest_leftright
bin/patest_stop_playout
bin/patest_wire

I then plugged in a Zoom AMS-22 USB device with hardware loopback support.
make loopback
bin/pa_loopback - PASS

